PR #281, opened 2018-06-08
Full endpoint should return all values

EmilyM1:


pkg/chargeback/http.go
```diff
@@ -440,19 +440,26 @@ type ReportResultEntry struct {
 }
 
 type ReportResultValues struct {
-	Name  string      `json:"name"`
-	Value interface{} `json:"value"`
+	Name        string      `json:"name"`
+	Value       interface{} `json:"value"`
+	TableHidden bool        `json:"tableHidden"`
+	Unit        string      `json:"unit"`
```
chancez: Lets set `json:"unit,omitempty"` so the `unit` field isn't in the output unless `unit` is non-empty.

chancez:
One minor comment. After that's fixed go ahead and squash the commits I can merge.

PR #284, opened 2018-06-12
Use separate column types for ReportGenerationQuery and PrestoTable CRs

chancez:
PrestoTable CRs shouldn't have information unrelated to Presto in them, so use a separate type for ReportGenerationQuery columns that can have things like TableHidden and Unit fields.
This also updates the code using these columns to use the correct column type.


chancez:
jenkins rebuild

PR #291, opened 2018-06-14
Http tableHidden set to false endpoint

EmilyM1:


pkg/chargeback/http.go
```diff
@@ -188,16 +189,27 @@ func (srv *server) getReportHandler(w http.ResponseWriter, r *http.Request) {
...
+	srv.getReport(logger, name, r.Form["format"][0], true, true, w, r)
+}
+
+func (srv *server) getReportV2TableHandler(w http.ResponseWriter, r *http.Request) {
+	logger := srv.newLogger(r)
+	name := chi.URLParam(r, "name")
+	if !srv.validateGetReportReq(logger, []string{"format"}, w, r) {
+		return
+	}
+	srv.logger.Infof("WITHIN V2 HANDLER")
```
chancez: Remove

pkg/chargeback/http.go
```diff
@@ -448,16 +460,15 @@ type ReportResultValues struct {
 }
 
 func convertsToGetReportResults(input []map[string]interface{}, columns []api.ReportGenerationQueryColumn) GetReportResults {
+	// converts to json/csv report with more fieldnames and ability to modify results if tableHidden is false
```
chancez: I would rephrase this as it's not super clear atm, and secondly, in Go you generally put the comment above the function name, and start with the function name's name.

```
// convertsToGetReportResults converts maps returned from `presto.ExecuteSelect` into a GetReportResults.
```

pkg/chargeback/http.go
```diff
@@ -471,14 +482,31 @@ func convertsToGetReportResults(input []map[string]interface{}, columns []api.Re
 	return results
 }
 
-func (srv *server) writeResultsV2(logger log.FieldLogger, format string, columns []api.ReportGenerationQueryColumn, results []map[string]interface{}, w http.ResponseWriter, r *http.Request) {
+func (srv *server) writeResultsV2(logger log.FieldLogger, full bool, format string, columns []api.ReportGenerationQueryColumn, results []map[string]interface{}, w http.ResponseWriter, r *http.Request) {
+	columnsMap := make(map[string]api.ReportGenerationQueryColumn)
+	var filteredColumns []api.ReportGenerationQueryColumn
+	for _, column := range columns {
```
chancez: I would add a comment on what this loop is for. Perhaps:

```
// Build a new list of columns if full is false, containing only columns with TableHidden set to false.
```

pkg/chargeback/http.go
```diff
@@ -471,14 +482,31 @@ func convertsToGetReportResults(input []map[string]interface{}, columns []api.Re
...
+		columnsMap[column.Name] = column
+		showColumn := !columnsMap[column.Name].TableHidden
+		if showColumn || full {
+			filteredColumns = append(filteredColumns, column)
+		}
+	}
+	for _, row := range results {
```
chancez: Same as above, a comment on this like:
```
// Remove columns and their values from `input` if full is false and the column's TableHidden is true.
```

pkg/chargeback/http.go
```diff
@@ -647,5 +675,5 @@ func getReportOrderByColumnsString(query *api.ReportGenerationQuery) string {
 			quotedColumns = append(quotedColumns, fmt.Sprintf(`"%s"`, colName))
 		}
 	}
-	return fmt.Sprintf("%s ASC", strings.Join(quotedColumns, ", "))
+	return fmt.Sprintf("period_start, period_end, %s ASC", strings.Join(quotedColumns, ", "))
```
chancez: This change shouldn't be required anymore.

pkg/chargeback/http.go
```diff
@@ -448,16 +458,15 @@ type ReportResultValues struct {
 }
 
 func convertsToGetReportResults(input []map[string]interface{}, columns []api.ReportGenerationQueryColumn) GetReportResults {
+	// convertsToGetReportResults converts maps returned from `presto.ExecuteSelect` into a GetReportResults
```
chancez: the comment goes above the function name

PR #318, opened 2018-07-17
Unit Tests to validate v2 Full endpoint

EmilyM1:


pkg/chargeback/http_test.go
```diff
@@ -333,3 +335,282 @@ func TestAPIV1ReportsGet(t *testing.T) {
...
+		expectedStatusCode int
+		expectedAPIError   string
+
+		queryerPrepareFunc func(*mockpresto.MockExecQueryer) []presto.Row
+		expectedResults    *GetReportResults
+	}{
+		"report-finished-with-results-full": {
+			reportName: testReportName,
+			report:     newTestReport(testReportName, namespace, testQueryName, reportStart, reportEnd, v1alpha1.ReportStatus{Phase: v1alpha1.ReportPhaseFinished}),
+			query: newTestReportGenQuery(testQueryName, namespace, []v1alpha1.ReportGenerationQueryColumn{
```
chancez: This query should have one field with tableHidden true, and one with false like the others to test that this is in fact returning columns with table hidden to true.

chancez:
The one query needs updating, but looks good otherwise @EmilyM1

pkg/chargeback/http_test.go
EmilyM1: Great catch! Thank you

chancez:
Failure was a flake on the GKE cluster, just going to re-trigger the build

PR #319, opened 2018-07-19
Makefile: allow customization of base image url

tschuy:
As part of my testing environment for work on the containers, I've created repos under my own name on Quay so that I have a little more room to breathe. This makefile change lets me build and push these images in an easy fashion.

(My eventual goal here is a three-liner to `docker-build-all`, `docker-push-all`, and template a custom values file that contains the images by sha so I can iterate on the image install as quickly as possible.)


chancez:
jenkins rebuild

PR #329, opened 2018-07-26
V2 tableHidden unit tests

EmilyM1:


pkg/chargeback/http_test.go
```diff
@@ -616,3 +616,317 @@ func TestAPIV2ReportsGet(t *testing.T) {
...
+					{
+						"timestamp": time.Time{},
+						"foo":       1,
+					},
+				}
+				mock.EXPECT().Query(gomock.Any()).Return(result, nil)
+				return result
+			},
+			expectedStatusCode: http.StatusOK,
+			expectedResults: &GetReportResults{
```
chancez: These expectedResults aren't correct. The timestamp value is missing.

pkg/chargeback/http_test.go
```diff
@@ -616,3 +616,317 @@ func TestAPIV2ReportsGet(t *testing.T) {
...
+								Name:        "foo",
+								Value:       1,
+								TableHidden: false,
+							},
+						},
+					},
+				},
+			},
+		},
+		//add here
```
chancez: remove comment

pkg/chargeback/http_test.go
```diff
@@ -616,3 +616,317 @@ func TestAPIV2ReportsGet(t *testing.T) {
...
+			query: newTestReportGenQuery(testQueryName, namespace, []v1alpha1.ReportGenerationQueryColumn{
+				{
+					Name:        "timestamp",
+					Type:        "timestamp",
+					TableHidden: true,
+				},
+				{
+					Name:        "foo",
+					Type:        "double",
+					TableHidden: true,
```
chancez: One of these should be false so we can see results with one column removed, and one kept in the results.

pkg/chargeback/http_test.go
```diff
@@ -616,3 +616,317 @@ func TestAPIV2ReportsGet(t *testing.T) {
...
+				{
+					Name: "timestamp",
+					Type: "timestamp",
+				},
+				{
+					Name: "foo",
+					Type: "double",
+				},
+			}),
+			queryerPrepareFunc: func(mock *mockpresto.MockExecQueryer) []presto.Row {
```
chancez: The queryPrepareFunc should be returning results, so we are actually testing that the columns aren't displayed when we get results back.

pkg/chargeback/http_test.go
EmilyM1: My thinking was that the tableHidden-true test would have two true bools and tableHiddenFalse would test two false bools.  On the test on line 646 both bools are false.
Would you like each test (tableHidden-true and tableHidden-false) to have one of each bool then? For consistency? Ie change tableHidden false test as well.

chancez: Sure that's fine.

pkg/chargeback/http_test.go
EmilyM1: fixed by bool change

pkg/chargeback/http_test.go
EmilyM1: done. Thank you

pkg/chargeback/http_test.go
EmilyM1: Done

pkg/chargeback/http_test.go
EmilyM1: Done

pkg/chargeback/http_test.go
chancez: Actually now that both one of these test cases is redundant since it tests both behaviors, hidding and showing a column. Can you remove `report-finished-with-results-tableHidden-true` and rename `report-finished-with-results-tableHidden-false` to `report-finished-with-results`.

chancez:
LGTM once tests pass. Can you squash?

pkg/chargeback/http_test.go
EmilyM1: Very true.
Done.

PR #331, opened 2018-07-27
V2 specific endpoint tests

EmilyM1:


pkg/chargeback/http.go
```diff
@@ -79,6 +79,8 @@ func newRouter(logger log.FieldLogger, queryer presto.ExecQueryer, rand *rand.Ra
 	router.HandleFunc(APIV1ReportsGetEndpoint, srv.getReportHandler)
 	router.HandleFunc("/api/v2/reports/{name}/full", srv.getReportV2FullHandler)
 	router.HandleFunc("/api/v2/reports/{name}/table", srv.getReportV2TableHandler)
+	router.HandleFunc("/api/v2/reports//full", srv.UnitTestHandler)
```
chancez: Can you rename the handlers to something else? More like `getReportV2FullNameMissingHandler`, and similar for table?
Can you also add a comment why these routes exist, ex:
> // These two routes handle returning a 400 when the name parameter is missing, rather than having a 404 returned.

pkg/chargeback/http.go
```diff
@@ -79,6 +79,9 @@ func newRouter(logger log.FieldLogger, queryer presto.ExecQueryer, rand *rand.Ra
 	router.HandleFunc(APIV1ReportsGetEndpoint, srv.getReportHandler)
 	router.HandleFunc("/api/v2/reports/{name}/full", srv.getReportV2FullHandler)
 	router.HandleFunc("/api/v2/reports/{name}/table", srv.getReportV2TableHandler)
+	// The following two routes handle returning a 400 when the name parameter is missing, rather than having a 404 returned.
+	router.HandleFunc("/api/v2/reports//full", srv.getReportV2NameMissingHandler)
+	router.HandleFunc("/api/v2/reports//table", srv.getReportV2NameMissingHandler)
```
chancez: you used the same handler in both routes here.

chancez:
LGTM can you squash?

EmilyM1:
Squashed and rebased

PR #333, opened 2018-08-01
Documentation,test: Fix proxy URL to metering service in docs

chancez:
Update to using http instead of 8080 in tests too

### DIFF
--- a/test/framework/chargeback.go
+++ b/test/framework/chargeback.go
@@ -5,10 +5,13 @@ import (
 	restclient "k8s.io/client-go/rest"
 )
 
-const chargebackServiceName = "metering"
+const (
+	chargebackServiceName     = "metering"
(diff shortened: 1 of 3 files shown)


test/framework/chargeback.go
```diff
@@ -5,10 +5,13 @@ import (
 	restclient "k8s.io/client-go/rest"
 )
 
-const chargebackServiceName = "metering"
+const (
+	chargebackServiceName     = "metering"
+	chargebackServicePortName = "http"
```
pruan-rht: `http` is a protocol, so I think the name should be changed to something like `chargebackServiceProtoName`

chancez: This is already what it is, this is just fixing existing things to use the correct port name. Mostly in docs but adjusting tests to use the right one too.

PR #345, opened 2018-08-13
Optimize helm-operator build process

tschuy:
Download fewer unnecessary packages for building helm and tiller, and build helm instead of downloading a precompiled version.

### DIFF
--- a/images/helm-operator/Dockerfile
+++ b/images/helm-operator/Dockerfile
@@ -5,7 +5,7 @@ FROM centos:7 as build
 RUN yum -y update && yum clean all
 
 RUN mkdir -p /go && chmod -R 777 /go && \
-    yum -y install epel-release && yum -y install hg git golang glide && yum group install -y "Development Tools"
+    yum -y install epel-release && yum -y install protobuf hg git golang glide make
 RUN yum clean all && rm -rf /var/cache/yum
 
 ENV HELM_VERSION 2.6.2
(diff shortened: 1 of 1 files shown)


images/helm-operator/Dockerfile
```diff
@@ -5,7 +5,7 @@ FROM centos:7 as build
 RUN yum -y update && yum clean all
 
 RUN mkdir -p /go && chmod -R 777 /go && \
-    yum -y install epel-release && yum -y install hg git golang glide && yum group install -y "Development Tools"
+    yum -y install epel-release && yum -y install protobuf hg git golang glide make
```
chancez: I thought kubectl was one too?

images/hive/Dockerfile
```diff
@@ -14,18 +14,8 @@ RUN set -x \
...
-
-# Install PostgreSQL JDBC
-RUN set -x \
-    && curl -fSLs -o "$HIVE_HOME/lib/$POSTGRESQL_JDBC_JAR" "https://jdbc.postgresql.org/download/$POSTGRESQL_JDBC_JAR"
-
-# Install MySQL JDBC
-RUN set -x \
-    && curl -fSLs "https://dev.mysql.com/get/Downloads/Connector-J/$MYSQL_JDBC_VERSION.tar.gz" | tar -zx --strip-components=1 -C "$HIVE_HOME/lib" "$MYSQL_JDBC_VERSION/$MYSQL_JDBC_JAR"
+# Install PostgreSQL and MySQL JDBCs
+RUN yum install -y mysql-connector-java postgresql-jdbc && yum clean all && rm -rf /var/cache/yum
```
chancez: These are definitely not going to necessarily be in the right spots, so you probably need to actually validate this with a mysql/postgresql setup.

chancez: Eg: "$HIVE_HOME/lib/$POSTGRESQL_JDBC_JAR"

We probably just need to make sure that where these go ends up in our Java classpath, and all these wrapper scripts correctly use that.

images/helm-operator/Dockerfile
tschuy: ah, it looks like I wasn't looking at `epel`. There's a 1.7.5 copy of `kubectl` in there.

I don't see it in brew, which is worth noting.

chancez: 1.5 is probably too old

PR #352, opened 2018-08-20
Set Slack alerts on CI

EmilyM1:


jenkins/main.groovy
```diff
@@ -127,16 +128,67 @@ pipeline {
...
+    }
+
+
+    notifySlack("${buildColor}",
+            """
+            |*${env.JOB_NAME} build ${env.BUILD_NUMBER}*
+            |${buildEmoji} Build ${buildStatus}.
+            |link: ${env.BUILD_URL}
+            |branch: ${env.BRANCH_NAME}
+            |environments: ${environments.join(', ')}
```
chancez: Gotta remove this.

jenkins/main.groovy
```diff
@@ -127,16 +128,67 @@ pipeline {
...
+            """.stripMargin())
+
+}
+
+def notifySlack(color, message) {
+    slackSend(
+        channel: '#team-metering-ci',
+        color: color,
+        message: message,
+        teamDomain: 'coreos',
```
chancez: Remove this

jenkins/main.groovy
```diff
@@ -127,16 +128,67 @@ pipeline {
...
+        tokenCredentialId: 'team-metering-ci-slack-token',
+        failOnError: true,
```
chancez: Remove this

jenkins/main.groovy
```diff
@@ -127,16 +128,63 @@ pipeline {
                     echo 'Updating Github PR status'
                     def status
                     def description
+                    def message
```
chancez: this and the uses of `message` below it can be removed

jenkins/main.groovy
```diff
@@ -127,16 +128,63 @@ pipeline {
...
+    // build status of null means successful
+    buildStatus =  buildStatus ?: 'SUCCESSFUL'
+
+    // Override default values based on build status
+    switch (buildStatus) {
+        case 'STARTED':
+            buildColor = 'warning'
+            buildEmoji = ':open_mouth:'
+            break;
+        case 'SUCCESSFUL':
```
chancez: I guess this should be `case 'SUCCESS':` based on the githubNotify logic checking for `SUCCESS` not `SUCCESSFUL`.

jenkins/main.groovy
```diff
@@ -127,16 +128,63 @@ pipeline {
...
         }
     }
 }
+
+def notifyBuild(String buildStatus = 'STARTED') {
+    def buildColor = ''
+    def buildEmoji = ''
+
+    // build status of null means successful
+    buildStatus =  buildStatus ?: 'SUCCESSFUL'
```
chancez: this should also be `'SUCCESS'` not `'SUCCESSFUL'`

jenkins/main.groovy
```diff
@@ -14,6 +14,7 @@ def prStatusContext = 'jenkins/main'
 if (isPullRequest) {
     echo 'Setting Github PR status'
     githubNotify context: prStatusContext, status: 'PENDING', description: 'Build started'
+    notifyBuild('STARTED')
```
chancez: We actually want to send notifications on non-PR builds too, so let's move this instance of `notifyBuild` just after line 18 so it's outside the conditional checking if it's a PR

jenkins/main.groovy
```diff
@@ -135,8 +136,52 @@ pipeline {
                         description = "Some stages failed"
                     }
                     githubNotify context: prStatusContext, status: status, description: description
+                    notifyBuild(currentBuild.currentResult)
```
chancez: Same thing here, we want to run this outside of PRs, so move it just after line 140 to be outside the `if (isPullRequest)` conditional

jenkins/main.groovy
```diff
@@ -15,6 +15,7 @@ if (isPullRequest) {
     echo 'Setting Github PR status'
     githubNotify context: prStatusContext, status: 'PENDING', description: 'Build started'
 }
+    notifyBuild('STARTED')
```
chancez: fix indentation please

jenkins/main.groovy
```diff
@@ -136,7 +137,50 @@ pipeline {
                     }
                     githubNotify context: prStatusContext, status: status, description: description
                 }
+                                    notifyBuild(currentBuild.currentResult)
```
chancez: fix indentation please

jenkins/continuous-upgrade.groovy
```diff
@@ -32,16 +34,26 @@ pipeline {
             when {
                 branch 'master'
             }
-            steps {
```
chancez: You shouldn't need to change this at all. The post section should just be nested within `pipeline { }` with just the notifiy stuff in it, none of the stuff you moved around.

PR #362, opened 2018-08-29
readme: add links to docs

robszumski:
We have a bunch of great content that we're not showing off!

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,16 +9,11 @@ Metering records historical cluster usage, and can generate usage reports showin
 
 Read more about the implemented and planned features in the documentation:
 
+ - [Installation Guide](Documentation/install-metering.md) - install Metering on your Kubernetes cluster
  - [Usage Guide](Documentation/using-metering.md) - start here to learn how to use the project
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -7,7 +7,12 @@ Metering records historical cluster usage, and can generate usage reports showin
 
 ## Project status: alpha
 
-Most major planned features are implemented, however the API is still subject to change.
+Read more about the implemented and planned features in the documentation:
```
chancez: These could all probably come after the installation section rather than preceding it.

robszumski: I thought it was more important to highlight what it is and what it will become before install given the alpha nature

robszumski: how about we move install into this list?

chancez: I just want it to be obvious which section you start with, and it should probably be installation since it is the entrypoint into other docs. This means putting install before other links, and making sure it's fairly obvious. We can move install into the list, but it should be the first link, and we should probably call it out as what most people should start with.

robszumski: Updated

PR #386, opened 2018-09-20
Support passing arbitrary inputs to Reports and ScheduledReports via spec.inputs

chancez:
This is an initial prototype for how we can make ReportGenerationQueries
more reusable and composable. It should be backwards compatible with
existing Reports and ScheduledReports, but existing
ReportGenerationQueries not in the repo would need to be updated to use
the correct template variables.


chancez:
All recent e2e test failures are due to recent CI changes, not the tests. THe tests are actually passing so I'm merging this to get some other stuff unblocked.

PR #394, opened 2018-09-24
pkg/{apis,operator}: Enable ReportGenerationQueries to reference Reports and ScheduledReports

chancez:
This should allow roll up reporting by having reports do smaller
aggregations on other reports that have lower granularity. For example,
if you have an hourly usage report, you could have a daily usage report
that uses a query to summarize the results from the hourly usage report
into daily buckets.

The other use case is to enable reports to calculate independent results
and then have a "mega-report" which combines the results all together.

In the future, if PR #386 is merged, which adds arbitrary inputs to
Reports/ScheduledReports and their queries, which could also allow for
these "roll-up" or "mega-reports" to behave differently based on
parameters passed in.


chancez:
Self reminder: still need to look into queuing a ReportGenerationQuery when it's dependent scheduledReports and reports tables are created.

PR #396, opened 2018-09-24
pkg/operator: Always hide TableHidden columns if output is CSV or Tabular

chancez:
Fixes an error when maps exist in the columns of a ReportGenerationQuery, since we cannot flatten those structures in CSV/Tabular format. So use TableHidden to always filter these columns from tabular/csv formatted results in both v1 and v2 endpoints.

### DIFF
--- a/pkg/operator/http.go
+++ b/pkg/operator/http.go
@@ -260,7 +260,7 @@ func (srv *server) getScheduledReport(logger log.FieldLogger, name, format strin
 		return
 	}
 
-	writeResultsResponse(logger, format, reportQuery.Spec.Columns, results, w, r)
+	writeResultsResponseV1(logger, format, reportQuery.Spec.Columns, results, w, r)
 }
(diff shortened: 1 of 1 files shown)


pkg/operator/http.go
```diff
@@ -499,21 +499,59 @@ func convertsToGetReportResults(input []presto.Row, columns []api.ReportGenerati
 	return results
 }
 
-func writeResultsResponseV2(logger log.FieldLogger, full bool, format string, columns []api.ReportGenerationQueryColumn, results []presto.Row, w http.ResponseWriter, r *http.Request) {
+func writeResultsResponseV1(logger log.FieldLogger, format string, columns []api.ReportGenerationQueryColumn, results []presto.Row, w http.ResponseWriter, r *http.Request) {
```
EmilyM1: Why remove the bool here? It was added to distinguish between table and full in V2.

EmilyM1:
LGTM 👍

PR #398, opened 2018-09-24
Add sprig template functions

chancez:
Adds functions from github.com/Masterminds/sprig to be available in our
ReportGenerationQuery query templates. Separated from #386 to avoid
adding too much in one of the more complex PRs.


chancez:
E2E is failing just on GKE because the nodes restarted and Prometheus lost it's data (not persisted on our GKE environment unfortunately) so I'm merging.

PR #402, opened 2018-09-26
Update to 0.7.1-latest now that 0.7.0 is tagged.

chancez:


chancez:
Re-ran integration and it passed.

PR #407, opened 2018-09-27
Support collecting metrics from custom Prometheus

tschuy:
Allows specifying a prometheus url in data source config:

```
spec:
  promsum:
    query: node-allocatable-cpu-cores
    prometheusConfig:
      url: https://second-prometheus-service
```


pkg/operator/operator.go
```diff
@@ -434,6 +415,29 @@ func (op *Reporting) Run(stopCh <-chan struct{}) error {
 	return nil
 }
 
+func (op *Reporting) newPrometheusConnFromURL(URL string) (prom.API, error) {
```
chancez: url should be lowercase

PR #414, opened 2018-10-01
Add table and full endpoint docs

EmilyM1:


Documentation/reportgenerationqueries.md
```diff
@@ -11,7 +11,8 @@ When writing a [report](report.md) you can specify the query it will use by sett
 - `columns`: A list of columns that match the schema of the results of the query. The order of these columns must match the order of the columns returned by the SELECT statement. Columns have 3 fields, `name`, `type`, and `unit`. Each field is covered in more detail below.
   - `name`: This is the name of the column returned in the `SELECT` statement.
   - `type`: This is the [Hive][hive-types] column type. Currently due to implementation details, column types are expressed using hive types. In the future, this will likely be switched to using the Presto native types. This also has an effect that queries with columns containing complex types such as `maps` or `arrays` cannot be used by `Reports` or `ScheduledReports`.
-  - `unit`:
+  - `unit`: Unit refers to the measurement of the column.
```
chancez: Normally I see the phrase "unit of measurement" rather than just "measurement", so maybe just adjust slightly like this?:
`Unit refers to the unit of measurement for the column.`

Documentation/reportgenerationqueries.md
```diff
@@ -11,7 +11,8 @@ When writing a [report](report.md) you can specify the query it will use by sett
...
+  - `tableHidden`: Takes a Bool the either shows or hides the column. This determines whether the column will be displayed or not in reports.
```
chancez: This probably should be more details about what it hides/shows and when.

`Takes a boolean, when true, hides the column from report results depending on the format and endpoint. If true, the /api/v2/reports/$reportName/table endpoint will filter tableHidden columns out of the results. Columns with TableHidden set to true are always hidden for tabular and CSV outputs.`

Documentation/reportgenerationqueries.md
```diff
@@ -164,6 +165,18 @@ spec:
     ORDER BY pod_request_memory_byte_seconds DESC
 ```
 
+## Modifying Columns For Report Display
+
+You can modify the custom resource to display all columns or to hide or show columns as needed. The full endpoint displays all columns.
```
chancez: Change `modify the custom resource` to `modify the ReportGenerationQuery`

Documentation/reportgenerationqueries.md
```diff
@@ -164,6 +165,18 @@ spec:
...
+An example report for `namespace-cpu-request` in JSON for the full endpoint would be:
```
chancez: This isn't an example report, it's an example of querying a report results from the reporting-operator API. Can you rephrase this to be more accurate?

Documentation/reportgenerationqueries.md
```diff
@@ -164,6 +165,18 @@ spec:
...
+`http://127.0.0.1:8001/api/v1/namespaces/metering/services/http:reporting-operator:http/proxy/api/v2/reports/namespace-cpu-request/full?format=json`
+To use the TableHidden display feature:
+- enter `TableHidden` field in query as true or false
```
chancez: Could you be more specific or maybe preferably instead, you could just mention an example where this is already set to true for a specific query.

Documentation/reportgenerationqueries.md
```diff
@@ -164,6 +165,18 @@ spec:
...
+- run report
```
chancez: Use complete sentences

Documentation/reportgenerationqueries.md
```diff
@@ -164,6 +165,18 @@ spec:
...
+An example report for `namespace-cpu-request` in JSON for the tableHidden endpoint would be:
```
chancez: Like the earlier comment, this is an example of querying a report, so it would help this was updated to be more accurate.

Documentation/reportgenerationqueries.md
chancez: Actually link to the API docs on the v2 table endpoint for the second sentence instead. 

Just say `See [api docs for details][apiTable].` and set `[apiTable]: api.md#v2-reports-table` at the bottom of the file.

So the full line would be:
```
Takes a boolean, when true, hides the column from report results depending on the format and endpoint. See [api docs for details][apiTable].
```

With this at the bottom of the file:
```
[apiTable]: api.md#v2-reports-table
```

Documentation/reportgenerationqueries.md
EmilyM1: all fixed in commit b04edc419b35fa20983f1ec9dfcfa082560962bf

chancez:
Looks good, can you squash?

PR #417, opened 2018-10-02
pkg/operator: Fix panic caused by incorrect assignment

chancez:
In the previous "else" case, after a recent change, the code incorrectly
used := instead of =, meaning the importer variable in the case of an
importer not existing, was not correctly being assigned to, and was
being shadowed in the else case. This corrects that problem by using a
closure to return the importer instead, which also fixed another issue
in not correctly unlocking the Mutex if creating the prometheus
connection failed by using defer.


chancez:
For posterity it was introduced in https://github.com/operator-framework/operator-metering/pull/407

PR #418, opened 2018-10-04
images/presto: Switch to using presto base image

chancez:


tschuy:
I like it. So much cleaner than we had before.

for PR context, the new base image is https://github.com/operator-framework/presto/blob/metering-0.212/Dockerfile

PR #424, opened 2018-10-09
Documentation: roll-up report guide & explanation

tschuy:
The reports generated with the examples in the docs produce the same output.

I'd like to rendez-vous with you about why we're doing both `<=` and `>=` in general in ReportGenerationQueries by default. I think that's probably a mistake in general, and will lead to things getting miscounted no matter what. After all, even without automated roll-up, there's at least someone somewhere who deals with these reports in chunks, and as it stands we're double-counting every start/end point.

Also, thoughts on what to do about the `data_start`/`data_end` on the aggregated reports? It seems to me that we either have to accept internal gaps that we can't expose on larger scales (how do we propagate the fact that on the 3rd we were missing twenty minutes of data?), fail out, or add a data_period that isn't just a pair of dates.

### DIFF
--- a/Documentation/report.md
+++ b/Documentation/report.md
@@ -297,3 +297,29 @@ A report can have the following states:
 
 
 [rfc3339]: https://tools.ietf.org/html/rfc3339#section-5.8
+
+## Roll-up Reports
+
+Report data is stored in the database much like metrics themselves, and can thus be used in aggregated or roll-up reports. A simple use case for a roll-up report is to spread the time required to produce a report over a longer period of time: instead of requiring a monthly report to query and add all data over an entire month, the task can be split into daily reports that each run over a thirtieth of the data.
+
+A custom roll-up report requires a custom generation query. The ReportGenerationQuery processor provides a macro that can get the necessary table name [from a report name](rollup-reports.md#2-create-the-aggregation-query):
+
+```
+# namespace-cpu-usage-aggregated-query.yaml
+inputs:
+- name: AggregatedReportName
+  required: true
+...
+ WHERE {| .Report.Inputs.AggregatedReportName | scheduledReportTableName |}.period_start >= timestamp '{| default .Report.ReportingStart .Report.Inputs.ReportingStart | prestoTimestamp |}'
+```
+
+```
+# aggregated-report.yaml
+spec:
+  generationQuery: "namespace-cpu-usage-aggregated"
+  inputs:
+  - name: "AggregatedReportName"
+    value: "namespace-cpu-usage-hourly"
+```
+
+For more information on setting up a roll-up report, see the [roll-up report guide](rollup-reports.md).
--- a/Documentation/index.md
+++ b/Documentation/index.md
@@ -15,6 +15,7 @@
 Custom Resources:
 
 - [Reports](report.md)
+  - [Roll-up Reports](rollup-reports.md)
 - [ReportGenerationQueries](reportgenerationqueries.md)
 - [ReportDataSources](reportdatasources.md)
 - [ReportPrometheusQueries](reportprometheusqueries.md)
--- a/Documentation/rollup-reports.md
+++ b/Documentation/rollup-reports.md
@@ -1,0 +1,82 @@
+# Guide: Roll-Up Reports
+
+Often, it makes sense to report on data collected in other reports, called roll-up reports. Roll-up reports can combine datatypes (memory *and* CPU usage per namespace) and can combine smaller periods of time into larger ones — for instance, in an environment where a large amount of data is collected, splitting up processing over a month instead of waiting for the end of the month can be an effective way to spread out compute time and get a quicker final report.
+
+Currently, there are no built-in roll-up reports. Therefore, a custom roll-up report requires a custom generation query. In the following guide, we will create a daily report that aggregates hourly reports:
+
+## 1. Create the sub-report
+
+```
+apiVersion: metering.openshift.io/v1alpha1
+kind: ScheduledReport
+metadata:
+  name: namespace-cpu-usage-hourly
+spec:
+  generationQuery: "namespace-cpu-usage"
+  reportingStart: '2018-10-09T00:00:00Z'
+  schedule:
+    period: "hourly"
+```
+
+## 2. Create the aggregation query
+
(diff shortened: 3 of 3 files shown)


tschuy:
Adding skip-all because although this represents a lot of work for me, it's actually still just a docs PR :stuck_out_tongue:

tschuy:
My personal opinion is that we change all `ReportGenerationQueries` to be left-inclusive — that is, to have `timestamp >= report start date AND timestamp < report end data`.

tschuy:
Also, I'd like to make an easier way to do this. We can of course always support doing it the manual way like this, but maybe adding the ability to automatically create and aggregate a smaller report when creating a larger report:

```
kind: ScheduledReport
metadata:
  name: namespace-cpu-usage-per-month
spec:
  generationQuery: "namespace-cpu-usage"
  reportingStart: '2018-10-09T00:00:00Z'
  schedule:
    period: "monthly"
    monthly:
      aggregateFrom: "daily"
```

My first thought is immediately that we'd run into issues with divisibility -- months can't be broken into weeks. The following would be feasible:
* monthly -> daily, hourly
* weekly -> daily, hourly
* daily -> hourly

chancez:
> My personal opinion is that we change all ReportGenerationQueries to be left-inclusive — that is, to have timestamp >= report start date AND timestamp < report end data.

I'm generally in agreement, though we need to consider regular reports, where in this scenario they won't capture the reportEnd time, and we may want to. Same for the very last period of a scheduledReport with a specific reportingEnd. If you do a monthly report, beginning of year to end, you won't get the full month of december, you'll get up to 11:59:59 though. Does this seem reasonable? I guess writing it out I don't find that all that wrong.

> Also, thoughts on what to do about the data_start/data_end on the aggregated reports?

I'm honestly not sure we need these. They seemed like something useful to have originally but it's not really all that useful as you've mentioned. We may want to just remove these in another PR, but that's somewhat invasive since we'll need to update our integration tests to not expect those fields.

> adding the ability to automatically create and aggregate a smaller report when creating a larger report

I would like it to be easier too, but as you've mentioned there's some complexity in it.

I don't think doing it in the core reporting logic is a what we want thought. It will never be perfect or as flexible as the manual approach, and I think we can make it relatively easy if we can make reusable queries with enough templating logic to handle the average cases that you've mentioned. With arbitrary inputs we can do this pretty easily. Additionally; we can provide out-of-the-box aggregation queries for the most common cases and just instruct users to use the one for the data they care about.

Additionally, I think I'd like to see more usage of this first before we make anything first class.

Documentation/report.md
```diff
@@ -292,3 +292,140 @@ A report can have the following states:
...
+    required: true
+  query: |
+    SELECT
+      timestamp '{| default .Report.ReportingStart .Report.Inputs.ReportingStart| prestoTimestamp |}' AS period_start,
+      timestamp '{| default .Report.ReportingEnd .Report.Inputs.ReportingEnd | prestoTimestamp |}' AS period_end,
+      namespace,
+      sum(pod_usage_cpu_core_seconds) as pod_usage_cpu_core_seconds
+    FROM {| .Report.Inputs.AggregatedReportName | scheduledReportTableName |}
+    WHERE {| .Report.Inputs.AggregatedReportName | scheduledReportTableName |}.period_start >= timestamp '{| default .Report.ReportingStart .Report.Inputs.ReportingStart | prestoTimestamp |}'
+    AND {| .Report.Inputs.AggregatedReportName | scheduledReportTableName |}.period_end <= timestamp '{| default .Report.ReportingEnd .Report.Inputs.ReportingEnd | prestoTimestamp |}'
```
chancez: Should this one be `<` not `<=`?

Documentation/report.md
```diff
@@ -292,3 +292,140 @@ A report can have the following states:
 
 
 [rfc3339]: https://tools.ietf.org/html/rfc3339#section-5.8
+
+## Roll-Up Reports
```
chancez: I like this. I think we'll want more details on the purpose and reason for why you want this and potentially a few more examples. I'm thinking perhaps we do a few things to make that easier to manage:

- We should probably add some built-in aggregated report queries. We can start with namespace for memory & cpu. 
- We can include one inline-version like you have here in the documentation to make it very clear what's happening.
- Perhaps this section could just link to a separate "guide" on roll-up reports. It's kind of a big concept and may deserve it's own document. What do you think?
- We should fix up the underlying issues you've discovered and then remove documentation about these work-arounds for the double-counting, and fix up these queries names.

chancez:
Did you end up getting any measurements on how long the aggregated report takes to run per-period compared to the naive report which runs over an entire period? Precise timings could be retrieved by port-forwarding to the presto UI, it will show how long individual queries take.

tschuy:
depends #424

Documentation/report.md
```diff
@@ -297,3 +297,29 @@ A report can have the following states:
 
 
 [rfc3339]: https://tools.ietf.org/html/rfc3339#section-5.8
+
+## Roll-up Reports
+
+Report data is stored in the database much like metrics themselves, and can thus be used in aggregated or roll-up reports. A simple use case for a roll-up report is to spread processing load across a period of time: instead of running a monthly report all at once, it can instead be a simple sum of daily reports.
```
chancez: `is to spread processing load` could potentially be more clear we're talking about the processing load to produce a report. Perhaps: ` is to spread the processing required to produce a report`? Not sure I like that either. I just think we need report/reporting in there some how.

Documentation/report.md
```diff
@@ -297,3 +297,29 @@ A report can have the following states:
...
+
+A custom roll-up report requires a custom generation query. The ReportGenerationQuery processor provides a macro that can get the necessary table name from a report name:
```
chancez: Can you perhaps make the `from a report name` text link to rollup-reports.md at the section where this query is introduced? It's not super obvious where `namespace-cpu-usage-aggregated-query.yaml` comes from since the link comes after these snippets. 

I think this would be good:

```
The ReportGenerationQuery processor provides a macro that can get the necessary table name [from a report name](rollup-reports.md#2-create-the-aggregation-query):
```

The ReportGenerationQuery processor provides a macro that can get the necessary table name [from a report name](rollup-reports.md#2-create-the-aggregation-query):

PR #425, opened 2018-10-09
Documentation/report: add custom inputs info

tschuy:
Another of the docs weaknesses that I found when working on #424

### DIFF
--- a/Documentation/report.md
+++ b/Documentation/report.md
@@ -286,6 +286,21 @@ kubectl -n $METERING_NAMESPACE get reportgenerationqueries namespace-memory-requ
 }
 ```
 
+### Inputs
+
+The `inputs` field of a Report `spec` can be used to pass custom values into a `ReportGenerationQuery`.
+
+It is a list of name-value pairs:
+
+```
+spec:
+  inputs:
+  - name: AggregatedReportName
+    value: namespace-cpu-usage-hourly
+```
+
+For an example of how this can be used, see it in action [in a roll-up report](rollup-reports.md#3-create-the-aggregator-report).
+
 ### Report Status
 
 The execution of a report can be tracked using its status field. Any errors occurring during the preparation of a report will be recorded here.


Documentation/report.md
```diff
@@ -281,7 +281,18 @@ kubectl -n $METERING_NAMESPACE get reportgenerationqueries namespace-memory-requ
...
+The `inputs` field of a Report `spec` can be used to pass custom values into a `ReportGenerationQuery`. It is an array of name-value pairs:
+
+```
+spec:
+  inputs:
+  - name: AggregatedReportName
+    value: namespace-cpu-usage-per-5-min
+```
+
+## Report Status
```
chancez: You changed the heading size

Documentation/report.md
```diff
@@ -281,7 +281,18 @@ kubectl -n $METERING_NAMESPACE get reportgenerationqueries namespace-memory-requ
...
-### Report Status
+### Inputs
+
+The `inputs` field of a Report `spec` can be used to pass custom values into a `ReportGenerationQuery`. It is an array of name-value pairs:
+
+```
+spec:
+  inputs:
+  - name: AggregatedReportName
+    value: namespace-cpu-usage-per-5-min
```
chancez: Perhaps use `namespace-cpu-usage-daily` rather than this less-common use-case.

Documentation/report.md
```diff
@@ -281,7 +281,18 @@ kubectl -n $METERING_NAMESPACE get reportgenerationqueries namespace-memory-requ
 }
 ```
 
-### Report Status
+### Inputs
+
+The `inputs` field of a Report `spec` can be used to pass custom values into a `ReportGenerationQuery`. It is an array of name-value pairs:
```
chancez: - Probably just use the term `list` rather than `array`. It's less jargon.
- Put the second sentence on it's own line.

We probably don't want to merge this until later since it really has minimal use right now and it probably needs more examples of how it connects to a real query. That's why I didn't document it yet; it's very difficult to see how this is used without having a built-in query that supports custom inputs. ReportingStart/ReportingEnd are also not the best example since they're a bit more special case.

Documentation/report.md
```diff
@@ -286,6 +286,21 @@ kubectl -n $METERING_NAMESPACE get reportgenerationqueries namespace-memory-requ
 }
 ```
 
+### Inputs
+
+The `inputs` field of a Report `spec` can be used to pass custom values into a `ReportGenerationQuery`.
+
+It is an array of name-value pairs:
```
chancez: `s/array/list/`

PR #433, opened 2018-10-18
ReportGenerationQueries: less-than instead of less-than-or-equal at end of reports

tschuy:
Make ReportGenerationQueries non-inclusive so that we can have these reports report all data just before, say, february 1 12:00:00am without having to rely on report writers to know what the proper epsilon is. Currently ScheduledReports also automatically use `x` and `x+period` as their ReportingStart/ReportingEnd, which means they duplicate data.

I think I'd like this to be documented in a "louder" way. @chancez, thoughts on how best to display this in the documentation? Right now, I've just put it in the variables section, but perhaps I should add comments to the most prominent examples?


Documentation/reportgenerationqueries.md
```diff
@@ -29,7 +29,7 @@ Most of these functions are for referring to other resources such as `ReportData
 
 - `Report`: This object has two fields, `ReportingStart` and `ReportingEnd` which are the value of the `spec.reportingStart` and `spec.reportingEnd` for a `Report`. For a `ScheduledReport` the values map to the specific period being collected when the `ScheduledReport` runs.
   - `ReportingStart`: A [time.Time][go-time] object that is generally used to filter the results of a `SELECT` query using a `WHERE` clause.
-  - `ReportingEnd`: A [time.Time][go-time] object that is generally used to filter the results of a `SELECT` query using a `WHERE` clause.
+  - `ReportingEnd`: A [time.Time][go-time] object that is generally used to filter the results of a `SELECT` query using a `WHERE` clause. The end timestamp is non-inclusive; a report set to start at 12:00am on January 1st and end at 12:00am on February 1st will collect data for the entire month of January, but not 12:00:00am on February 1st.
```
chancez: Technically the inclusiveness depends on the query contents itself, since ReportingEnd is passed in unchanged. I think maybe we should just mention that all built-in queries will use a `<= ReportingEnd` when filtering results and leave it at that right here.

chancez:
I think it would make sense to call out separately more than in the ReportingEnd field's doc, since it's more about if the query itself uses the value, and how. It might also be something worth explaining more in the writing-custom-queries.md guide.

PR #436, opened 2018-10-26
ScheduledReport: add preliminary run-once ability

tschuy:
Just a very preliminary test implementation. Not ready for review, I just want the CI to tell me if I've screwed anything else up in the process :)


pkg/operator/scheduled_reports.go
```diff
@@ -503,10 +513,21 @@ func (op *Reporting) runScheduledReport(logger log.FieldLogger, report *cbTypes.
 	return nil
 }
 
-func getNextReportPeriod(schedule reportSchedule, period cbTypes.ScheduledReportPeriod, lastScheduled time.Time) reportPeriod {
+func getRunOnceReportPeriod(report *cbTypes.ScheduledReport) (*reportPeriod, error) {
+	if report.Spec.ReportingEnd == nil || report.Spec.ReportingStart == nil {
```
pruan-rht: @tschuy, I think we should add more error checking on validity of start and end time.

tschuy: the ReportingEnd and ReportingStart are compared to ensure they're in the right order upon running the report: https://github.com/operator-framework/operator-metering/blob/master/pkg/operator/scheduled_reports.go#L212-L218

We could duplicate this check here as well. That would mean the actual report itself fails to validate rather than the run of it. @chancez thoughts on this?

chancez: We're planning on supporting reports without a reportingStart/reportingEnd, so I don't think it makes sense. We just need to make our ReportGenerationQuery spec.query templates handle it and we can actually remove the validation.

chancez:
Looks fine overall. Let's get some E2E tests added.

pkg/operator/scheduled_reports.go
pruan-rht: What would happen if the user gave a reportingEnd time that is earlier than reportingStart?

chancez:
I just realized for parity with reports we will need 2 more things, which could be done as separate PRs, or in this one:
- Need to support spec.runImmediately which doesn't wait for reportingStart, or reportingEnd
- By default, reports will overwrite any data that might exist in an existing table. currently we try to drop tables after delete events, but it isn't guaranteed, so we drop the table (deleteExistingData on GenerateReport) before inserting data into it. For scheduledReports, we don't drop the table at all except on deletes. I think we can probably update this logic to drop the table if the scheduledReport.status.tableName is "" (eg: when we haven't tried to create the table yet, and haven't set the status.tableName).

pkg/operator/scheduled_reports.go
```diff
@@ -499,10 +509,21 @@ func (op *Reporting) runScheduledReport(logger log.FieldLogger, report *cbTypes.
...
+		return nil, fmt.Errorf("run-once reports must have both ReportingEnd and ReportingStart")
+	}
+	reportPeriod := &reportPeriod{
+		periodEnd:   report.Spec.ReportingStart.UTC(),
```
chancez: These are mixed up

PR #450, opened 2018-10-31
Cluster pvc storage request 

EmilyM1:


charts/reporting-operator/templates/custom-resources/report-queries/pod-volume-request.yaml
```diff
@@ -165,3 +165,48 @@ spec:
...
+  - name: period_end
+    type: timestamp
+    unit: date
+  - name: data_start
+    type: timestamp
+    unit: date
+  - name: data_end
+    type: timestamp
+    unit: date
+  - name: volume_request_memory_bytes_seconds
```
tschuy: we probably don't want to refer to storage as memory, since that's sort of ambiguous (could be RAM). Perhaps `volume_request_storage_bytes_seconds`?

test/e2e/main_test.go
```diff
@@ -151,6 +151,12 @@ func TestReportingE2E(t *testing.T) {
 			timeout:       reportTestTimeout,
 			skip:          !runAWSBillingTests,
 		},
+		{
+			name:          "cluster-persistentvolumeclaim-request",
+			queryName:     "cluster-persistentvolumeclaim-request",
+			newReportFunc: testFramework.NewSimpleReport,
+			timeout:       reportTestTimeout + time.Minute,
```
chancez: I think timeout can just be `reportTestTimeout`.

charts/reporting-operator/templates/custom-resources/report-queries/pod-volume-request.yaml
chancez: That's a good point. It's `spec.resources.requests.storage` in the PVC yaml. We should update the other queries too (in a separate commit).

EmilyM1: Actually, this is referenced everywhere as memory- including the raw query. I will open a PR for the renaming and rename the file in one separate PR

test/e2e/main_test.go
EmilyM1: Fixed in commit 1e1ff2d9f183d3c30a279ea8f1865aa378d8c7ff

charts/reporting-operator/templates/custom-resources/report-queries/pod-volume-request.yaml
```diff
@@ -165,3 +165,48 @@ spec:
...
+  - name: period_end
+    type: timestamp
+    unit: date
+  - name: data_start
+    type: timestamp
+    unit: date
+  - name: data_end
+    type: timestamp
+    unit: date
+  - name: volume_request_storage_bytes_seconds
```
chancez: `bytes` should be `byte` here.

EmilyM1: Fixed in commit 8c15b524ce065275b661be4f40cab78955e1dab8

PR #454, opened 2018-11-01
Set ScheduledReport condition failure if validation fails

chancez:
The idea here is to surface dependency validation errors to users when a scheduledReport has a reportGenerationQuery which is uninitialized, or has uninitialized ReportDataSource/ReportGenerationQuery dependencies itself.

To test this:
- Modify a reportGenerationQuery and add a non-existent ReportDataSource and/or ReportGenerationQuery to it's spec.reportDataSources or spec.reportQueries field.
   - `kubectl edit reportgenerationquery namespace-cpu-usage` then edit the `spec.reportQueries` list to include an item that doesn't exist.
- Create a scheduledReport that uses the above reportGenerationQuery

Query the scheduledReport and check it's status:
```
kubectl get scheduledreport
NAME                            QUERY                    SCHEDULE   RUNNING                     FAILED             TABLE NAME                                       LAST REPORT TIME       AGE
namespace-cpu-usage-hourly      namespace-cpu-usage      hourly     ValidatingScheduledReport   FailedValidation                                                    2018-11-01T12:00:00Z   14s
```

### DIFF
--- a/pkg/apis/metering/v1alpha1/util/scheduled_report_util.go
+++ b/pkg/apis/metering/v1alpha1/util/scheduled_report_util.go
@@ -19,6 +19,10 @@ const (
 	// spec.reportingStart.
 	InvalidReportingEndReason = "InvalidReportingEnd"
 
+	// FailedValidationReason is added to a ScheduledReport when the it's
+	// ReportGenerationQuery or it's dependencies aren't ready
+	FailedValidationReason = "FailedValidation"
+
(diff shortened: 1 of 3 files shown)


chancez:
We definitely only validate later in the reconcile after we decide if it's time to run yet. I'm not sure what the best behavior is, but maybe we should be validating earlier. That can be done in another PR though.

PR #458, opened 2018-11-01
Drop scheduledReport table before creating

chancez:
It's possible the scheduledReport table could already exist when seeing a new report due to a previous failed cleanup, so this just makes the handler more robust.
Ensures that the scheduledReport table is empty when we're handling a new scheduledReport by checking for status.tableName.

Additionally, we can avoid doing the createTable and updating the CR if if the status.tableName is set.


EmilyM1:
LGTM

PR #460, opened 2018-11-02
 Remove hadoop image build and update chart to handle configuration and adding of scripts

chancez:
This removes the need for the metering-hadoop image which is mostly some scripts and configuration files that we can handle directly in our chart instead.

https://github.com/operator-framework/hadoop/blob/metering-3.1.1/Dockerfile is the location of the base image Dockerfile. 

Overall this simplifies our build and makes integrating with OSBS simpler because we have one less image to worry about.


chancez:
Yeah, I've tested it locally and e2e is a good test of this 👍

PR #463, opened 2018-11-02
pkg/operator: Update resource DeepCopies to happen earlier

chancez:
To ensure we are never mutating the objects returned from the cache,
create the copy immediately after it's retrieved.

### DIFF
--- a/pkg/operator/datasources.go
+++ b/pkg/operator/datasources.go
@@ -79,11 +79,12 @@ func (op *Reporting) syncReportDataSource(logger log.FieldLogger, key string) er
 		return err
 	}
 
-	return op.handleReportDataSource(logger, reportDataSource)
+	// Deep-copy otherwise we are mutating our cache
+	ds := reportDataSource.DeepCopy()
+	return op.handleReportDataSource(logger, ds)
 }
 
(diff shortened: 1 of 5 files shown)


pkg/operator/datasources.go
```diff
@@ -79,11 +79,12 @@ func (op *Reporting) syncReportDataSource(logger log.FieldLogger, key string) er
 		return err
 	}
 
-	return op.handleReportDataSource(logger, reportDataSource)
+	// Deep-copy otherwise we are mutating our cache
+	ds := reportDataSource.DeepCopy()
+	return op.handleReportDataSource(logger, ds)
 }
 
 func (op *Reporting) handleReportDataSource(logger log.FieldLogger, dataSource *cbTypes.ReportDataSource) error {
```
pbx0: Should all callers of this function use the deepcopy beforehand?

chancez: This is the only caller. The handle functions generally only get called in one spot.

chancez:
Mostly I moved them to ensure we're never mutating them before deepcopying, which is required to avoid breaking the informer cache/deltaFifo queue.

PR #465, opened 2018-11-03
Remove hive image build, customize hive in chart

chancez:
Removes the metering-hive image and uses the quay.io/coreos/hive image
directly, and instead customizes configuration and entrypoint scripts using
configmaps/secrets and helm templates.

The hive base image is at https://quay.io/repository/coreos/hive?tab=tags and the Dockerfile is https://github.com/operator-framework/hive/tree/metering-2.3.3/Dockerfile


chancez:
Tested with mysql and postgresql as well as S3 for storage. Everything works.

PR #471, opened 2018-11-09
Only validate non-zero imported metric count of ReportDataSources that are dependencies of queries being tested

chancez:
Previously CollectMetricsOnce blindly assumed all ReportDataSources
should have non-zero metrics, but we're getting to a point where we
aren't always using every ReportDataSource in every e2e env, so we need
the ability to only validate the metric import count for a
ReportDataSource is non-zero if it's a direct dependency of the queries
we're testing.

### DIFF
--- a/pkg/operator/reporting/validate.go
+++ b/pkg/operator/reporting/validate.go
@@ -193,7 +193,7 @@ func GetDependentViewGenerationQueries(queryGetter reportGenerationQueryGetter, 
 		viewQueries = append(viewQueries, query)
 	}
 	dataSources := make([]*metering.ReportDataSource, 0, len(dataSourcesAccumulator))
-	for _, ds := range dataSources {
+	for _, ds := range dataSourcesAccumulator {
 		dataSources = append(dataSources, ds)
 	}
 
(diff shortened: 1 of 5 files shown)


pkg/operator/reporting/validate.go
```diff
@@ -272,14 +324,29 @@ func NewReportDataSourceClientGetter(getter meteringClient.ReportDataSourcesGett
...
+}
+
+func GetDependentDataSources(dataSourceGetter reportDataSourceGetter, generationQuery *metering.ReportGenerationQuery) ([]*metering.ReportDataSource, error) {
+	dataSourceAccumulator := make(map[string]*metering.ReportDataSource)
+	err := GetDependentDataSourcesMemoized(dataSourceGetter, generationQuery, dataSourceAccumulator)
+	if err != nil {
+		return nil, err
+	}
+	dataSources := make([]*metering.ReportDataSource, 0, len(dataSourceAccumulator))
+	for _, ds := range dataSources {
```
pbx0: I'm a little confused by this for loop, but it may be me. Shouldn't we range over `dataSourceAccumulator` and put that in the `dataSources` slice?

chancez: Yeah good catch.

PR #473, opened 2018-11-09
charts/reporting-operator: Make node-memory/cpu-capacity/allocatable queries filterable

chancez:
Moves the aliasing to a -raw suffixed query and adds a query with
reportingStart/reportingEnd inputs to replace the previous.

### DIFF
--- a/charts/reporting-operator/templates/custom-resources/report-queries/node-cpu.yaml
+++ b/charts/reporting-operator/templates/custom-resources/report-queries/node-cpu.yaml
@@ -1,7 +1,7 @@
 apiVersion: metering.openshift.io/v1alpha1
 kind: ReportGenerationQuery
 metadata:
-  name: "node-cpu-capacity"
+  name: "node-cpu-capacity-raw"
   labels:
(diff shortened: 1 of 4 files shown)


charts/reporting-operator/templates/custom-resources/report-queries/node-cpu.yaml
```diff
@@ -45,7 +45,53 @@ spec:
...
+    GROUP BY node, resource_id
+    ORDER BY node, resource_id ASC, node_capacity_cpu_core_seconds DESC
+
+---
+
+apiVersion: metering.openshift.io/v1alpha1
+kind: ReportGenerationQuery
+metadata:
+  name: "node-cpu-allocatable-raw"
   labels:
```
EmilyM1: We have request other places. Are allocatable and request the same use case?

chancez: No. Allocatable is the amount of a resource a node has available to be requested.

PR #479, opened 2018-11-10
pkg/operator: Do not immediately import after creating ReportDataSource table

chancez:
This allows other ReportDataSources to be processed when many are
submitted at the same time like during install or when a group are
created via kubectl apply.


chancez:
It is a single queue, but we have many workers (go routines) processing from it. The jitter is add a small random variation to ideally just have each query slightly staggered.

PR #480, opened 2018-11-12
charts/hdfs: Add rack awareness

chancez:
Currently this configures each pod as as a separate rack. Our default
tolerations disallow pods from running on the same node so this should
suffice.

This makes it so HDFS will actually properly do replication across datanodes. Currently it will not really attempt to replicate because without the network topology script, the nodes all run in the "default rack" which causes them to be considered within a shared failure domain, which means replication isn't done since replications is done across racks to ensure multiple replicas aren't going to be affected by a rack outage. Since we're in the cloud, in Kubernetes we're just considering each node equivalent to a rack.

### DIFF
--- a/charts/hdfs/templates/configmap.yaml
+++ b/charts/hdfs/templates/configmap.yaml
@@ -53,6 +53,14 @@ data:
         <name>dfs.datanode.data.dir.perm</name>
         <value>{{ .Values.spec.config.datanodeDataDirPerms }}</value>
       </property>
+      <property>
+        <name>dfs.replication</name>
+        <value>{{ .Values.spec.config.replicationFactor }}</value>
+      </property>
+      <property>
+        <name>net.topology.script.file.name</name>
+        <value>/hadoop-config/topology-configuration.sh</value>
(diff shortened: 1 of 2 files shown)


tschuy:
Looks good from what I can tell, but I'm not an HDFS guru :smile: What caused you to include the bash scripts? Are there docs on how to do rack awareness that call out under-replication as a major source of problems? Mostly want to make sure that it's the *only* problem-fixing bash script to include. Also, if I'm reading it correctly, you're just putting this on disk to make debugging in the future easier, correct?

chancez:
@tschuy that's how it works. If you look the doc linked in the script. You provide a script that takes a list of IPs (each datanode IP) and output the rackname for each. Usually you do some math on subnets to figure out racks.

The fix script is taken from a hortonworks knowledge base article and is how I forced existing files to be set to replicated after adding new datanodes, we can use it if we ever have to do it again. I intend for this to be usable via a kubernetes job/cronJob potentially in the future though.

chancez:
@tschuy  Added.

PR #481, opened 2018-11-12
Partition promsum tables by year-month-day

chancez:
Adds a partition column `dt` which is a timestamp formatted as a string like `'2006-01-02'` representing year-month-day.
Should improve performance when querying data.

This is a breaking change because we don't currently have this partition column, and we don't currently migrate schemas forward.


chancez:
@EmilyM1 Nope, it's not using prometheus reportDataSource data. It's using the awsBilling reportDataSource which is a different thing, and actually already uses partitions.

EmilyM1:
Ok!

PR #484, opened 2018-11-13
Support setting max query length in presto and reporting-operator

chancez:
Sets new default for maxQueryLength to 10,000,000 which is around 10Mb (ish)

### DIFF
--- a/pkg/operator/prestostore/prometheusmetric.go
+++ b/pkg/operator/prestostore/prometheusmetric.go
@@ -147,7 +147,7 @@ func StorePrometheusMetricsWithBuffer(queryBuf *bytes.Buffer, ctx context.Contex
 		newBufferSize := (bytesToWrite + queryBuf.Len())
 
 		// if writing the current metricValue to the buffer would exceed the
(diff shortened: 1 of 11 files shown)


pkg/operator/prestostore/prometheusmetric.go
```diff
@@ -130,13 +141,13 @@ func StorePrometheusMetrics(ctx context.Context, queryer db.Queryer, tableName s
...
 
-		// There's a character limit of prestoQueryCap on insert
+		// There's a character limit of bufferCapacity on insert
 		// queries, so let's chunk them at that limit.
 		bytesToWrite := len(metricValue)
 		newBufferSize := (bytesToWrite + queryBuf.Len())
 
 		// if writing the current metricValue to the buffer would exceed the
-		// prestoQueryCap, preform the insert query, and reset the buffer
+		// bufferCapacity, preform the insert query, and reset the buffer
```
pbx0: preform -> perform

chancez: Fixed in latest commit

PR #485, opened 2018-11-13
charts: Set JVM options for hdfs, hive and presto

chancez:
Use ExitOnOutOfMemoryError instead of OnOutOfMemoryError=kill -9 %p
which doesn't work in a container when the JVM is PID 1 which can't be
killed. Also updates hdfs/hive to use the G1 garbage collector for consistency 
with Presto and be verbose when there are GC pauses or errors. Also enables 
JMX options which provide GC info/metrics. Taken from hadoop-env.sh and hive-env.sh.


chancez:
I tested this and verified it does cause the JVM to properly exit.

PR #492, opened 2018-11-16
Documentation,manifests: Document exposing reporting API outside of cluster

chancez:
Adds documentation on how to expose the reporting API so that you can query reports outside of the cluster.
Also documents how to enable the openshift oauth proxy to protect the API when exposing it.

### DIFF
--- a/manifests/metering-config/expose-route.yaml
+++ b/manifests/metering-config/expose-route.yaml
@@ -1,0 +1,37 @@
+apiVersion: metering.openshift.io/v1alpha1
+kind: Metering
+metadata:
+  name: "operator-metering"
+spec:
+  reporting-operator:
+    spec:
+      route:
+        enabled: true
+
+      authProxy:
+        enabled: true
+
+        # htpasswdData can contain htpasswd file contents for allowing auth
+        # using a static list of usernames and their password hashes.
+        #
+        # username is 'testuser' password is 'testpassword123'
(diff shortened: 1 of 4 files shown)


chancez:
With #491 merged, I'm trying to decide if this could be pulled into a separate document. I feel like this plus the prometheus URL config _could_ go into a new `configuring-reporting-operator.md` document which we link from both the `metering-config.md` and from `common-configuration.md` since this is such a common piece. 

This is making me want auto-generated table of contents for markdown.

manifests/metering-config/expose-route.yaml
```diff
@@ -0,0 +1,37 @@
...
+        #   testuser:{SHA}y/2sYAj5yrQIN4TL0YdPdmGNKpc=
+        #
+        # change REPLACEME to the output of your htpasswd command
+        htpasswdData: |
+          REPLACEME
+
+        # cookieSeed is used to protect the cookie created if accessing the API
+        # via your browser.
+        #
+        # generate a 32 character random string using `pwgen 32`
```
pruan-rht: FYI, `pwgen` is not included in the default openshift installation as well as Mac/Windows.  I have a better luck using `openssl rand -base64 32` . Perhaps we can use that suggestion instead

chancez: Feel free to open a PR

pruan-rht: ok, opened #497

PR #501, opened 2018-11-20
pkg,Documentation: Support more than 1 type for ReportGenerationQuery inputs

chancez:
Support more than string inputs, and document the inputs field of
queries.


chancez:
Just tested this again using the following:
```
kind: ReportGenerationQuery
metadata:
  name: "test-query"
spec:
  columns:
  - name: test
    type: int
  view:
    disabled: true
  inputs:
  - name: "testValue"
    type: int
  query: |
      SELECT {| .Report.Inputs.testValue |} as test
```

```
kind: Report
metadata:
  name: test-query
spec:
  generationQuery: "test-query"
  reportingStart: "2018-12-08T00:00:00Z"
  reportingEnd: "2018-12-09T00:00:00Z"
  runImmediately: true
  inputs:
  - name: testValue
    value: 1
```

Then after creating it I queried the result out (running reporting-operator locally):
```
curl -k "127.0.0.1:8080/api/v1/reports/get?name=test-query&format=tab"
test
1
```

Which matches the custom input value for the query.

PR #516, opened 2018-11-28
reporting-operator: add pvc-usage and capacity

pbx0:
This is the basic I'm using to figure out my non-raw rgq. I was using pod-memory as one of the examples to look at.

### DIFF
--- a/charts/reporting-operator/templates/custom-resources/report-queries/persistentvolumeclaim-capacity.yaml
+++ b/charts/reporting-operator/templates/custom-resources/report-queries/persistentvolumeclaim-capacity.yaml
@@ -1,0 +1,44 @@
+apiVersion: metering.openshift.io/v1alpha1
+kind: ReportGenerationQuery
+metadata:
+  name: "persistentvolumeclaim-capacity-raw"
+  labels:
+    operator-metering: "true"
+{{- block "extraMetadata" . }}
+{{- end }}
+spec:
+  reportDataSources:
+  - "persistentvolumeclaim-capacity-bytes"
+  columns:
+  - name: namespace
+    type: string
+  - name: persistentvolumeclaim
+    type: string
+  - name: labels
+    type: map<string, string>
+    tableHidden: true
+  - name: persistentvolumeclaim_capacity_bytes
+    type: double
+    unit: bytes
+  - name: timeprecision
+    type: double
+    unit: seconds
+  - name: persistentvolumeclaim_capacity_byte_seconds
+    type: double
+    unit: byte_seconds
+  - name: timestamp
+    type: timestamp
+    unit: date
+  - name: dt
+    type: string
+  query: |
+    SELECT
+        labels['exported_namespace'] as namespace,
+        labels['persistentvolumeclaim'] as persistentvolumeclaim,
+        labels,
+        amount as persistentvolumeclaim_capacity_bytes,
+        timeprecision,
+        amount * timeprecision as persistentvolumeclaim_capacity_byte_seconds,
+        "timestamp",
+        dt
+    FROM {| dataSourceTableName "persistentvolumeclaim-capacity-bytes" |}
--- a/charts/reporting-operator/templates/custom-resources/report-queries/persistentvolumeclaim-usage.yaml
+++ b/charts/reporting-operator/templates/custom-resources/report-queries/persistentvolumeclaim-usage.yaml
@@ -1,0 +1,44 @@
+apiVersion: metering.openshift.io/v1alpha1
+kind: ReportGenerationQuery
+metadata:
+  name: "persistentvolumeclaim-usage-raw"
+  labels:
+    operator-metering: "true"
+{{- block "extraMetadata" . }}
+{{- end }}
+spec:
+  reportDataSources:
+  - "persistentvolumeclaim-usage-bytes"
+  columns:
+  - name: namespace
+    type: string
+  - name: persistentvolumeclaim
+    type: string
+  - name: labels
+    type: map<string, string>
+    tableHidden: true
+  - name: persistentvolumeclaim_usage_bytes
+    type: double
+    unit: bytes
+  - name: timeprecision
+    type: double
+    unit: seconds
+  - name: persistentvolumeclaim_usage_byte_seconds
+    type: double
+    unit: byte_seconds
+  - name: timestamp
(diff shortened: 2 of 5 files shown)


charts/reporting-operator/templates/custom-resources/prom-queries/pvc-capacity.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: metering.openshift.io/v1alpha1
+kind: ReportPrometheusQuery
+metadata:
+  name: pvc-capacity-bytes
+spec:
+  query: |
+    max(kubelet_volume_stats_capacity_bytes) by (exported_namespace, persistentvolumeclaim)
```
chancez: We don't actually want the max. We want the value at every moment in time, and we want all the metric labels, so this can just be `kubelet_volume_stats_capacity_bytes`.

charts/reporting-operator/templates/custom-resources/prom-queries/pvc-usage.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: metering.openshift.io/v1alpha1
+kind: ReportPrometheusQuery
+metadata:
+  name: pvc-usage-bytes
+spec:
+  query: |
+    max(kubelet_volume_stats_used_bytes) by (exported_namespace, persistentvolumeclaim)
```
chancez: Same here, this can just be `kubelet_volume_stats_used_bytes`

charts/reporting-operator/templates/custom-resources/report-queries/pvc-capacity.yaml
```diff
@@ -0,0 +1,37 @@
...
+    type: timestamp
+    unit: date
+  query: |
+    SELECT
+        labels['exported_namespace'] as namespace,
+        labels['persistentvolumeclaim'] as persistantvolumeclaim,
+        amount as pvc_capacity_bytes,
+        timeprecision,
+        amount * timeprecision as pvc_used_byte_seconds,
+        "timestamp",
```
chancez: Don't forget to select `dt` and add it to your columns as a string. This is the partition for the data, and other queries querying the view for this raw query can filter on `dt` to improve performance. The partition is just a string in the form of `2018-11-28` which is the day for the metric.

charts/reporting-operator/templates/custom-resources/report-queries/pvc-capacity.yaml
```diff
@@ -0,0 +1,37 @@
...
+  - name: timestamp
+    type: timestamp
+    unit: date
+  query: |
+    SELECT
+        labels['exported_namespace'] as namespace,
+        labels['persistentvolumeclaim'] as persistantvolumeclaim,
+        amount as pvc_capacity_bytes,
+        timeprecision,
+        amount * timeprecision as pvc_used_byte_seconds,
```
chancez: This column name doesn't match the one in the `columns` list. Used should be capacity.

charts/reporting-operator/templates/custom-resources/prom-queries/pvc-capacity.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: metering.openshift.io/v1alpha1
+kind: ReportPrometheusQuery
+metadata:
+  name: pvc-capacity-bytes
```
chancez: Can you use `persistentvolumeclaim-` as the prefix in the name for consistency with how we name the other queries?

charts/reporting-operator/templates/custom-resources/prom-queries/pvc-usage.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: metering.openshift.io/v1alpha1
+kind: ReportPrometheusQuery
+metadata:
+  name: pvc-usage-bytes
```
chancez: Can you use `persistentvolumeclaim-` as the prefix in the name for consistency with how we name the other queries?

charts/reporting-operator/values.yaml
```diff
@@ -108,6 +108,14 @@ spec:
         spec:
           promsum:
             query: "node-capacity-cpu-cores"
+      pvc-capacity-bytes:
```
chancez: The key here is the datasources name, so you'll want to change this to `persistentvolumeclaim-usage-bytes` also, since that's what's used in the `reportDataSources` list of your reportGenerationQuery.

charts/reporting-operator/values.yaml
```diff
@@ -108,6 +108,14 @@ spec:
...
+        spec:
+          promsum:
+            query: "persistentvolumeclaim-capacity-bytes"
+      pvc-usage-bytes:
```
chancez: Same here

pbx0:
PTAL

charts/reporting-operator/templates/custom-resources/report-queries/persistentvolumeclaim-usage.yaml
```diff
@@ -0,0 +1,44 @@
...
+    unit: byte_seconds
+  - name: timestamp
+    type: timestamp
+    unit: date
+  - name: dt
+    type: string
+  query: |
+    SELECT
+        labels['exported_namespace'] as namespace,
+        labels['persistentvolumeclaim'] as persistantvolumeclaim,
```
chancez: `persistantvolumeclaim` should be `persistentvolumeclaim`

charts/reporting-operator/templates/custom-resources/report-queries/persistentvolumeclaim-capacity.yaml
```diff
@@ -0,0 +1,44 @@
...
+    unit: byte_seconds
+  - name: timestamp
+    type: timestamp
+    unit: date
+  - name: dt
+    type: string
+  query: |
+    SELECT
+        labels['exported_namespace'] as namespace,
+        labels['persistentvolumeclaim'] as persistantvolumeclaim,
```
chancez: `persistantvolumeclaim` should be `persistentvolumeclaim`

pbx0: oh, good catch

PR #517, opened 2018-11-29
Fix AWS ReportDataSource

chancez:
AWS Billing tables are external so we don't want to modify the location
by appending the table's name to the location.

The queries were broken when I previously made the node-cpu/memory-allocatable-raw
queries, which were intended to be the re-usable views, and all queries except the AWS ones
got updated to use the -raw one, so this fixes that as well.


chancez:
I've validated this by directly quering the `datasource_aws_billing` in presto and verifying it can be queried. Previously the operator failed on adding partitions, and that's succeeding and my queries can use them.

PR #518, opened 2018-11-29
Improve ReportGenerationQuery dependency validation error handling

chancez:
Use dedicated error struct and functions to allow differentiating between the
kind of error to allow handling different types of dependency validation errors
differently.

### DIFF
--- a/pkg/operator/reporting/validate.go
+++ b/pkg/operator/reporting/validate.go
@@ -40,11 +40,11 @@ func GetAndValidateGenerationQueryDependencies(
 		generationQuery,
 	)
 	if err != nil {
-		return nil, fmt.Errorf("unable to get dependencies for ReportGenerationQuery %s: %v", generationQuery.Name, err)
+		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


pkg/operator/reporting/validate.go
```diff
@@ -127,12 +105,61 @@ func ValidateGenerationQueryDependencies(deps *ReportGenerationQueryDependencies
...
+		len(validationErr.uninitializedDataSourceNames) != 0 ||
+		len(validationErr.uninitializedReportNames) != 0 ||
+		len(validationErr.uninitializedScheduledReportNames) != 0 {
+		return validationErr
 	}
 	return nil
 }
 
+func IsUninitializedDependencyError(err error) bool {
+	validationErr, ok := err.(*reportGenerationQueryDependenciesValidationError)
```
tschuy: Slightly confused as to how this works. Maybe my golang is a bit rusty?

chancez: It's a type assertion: https://golang.org/doc/effective_go.html#interface_conversions

PR #523, opened 2018-11-29
charts,manifests,Documentation: Move metering CRs to openshift-reporting and openshift-metering charts

chancez:
Overall this reorganizes our charts by moving the custom resources the
report-operator manages into their own chart, allows for a more minimal
base installations of reporting-operator itself which can be helpful
when debugging or developing new sets of reporting related resources.

openshift-reporting has all the default ReportPrometheusQueries,
ReportDataSources, and ReportGenerationQueries for use with openshift
cluster monitoring, and is a subchart dependency of openshift-metering
which is enabled by default.

openshift-metering now controls the default storage location resource
created that previous was created in the reporting-operator chart.

In the future we will have a telemeter-metering chart which will have a
different set of default values and does not use hdfs by default.


chancez:
Note: this will break existing configurations using custom storage or aws billing correlation. Updating is fairly simple however. That said, the ansible role to install metering has a task to configure S3 as defaultStorage which will no longer work correctly, and a custom metering config will need to be added to set the defaultStorage appropriately.

PR #527, opened 2018-12-03
charts/presto: Configure hive server and metastore readiness/liveness probes

chancez:
This should at least cause the pods to restart if they get to a point where we're unable to talk to the thrift RPC port. Overall it's still a very simple liveness and readiness check that we should look to improve, but I believe something here is better than nothing.


chancez:
This causes hive to crash in CI but not when I run the same manually so need to debug further...

PR #529, opened 2018-12-04
pkg/operator: Don't add comma unless a new metric is being added to buffer

chancez:
The previously logic appended a comma when VALUES wasn't written to the
buffer, but this was broken when nothing else can be written due to the
size of the metric.

Also avoids running INSERT INTO without any values which could occur
previously if the buffer was too small.


chancez:
Had some logic bugs previously. Mainly that the last metric in the list wasn't being written.

chancez:
@tschuy ready for another review. This time I had to refactor things a bit more than before since the logic wasn't quite as well organized for it before.

PR #532, opened 2018-12-11
Add hive.metastore-timeout to the hive catalog properties

miguel-santiago:
While running reporting-operator, we noticed that the operator-metering is not ready and gets restarted a lot of times due to Hive timeouts to Presto, this is the status of a reporting-operator container without hive timeout running for 4 days:
```
    - name: reporting-operator
      state:
        running:
          startedAt: '2018-12-11T10:34:40Z'
      lastState:
        terminated:
          exitCode: 137
          reason: Error
          startedAt: '2018-12-11T10:24:50Z'
          finishedAt: '2018-12-11T10:29:37Z'
          containerID: >-
            docker://ad4a32eac6faec0d689f297510ffd23834945e8e86200d18d0b776605e34dd26
      ready: false
      restartCount: 674
```

This is because when the operator tries to write the health_check, Hive returns a timeout to Presto (even if Hive writes the health_check correctly). Adding the timeout property fixes the problem and Hive doesn't return a timeout anymore. My guess is that the default value for the timeout is really low or is just being ignored.

Just adding the timeout fixes the problem, this is the status of a different reporting-operator pod running for 4 days with the timeout set:
```
    - name: reporting-operator
      state:
        running:
          startedAt: '2018-12-10T18:32:26Z'
      lastState:
        terminated:
          exitCode: 1
          reason: Error
          startedAt: '2018-12-10T18:27:14Z'
          finishedAt: '2018-12-10T18:32:25Z'
          containerID: >-
            docker://504f40c189708f1535c29e15f30d9d219d987acc6f47155bba53a220b1bdec3e
      ready: true
      restartCount: 1
```

Hope this helps!


chancez:
Oh I misread your original message, you're entirely right it could be too low for the current amount of resources we give things. I guess I see this a lot less frequently since I'm provisioning more resource for things. I have a few comments, but I'll add them in the review.

chancez:
Also, are you seeing any of the other pods restart?

miguel-santiago:
Hi @chancez, no I don't see any other pods restarting, just the operator-metering one. 

We have operator-metering running in 5 different clusters, and at the beginning we didn't have any timeout problems, but after some weeks running and generating reports, it started happening every day more frequently... until it was basically impossible to get the operator to a ready state.

I applied the change to all the clusters yesterday and they didn't have a restart since then. 

I can update the PR if you want to add the timeout only if we specifically define it in the values (right now with my change it will always be added), but I think it's quite important to be able to define it to avoid problems with timeouts, for us would be amazing cause right now we are using our custom docker images with the updated chart to deploy the operator :)

miguel-santiago:
Ok, this way the timeout will only be added to the secret if we define a value for metastoreTimeout.

I left 20 seconds in the comment as example.

By the way, to answer your question, with 20 seconds my problem was fixed too, I just went for a big value like 1 minute just to be sure that it won't fail for this reason during the holidays :)

chancez:
Can you squash your commits? After that I can approve and merge.

Also, I did end up setting up another env which was able to reproduce your issues. So far it seems somewhat related to resources. I need to investigate more, but I think it's probably JVM pauses.

miguel-santiago:
Done :)

chancez:
@miguel-santiago I missed this originally, but this should have gone under the `spec.presto.config` section. #572 is going to change where we get the value from, so if you update to 0.13.0-latest, or beyond know that you'll need to adjust where you specify the timeout value for presto.

PR #546, opened 2019-01-02
Improve authorization support

chancez:
- Adds support for creating rolebindings granting access to various metering resources via roles.
- Changes permissions required to query reporting API with authentication enabled from GET namespaces to GET reports/export sub-resource.
- Adds support for granting permission to reporting API by creating a rolebinding which grants the reports/export GET permission.

### DIFF
--- a/Documentation/configuring-reporting-operator.md
+++ b/Documentation/configuring-reporting-operator.md
@@ -41,9 +41,9 @@ Additionally, on Openshift:
 - We deploy the [Openshift OAuth proxy][oauth-proxy] as a side-car container for reporting-operator, which protects the reporting API with authentication
 
 There are a few ways to do authentication, you can use service account tokens for authentication, and/or you can also use a static username/password via an httpasswd file.
-
+See the [openshift authentication](#openshift-authentication) section below for details on how authentication and authorization works.
 See the [expose-route.yaml][expose-route-config] configuration for an example of setting enabling an Openshift route and configuring authentication with both options enabled.
-See the [openshift authentication][#openshift-authentication] section below for details on how authentication is done.
+
 Make sure you modify the `reporting-operator.spec.authProxy.httpasswdData` and `reporting-operator.spec.authProxy.cookieSeed` values.
 
 Once installed with the customized configuration to enable the route, you should query in your namespace to check for the route:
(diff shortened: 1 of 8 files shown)


Documentation/configuring-reporting-operator.md
```diff
@@ -70,6 +71,26 @@ And to authenticate using a username and password, use basic authentication:
 curl -u testuser:password123 -k "https://metering-openshift-metering.apps.example.com/api/v1/reports/get?name=cluster-memory-capacity-hourly&format=tab"
 ```
 
+#### Openshift Authentication
+
+When the following options are set to true, it enables authenticating using a bearer token from a serviceAccount or for your user when querying the reporting rest API:
```
pbx0: Slighty awkward sentence. Maybe try "When the following options are set to true, authentication using a bearer token is enabled for the reporting rest API. Bearer tokens may come from serviceAccounts or users."

manifests/metering-config/expose-route.yaml
```diff
@@ -33,6 +33,11 @@ spec:
         # change REPLACEME to the output of your command
         cookieSeed: "REPLACEME"
 
-        # enables authenticating using a serviceAccount token
+        # Enables authenticating using a bearer token from a serviceAccount or
+        # for your user.
```
pbx0: `serviceAccount or user` to eliminate second person perspective.

Documentation/configuring-reporting-operator.md
```diff
@@ -129,10 +109,47 @@ In this example the externalIP of the LoadBalancer is `35.227.172.86` and the po
 curl "http://35.227.172.86:8080/api/v1/reports/get?name=cluster-memory-capacity-hourly&format=tab"
 ```
 
+### Openshift Authentication
+
+Authentication can be enabled by setting the options below to true.
+Enabling authentication configures the reporting-operator pod to run the Openshift auth-proxy as a sidecar container in the pod, adjusting the ports so that the reporting-operator API isn't exposed directly, but instead using the auth-proxy pod.
```
pbx0: Maybe break-up sentence into two. Something like:
```
... Openshift auth-proxy as a sidecar container in the pod. The sidecar adjusts the ports so that the reporting-operator API isn't exposed directly, but instead uses the auth-proxy pod.
```

Documentation/configuring-reporting-operator.md
chancez: Yeah this sounds way better

manifests/metering-config/expose-route.yaml
chancez: Done

Documentation/configuring-reporting-operator.md
chancez: Good idea. I made an adjustment similar to your suggestion. How does it look now?

pbx0: Looks good now!

PR #556, opened 2019-01-07
Update the Metering side for OLM inclusion

EmilyM1:


charts/metering-alm/templates/package.yaml
```diff
@@ -0,0 +1,4 @@
+#! package-manifest: deploy/chart/catalog_resources/rh-operators/metering-operator.v0.12.0.clusterserviceversion.yaml
```
chancez: This needs to use the `{{ .Values.packageName }}` and `{{ .Values.spec.version }}` values to template in the file name, like the original package-v1.yaml did

charts/metering-alm/templates/subscription.yaml
```diff
@@ -1,8 +1,9 @@
-apiVersion: app.coreos.com/v1alpha1
-kind: Subscription-v1
+apiVersion: operators.coreos.com/v1alpha1
+kind: Subscription
 metadata:
   name: {{ .Values.name }}
+  namespace: local
```
chancez: Remove this. Namespace can be controlled with `--namespace` in `kubectl create`

manifests/deploy/common-alm-values.yaml
```diff
@@ -1,20 +1,318 @@
 name: metering-operator.vlatest
+labels:
```
chancez: No need to add `labels` to this file. It isn't a kubernetes manifest. Just a values file for helm.

manifests/deploy/common-alm-values.yaml
```diff
@@ -1,20 +1,318 @@
...
+  installModes:
+      - type: OwnNamespace
+        supported: true
+      - type: SingleNamespace
+        supported: false
+      - type: MultiNamespace
+        supported: false
+      - type: AllNamespaces
+        supported: false
+  permissions:
```
chancez: So permissions is actually not supposed to be set in this file, because we set them somewhere else, and use the output of another file we generate, as the permissions value for this helm chart. It works like this:

So permissions actually is extracted from a different file, a file that is also generated using helm charts. So it works like this:

- https://github.com/operator-framework/operator-metering/blob/master/hack/create-alm-manifests.sh takes the `rules` from the `metering-operator-role.yaml` https://github.com/operator-framework/operator-metering/blob/master/manifests/deploy/openshift/helm-operator/metering-operator-role.yaml#L5 and uses them for the permissions value for this chart when it generates the manifests.
- The `metering-operator-role.yaml` file, is a file generated via the helm-operator chart via this script: https://github.com/operator-framework/operator-metering/blob/master/hack/create-deploy-manifests.sh. create-deploy-manifests.sh uses https://github.com/operator-framework/operator-metering/blob/master/manifests/deploy/common-helm-operator-values.yaml#L30 as the values for the permissions for the helm-operator chart.

So the way that ends up working is:
- make metering-manifests runs `./hack/create-metering-manfests.sh`
- `./hack/create-metering-manfests.sh` runs `./hack/create-deploy-manifests.sh` which uses https://github.com/operator-framework/operator-metering/blob/master/manifests/deploy/common-helm-operator-values.yaml and produces the files in https://github.com/operator-framework/operator-metering/tree/master/manifests/deploy/openshift/helm-operator
- `./hack/create-metering-manfests.sh` then runs `./hack/create-alm-manifests.sh` which uses the https://github.com/operator-framework/operator-metering/blob/master/manifests/deploy/common-alm-values.yaml, and the `manifests/deploy/openshift/helm-operator` files and produces the files in https://github.com/operator-framework/operator-metering/tree/master/manifests/deploy/openshift/alm

chancez:
Can you squash your commits?

PR #574, opened 2019-02-07
PVC Storage Used ReportGenerationQuery/Report

EmilyM1:


charts/openshift-reporting/templates/report-queries/persistentvolumeclaim-usage.yaml
```diff
@@ -42,3 +87,87 @@ spec:
...
+    type: string
+  - name: phase_value
+    type: double
+  - name: persistentvolumeclaim_usage_bytes
+    type: double
+    unit: bytes
+  - name: dt
+    type: string
+  query: |
+    SELECT phase.namespace, phase.timestamp, phase.persistentvolumeclaim, phase.phase, phase.value AS phase_value, persistentvolumeclaim_usage_bytes,phase.dt
```
chancez: Can you put each column on a separate line like we do in other queries?

charts/openshift-reporting/templates/report-queries/persistentvolumeclaim-usage.yaml
```diff
@@ -42,3 +87,87 @@ spec:
...
+    FROM {| generationQueryViewName "persistentvolumeclaim-phase-raw" |} AS phase
+    FULL OUTER JOIN
+    {| generationQueryViewName "persistentvolumeclaim_usage_raw" |} AS usage
```
chancez: These `_` should be `-`. It doesn't error because we have a naive mapping, but this could break in the future. It must match the value in `spec.reportDataSources`

charts/openshift-reporting/templates/report-queries/persistentvolumeclaim-usage.yaml
```diff
@@ -42,3 +87,87 @@ spec:
...
+  - name: phase_value
+    type: double
+  - name: persistentvolumeclaim_usage_bytes
+    type: double
+    unit: bytes
+  inputs:
+  - name: ReportingStart
+  - name: ReportingEnd
+  query: |
+    SELECT namespace, timestamp, persistentvolumeclaim, phase, phase_value, persistentvolumeclaim_usage_bytes
```
chancez: Can these columns each be on their own line?

chancez:
This looks great with some minor formatting updates it looks correct.
We need to add e2e tests and integration tests however for this to be complete.

charts/openshift-reporting/values.yaml
```diff
@@ -45,10 +45,10 @@ spec:
...
         promsum:
           query: "persistentvolumeclaim-usage-bytes"
-    persistentvolumeclaim-phase:                                                                                                          
-        spec:                                                                                                                               
-          promsum:                                                                                                                          
-            query: "persistentvolumeclaim-phase" 
+    persistentvolumeclaim-phase:
+      spec:
+        promsum:
+          query: "persistentvolumeclaim-phase"
```
tschuy: trim whitespace

charts/openshift-reporting/templates/report-queries/persistentvolumeclaim-usage.yaml
```diff
@@ -42,3 +87,98 @@ spec:
...
+  query: |
+    SELECT
+      phase.namespace,
+      phase.timestamp,
+      phase.persistentvolumeclaim,
+      phase.phase, phase.value AS phase_value, persistentvolumeclaim_usage_bytes,
+      phase.dt
+    FROM {| generationQueryViewName "persistentvolumeclaim-phase-raw" |} AS phase
+    FULL OUTER JOIN
+    {| generationQueryViewName "persistentvolumeclaim-usage-raw" |} AS usage
```
chancez: nit: this should be on the same line as the `FULL OUTER JOIN` text

charts/openshift-reporting/templates/report-queries/persistentvolumeclaim-usage.yaml
```diff
@@ -42,3 +87,97 @@ spec:
...
+apiVersion: metering.openshift.io/v1alpha1
+kind: ReportGenerationQuery
+metadata:
+  name: "persistentvolumeclaim-usage-with-phase-raw"
+  labels:
+    operator-metering: "true"
+{{- block "extraMetadata" . }}
+{{- end }}
+spec:
+  reportDataSources:
```
chancez: This list is incorrect. This query uses 2 reportGenerationQueries, and the names here don't match either.

PR #586, opened 2019-02-21
olm: update CSV description

robszumski:
- update description
- improve capitalization and descriptions of CRDs
- reorder CRDs so the most used ones are first


robszumski:
Renders to:

image

PR #590, opened 2019-02-25
reporting-operator: add ability to configure CA

tschuy:


pkg/operator/operator.go
```diff
@@ -605,10 +604,18 @@ func (op *Reporting) newPrometheusConnFromURL(url string) (prom.API, error) {
 
 	transportConfig := *kubeTransportConfig
 
-	if _, err := os.Stat(serviceServingCAFile); err == nil {
-		// use the service serving CA for prometheus
-		transportConfig.TLS.CAFile = serviceServingCAFile
-		op.logger.Infof("using %s as CA for Prometheus", serviceServingCAFile)
+	if op.cfg.PrometheusDataSourceCAFile != "" {
+		if _, err := os.Stat(op.cfg.PrometheusDataSourceCAFile); err == nil {
+			// use the service serving CA for prometheus
```
chancez: `// Use the configured CA for communicating to Prometheus`

pkg/operator/operator.go
```diff
@@ -605,10 +604,18 @@ func (op *Reporting) newPrometheusConnFromURL(url string) (prom.API, error) {
 
 	transportConfig := *kubeTransportConfig
```
chancez: I think we want to create a new transportConfig instead of using the kube one. The kube one isn't getting us anything anymore I don't think.

pkg/operator/operator.go
```diff
@@ -108,6 +107,7 @@ type Config struct {
 	PrometheusDataSourceMaxQueryRangeDuration     time.Duration
 	PrometheusDataSourceMaxBackfillImportDuration time.Duration
 	PrometheusDataSourceGlobalImportFromTime      *time.Time
+	PrometheusDataSourceCAFile                    string
```
chancez: I feel like this should be on `PrometheusConfig`

cmd/reporting-operator/start.go
```diff
@@ -73,6 +73,8 @@ func init() {
 	startCmd.Flags().StringVar(&cfg.PrometheusConfig.Address, "prometheus-host", defaultPromHost, "the URL string for connecting to Prometheus")
 	startCmd.Flags().BoolVar(&cfg.PrometheusConfig.SkipTLSVerify, "prometheus-skip-tls-verify", false, "Skip TLS verification")
 	startCmd.Flags().StringVar(&cfg.PrometheusConfig.BearerToken, "prometheus-bearer-token", "", "Bearer token to authenticate against Prometheus.")
+	startCmd.Flags().StringVar(&cfg.PrometheusConfig.BearerTokenFile, "prometheus-bearer-token-file", "", "File containing bearer token to authenticate against Prometheus. Takes precedence over prometheus-bearer-token.")
+	startCmd.Flags().StringVar(&cfg.PrometheusConfig.CAFile, "prometheus-datasource-ca-file", "", "The path to the certificate authority to use to connect to the Prometheus data source. If empty, defaults to system CAs")
```
chancez: The name of the flag here should be updated to match, same with the env vars.

chancez: `prometheus-ca-file`, etc

PR #601, opened 2019-02-27
reporting-operator: use secret for token config

tschuy:
I'm not particularly happy with the configuration structure for bearer tokens now that we have three options, two intertwined, for bearer config. 

Would it be better perhaps to refactor into something more like the TLS config:

```
prometheusBearer:
    # use either tokenFile or token; file takes precedence
    tokenFile: "/var/run/secrets/kubernetes.io/serviceaccount/token" 
    token: null # "my_token_here"
    createSecret: false
    tokenSecretName: reporting-operator-prometheus-bearer-secrets
```
?

I'm also not particularly happy with that, since it still leads to awkward-feeling options between the tokenFile and the rest of the options that are specifically about the secret.

Thoughts?

### DIFF
--- a/Documentation/configuring-reporting-operator.md
+++ b/Documentation/configuring-reporting-operator.md
@@ -3,7 +3,7 @@
 reporting-operator is responsible for collecting data from Prometheus, storing the metrics in Presto, running report queries against Presto, and exposing their results via an HTTP API.
 Configuring the operator is done primarily within a `Metering` CR's `spec.reporting-operator.spec` section.
 
-## Prometheus URL
+## Prometheus connection
 
 Depending on how you installed Metering, the default Prometheus URL varies.
 If you installed for Openshift then the default assumes Prometheus is available at `https://prometheus-k8s.openshift-monitoring.svc:9091/`.
@@ -20,7 +20,33 @@ spec:
         prometheusURL: "http://prometheus.cluster-monitoring.svc:9090"
 ```
 
-> Note: currently we do not support https connections or authentication to Prometheus except for in Openshift, but support for it is being developed.
+To secure the connection to Prometheus, the default Metering installation uses the Openshift certificate authority. If your Prometheus instance uses a different CA, the path to the CA file can be referenced with the `prometheusCAFile` option:
+
+```
+spec:
+  reporting-operator:
+    spec:
+      config:
+        prometheusCAFile: "/path/to/my/ca.crt"
+```
+
+Alternatively, to use the system certificate authorities for publicly valid certificates, it can be set to `""`.
+
+Reporting-operator can also be configured to use a specified bearer token to auth with Prometheus:
+
+```
+spec:
+  reporting-operator:
+    spec:
+      config:
+        prometheusImporter:
+          auth:
+            useServiceAccountToken: false
+            tokenSecret:
+              enabled: true
+              create: true
+              value: "abc-123"
+```
 
 ## Exposing the reporting API
 
--- a/charts/reporting-operator/templates/reporting-operator-config.yaml
+++ b/charts/reporting-operator/templates/reporting-operator-config.yaml
@@ -40,8 +40,10 @@ data:
 {{- if .Values.spec.config.prometheusDatasourceImportFrom }}
   prometheus-datasource-import-from: {{ .Values.spec.config.prometheusDatasourceImportFrom | quote }}
 {{- end }}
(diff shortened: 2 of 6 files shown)


charts/reporting-operator/templates/reporting-operator-deployment.yaml
```diff
@@ -156,12 +156,21 @@ spec:
               name: reporting-operator-config
               key: prometheus-datasource-import-from
               optional: true
+{{- /* require tokenSecret XOR serviceAccount */ }}
+{{- required (.Values.spec.config.prometheusImporter.auth.tokenSecret.value || .Values.spec.config.prometheusImporter.auth.useServiceAccountToken) && !(.Values.spec.config.prometheusImporter.auth.tokenSecret.value || .Values.spec.config.prometheusImporter.auth.useServiceAccountToken)  }}
```
chancez: You probably want this in separate `if` conditions so you can set an error message based on which case your hitting. Currently you have no error message for `required`.

tschuy:
I've verified, with this latest change, that this will indeed fail when useServiceAccount is set to true and token value is set.

charts/reporting-operator/templates/reporting-operator-config.yaml
```diff
@@ -40,8 +40,8 @@ data:
 {{- if .Values.spec.config.prometheusDatasourceImportFrom }}
   prometheus-datasource-import-from: {{ .Values.spec.config.prometheusDatasourceImportFrom | quote }}
 {{- end }}
-{{- if .Values.spec.config.prometheusBearerTokenFile }}
-  prometheus-bearer-token-file: "{{ .Values.spec.config.prometheusBearerTokenFile }}"
+{{- if .Values.spec.config.prometheusImporter.auth.useServiceAccountToken }}
+  prometheus-bearer-token-file: "/var/run/secrets/kubernetes.io/serviceaccount/token"
```
chancez: This should point to the volume mount for the tokenSecret if this isn't true.

charts/reporting-operator/templates/reporting-operator-deployment.yaml
```diff
@@ -156,12 +156,21 @@ spec:
...
               optional: true
+{{- /* neither specified = no auth used; both specified = error; either = correct & authenticated */ -}}
+{{- if .Values.spec.config.prometheusImporter.auth.tokenSecret.enabled }}  {{- if .Values.spec.config.prometheusImporter.auth.useServiceAccountToken  }} {{ required "cannot use both token from secret and token from service account" null }} {{- end}} {{- end }}
         - name: REPORTING_OPERATOR_PROMETHEUS_BEARER_TOKEN_FILE
           valueFrom:
             configMapKeyRef:
               name: reporting-operator-config
               key: prometheus-bearer-token-file
               optional: true
+{{- if .Values.spec.config.prometheusImporter.auth.tokenSecret.enabled }}
```
chancez: We decided on use a volumeMount for `tokenSecret` so this env var doesn't need to be set, instead we should be mounting the file and setting `prometheus-bearer-token-file` in the configmap to the location of the file

charts/reporting-operator/templates/reporting-operator-deployment.yaml
```diff
@@ -156,6 +156,10 @@ spec:
               name: reporting-operator-config
               key: prometheus-datasource-import-from
               optional: true
+{{- /* neither specified = no auth used; both specified = error; either = correct & authenticated */ -}}
+{{- if and .Values.spec.config.prometheusImporter.auth.tokenSecret.enabled .Values.spec.config.prometheusImporter.auth.useServiceAccountToken  }}
+  {{ required "cannot use both token from secret and token from service account" nil }}
```
chancez: You can use `fail` instead of `required` since you're not using the 2nd argument value

Documentation/configuring-reporting-operator.md
```diff
@@ -3,7 +3,7 @@
 reporting-operator is responsible for collecting data from Prometheus, storing the metrics in Presto, running report queries against Presto, and exposing their results via an HTTP API.
 Configuring the operator is done primarily within a `Metering` CR's `spec.reporting-operator.spec` section.
 
-## Prometheus URL
+## Prometheus
```
chancez: Do you think this should be `Prometheus Connection` ?

chancez:
Minor nits left. Can you squash once you make the changes as well?

PR #609, opened 2019-03-05
telemeter-reporting: add basic telemeter chart

tschuy:
This pull request adds:
* telemeter-reporting chart
    * chart with telemeter-specific queries
* documentation on setting up Metering to use Telemeter and telemeter-reporting chart

### DIFF
--- a/Documentation/index.md
+++ b/Documentation/index.md
@@ -17,6 +17,7 @@
     - [storing data in s3](configuring-storage.md#storing-data-in-s3)
   - [configuring the Hive metastore](configuring-hive-metastore.md)
   - [configuring aws billing correlation for cost correlation](configuring-aws-billing.md)
+  - [configuring for use with Telemeter](configuring-telemeter.md)
 - [Using Metering](using-metering.md)
 - [Resource Tuning](tuning.md)
 - [Troubleshooting](troubleshooting-metering.md)
--- a/charts/telemeter-reporting/templates/datasources.yaml
+++ b/charts/telemeter-reporting/templates/datasources.yaml
@@ -1,0 +1,13 @@
+{{- range $name, $body := .Values.spec.defaultReportDataSources }}
+---
+apiVersion: metering.openshift.io/v1alpha1
+kind: ReportDataSource
+metadata:
+  name: "{{ $name }}"
+  labels:
+    telemeter-metering: "true"
+{{- block "extraMetadata" $ }}
+{{- end }}
+spec:
+{{ toYaml $body.spec | indent 2 }}
+{{- end }}
--- a/charts/telemeter-reporting/templates/report-queries/machine-cpu-cores-by-cluster-id.yaml
+++ b/charts/telemeter-reporting/templates/report-queries/machine-cpu-cores-by-cluster-id.yaml
@@ -1,0 +1,81 @@
+apiVersion: metering.openshift.io/v1alpha1
+kind: ReportGenerationQuery
+metadata:
+  name: "machine-cpu-cores-by-cluster-id-raw"
+  labels:
+    telemeter-metering: "true"
+{{- block "extraMetadata" . }}
+{{- end }}
+spec:
+  reportDataSources:
+  - "machine-cpu-cores-by-cluster-id"
+  columns:
+  - name: machine_cpu_cores
+    type: double
+    unit: cpu_cores
+  - name: machine_cpu_core_seconds
+    type: double
+    unit: cpu_core_seconds
+  - name: cluster_uuid
+    type: string
+    unit: uuid
+  - name: timestamp
+    type: timestamp
+    unit: date
+  - name: timeprecision
+    type: double
+    unit: seconds
+  - name: dt
+    type: string
+  query: |
+      SELECT amount as machine_cpu_cores,
+          amount * timeprecision as machine_cpu_core_seconds,
+          element_at(labels, '_id') as cluster_uuid,
+          timestamp,
+          timeprecision,
+          dt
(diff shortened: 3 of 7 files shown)


charts/telemeter-reporting/templates/report-queries/machine-cpu-cores-per-cluster.yaml
```diff
@@ -0,0 +1,84 @@
...
+metadata:
+  name: "machine-cpu-cores-per-cluster-raw"
+  labels:
+    operator-metering: "true"
+{{- block "extraMetadata" . }}
+{{- end }}
+spec:
+  reportDataSources:
+  - "machine-cpu-cores-per-cluster"
+  columns:
```
chancez: Needs `dt` added to the columns https://github.com/operator-framework/operator-metering/blob/master/charts/openshift-reporting/templates/report-queries/pod-cpu.yaml#L37-L38

charts/telemeter-reporting/templates/report-queries/machine-cpu-cores-per-cluster.yaml
```diff
@@ -0,0 +1,84 @@
...
+    SELECT
+      timestamp '{| default .Report.ReportingStart .Report.Inputs.ReportingStart| prestoTimestamp |}' AS period_start,
+      timestamp '{| default .Report.ReportingEnd .Report.Inputs.ReportingEnd | prestoTimestamp |}' AS period_end,
+      cluster_uuid,
+      min("timestamp") as data_start,
+      max("timestamp") as data_end,
+      sum(machine_cpu_core_seconds) as machine_cpu_core_seconds
+    FROM {| generationQueryViewName "machine-cpu-cores-per-cluster-raw" |}
+    WHERE "timestamp" >= timestamp '{| default .Report.ReportingStart .Report.Inputs.ReportingStart | prestoTimestamp |}'
+    AND "timestamp" < timestamp '{| default .Report.ReportingEnd .Report.Inputs.ReportingEnd | prestoTimestamp |}'
```
chancez: Filter on `dt`

charts/telemeter-reporting/templates/datasources.yaml
```diff
@@ -0,0 +1,13 @@
+{{- range $name, $body := .Values.spec.defaultReportDataSources }}
+---
+apiVersion: metering.openshift.io/v1alpha1
+kind: ReportDataSource
+metadata:
+  name: "{{ $name }}"
+  labels:
+    operator-metering: "true"
```
chancez: Lets use `telemeter-metering: "true"`

charts/telemeter-reporting/templates/prom-queries/machine-cpu-cores-per-cluster.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: metering.openshift.io/v1alpha1
+kind: ReportPrometheusQuery
+metadata:
+  name: "machine-cpu-cores-per-cluster"
```
chancez: Maybe `machine-cpu-cores-by-cluster-id`.

Documentation/index.md
```diff
@@ -17,6 +17,7 @@
     - [storing data in s3](configuring-storage.md#storing-data-in-s3)
   - [configuring the Hive metastore](configuring-hive-metastore.md)
   - [configuring aws billing correlation for cost correlation](configuring-aws-billing.md)
+  - [configuring for use with Telemeter](configuring-telemeter.md)
 - [Using Metering](using-metering.md)
 - [Resource Tuning](tuning.md)
 - [Troubleshooting](troubleshooting-metering.md)
```
EmilyM1: Super minor... but idents appear off

tschuy: It falls under the Configuring Metering header, so it should be indented to this amount.

charts/telemeter-reporting/templates/report-queries/machine-cpu-cores-by-cluster-id.yaml
```diff
@@ -0,0 +1,89 @@
...
+    unit: cpu_core_seconds
+  inputs:
+  - name: ReportingStart
+  - name: ReportingEnd
+  query: |
+    SELECT
+      timestamp '{| default .Report.ReportingStart .Report.Inputs.ReportingStart| prestoTimestamp |}' AS period_start,
+      timestamp '{| default .Report.ReportingEnd .Report.Inputs.ReportingEnd | prestoTimestamp |}' AS period_end,
+      cluster_uuid,
+      min("timestamp") as data_start,
```
chancez: I don't think we need these `data_` columns

chancez:
This looks good. I don't see anything major. Thoughts on those `data_` columns?

charts/telemeter-reporting/templates/report-queries/machine-cpu-cores-by-cluster-id.yaml
tschuy: I based this query off of a query in the openshift-metering chart. If we don't want data_start/data_end here, do we want them in the other charts? They can be ~vaguely useful for knowing if we have gaps in our data. Since this is per-cluster, it would be helpful if we wanted to know if telemeter was offline for a period, for instance.

chancez: This doesn't actually tell you if you have gaps, thats why this column is bad. I'd like to remove it from existing queries, but that would need to be done more as a breaking change that we handle with versioning, which i don't want to deal with now. I think it's best we avoid adding it to new queries when possible though.

PR #610, opened 2019-03-05
Wait to run Reports until dependencies have data

chancez:
Update Report handler to only run Reports when their dependencies have data for the reportingPeriod being processed.

Adds a new ImportDataEndTime to Promsum ReportDataSource status, to track what's been queried versus what has been imported.
This information is used by Reports to more accurately determine if a can be processed yet.

When a Prometheus ReportDataSource has data imported, or a sub-Report is processed, the reporting-operator will then requeue any Reports dependent on those resources to be re-processed.


chancez:
WIP because this still needs a good amount of validation and testing. I also need to update docs on status to reflect the changes I believe.

hack/run-e2e-tests.sh
```diff
@@ -12,7 +12,7 @@ set -x
 go test \
     -test.short="${METERING_SHORT_TESTS}" \
     -test.v \
-    -timeout 20m \
+    -timeout 15m \
```
tschuy: any rhyme or reason, or just because it makes tests Go Faster™?

pkg/apis/metering/v1alpha1/report.go
```diff
@@ -116,6 +116,7 @@ type ReportScheduleMonthly struct {
 type ReportStatus struct {
 	Conditions     []ReportCondition `json:"conditions,omitempty"`
 	LastReportTime *meta.Time        `json:"lastReportTime,omitempty"`
+	NextReportTime *meta.Time        `json:"nextReportTime,omitempty"`
```
tschuy: what sets this?

pkg/operator/datasources.go
```diff
@@ -193,55 +195,75 @@ func (op *Reporting) handlePrometheusMetricsDataSource(logger log.FieldLogger, d
...
 
 		// if there is no existing timestamp then this must be the first import
 		// and we should set the earliestImportedMetricTime
 		if earliestImportedMetricTime == nil {
-			earliestImportedMetricTime = &metav1.Time{earliestTS}
-		} else if earliestImportedMetricTime.After(earliestTS) {
+			earliestImportedMetricTime = &metav1.Time{firstMetric.Timestamp}
+		} else if earliestImportedMetricTime.After(firstMetric.Timestamp) {
 			dataSourceLogger.Errorf("detected time new metric import has older data than previously imported, data is likely duplicated.")
 			// TODO(chance): Look at adding an error to the status.
```
tschuy: if we're redoing statuses in this pull request, have you thought about what to do about this?

chancez: I'm only really doing changes to Reports right now. There's a lot more to consider with ReportDataSources since we intend to have different kinds in the future, and errors may need to go into conditions, not a specific section of status. The main changes here are what values are being used to set things in the status so that importing progress is tracked more accurately.

pkg/apis/metering/v1alpha1/report.go
chancez: https://github.com/operator-framework/operator-metering/pull/610/files#diff-d249f334a4e9a4b17d52d7b1f95289faR497

Github suppressed the diff to `reports.go` so you probably need to expand it.

chancez:
/hold

I still need to update docs I just remembered.

pkg/operator/queues.go
```diff
@@ -185,15 +185,18 @@ func (op *Reporting) updateReportGenerationQuery(prev, cur interface{}) {
...
-	if curReportGenerationQuery.ResourceVersion == prevReportGenerationQuery.ResourceVersion {
-		// Periodic resyncs will send update events for all known ReportGenerationQuerys.
-		// Two different versions of the same reportGenerationQuery will always have
-		// different ResourceVersions.
-		logger.Debugf("ReportGenerationQuery %s/%s resourceVersion is unchanged, skipping update", curReportGenerationQuery.Namespace, curReportGenerationQuery.Name)
-		return
-	}
-	if reflect.DeepEqual(prevReportGenerationQuery.Spec, curReportGenerationQuery.Spec) {
-		logger.Debugf("ReportGenerationQuery %s/%s spec is unchanged, skipping update", curReportGenerationQuery.Namespace, curReportGenerationQuery.Name)
+	// Only skip queuing when the if we're not missing a view
```
tschuy: > when the if

pkg/operator/reporting/validate.go
```diff
@@ -300,33 +300,6 @@ func GetDependentGenerationQueriesWithDataSourcesMemoized(queryGetter reportGene
 	return nil
 }
 
-func GetDependentGenerationQueriesMemoized(queryGetter reportGenerationQueryGetter, generationQuery *metering.ReportGenerationQuery, depth, maxDepth int, queriesAccumulator map[string]*metering.ReportGenerationQuery, dynamicQueries bool) error {
```
tschuy: to verify, this is just dead code right? looks so from grepping the repo.

chancez: Yeah

hack/run-e2e-tests.sh
chancez: No I lowered it so that it will timeout before jenkins times out, so our tests exit our scripts can run cleanup steps before jenkins kills the script which collects information. Plus our tests already run in under 15, so I'd like to keep it that way with these changes.

tschuy: Ok, if cleanup depends on this number being <20 we need to add a comment to that effect. Reasonable goal though.

tschuy:
Looks good so far. A few comments, and I want to give it another look on Monday when I have a fresh mind to look at the reports.go with.

hack/run-e2e-tests.sh
chancez: It's really arbitrary based on the environment running it, and only because the way jenkins kills the process doesn't let our scripts receive the signal apparently (might be due to how we run things). Once we move to origin CI, our script will get a proper signal if there's a timeout.

test/e2e/main_test.go
```diff
@@ -59,14 +49,15 @@ func TestMain(m *testing.M) {
...
 	hourlySchedule := &meteringv1alpha1.ReportSchedule{
 		Period: meteringv1alpha1.ReportPeriodHourly,
 	}
 
 	queries := []struct {
-		queryName string
-		skip      bool
+		queryName   string
+		skip        bool
+		nonParallel bool
```
tschuy: `serial`? not sure that's better, but it's at least not going to lead to `!query.nonParallel`

tschuy:
/meow
This looks good to me, but I want @EmilyM1 to give the lgtm on this.

chancez:
I added 1 commit for removing gracePeriod and 1 commit to fix the default namespaces used for e2e/integration that caused the origin `integration-aws-metering` job to fail previously.

PR #614, opened 2019-03-06
update manifests/deploy/openshift/helm-operator/metering.yaml to bypa…

pruan-rht:
…ss error check
having empty `spec:` results in validation error and the metering CRD was not applied.
`error: error validating "/home/pruan/workspace/operator-metering/manifests/deploy/openshift/helm-operator/metering.yaml": error validating data: unknown object type "nil" in Metering.spec; if you choose to ignore these errors, turn validation off with --validate=false
`

### DIFF
--- a/manifests/deploy/openshift/helm-operator/metering.yaml
+++ b/manifests/deploy/openshift/helm-operator/metering.yaml
@@ -2,4 +2,4 @@ apiVersion: metering.openshift.io/v1alpha1
 kind: Metering
 metadata:
   name: operator-metering
-spec:
+spec: []


manifests/deploy/openshift/helm-operator/metering.yaml
```diff
@@ -2,4 +2,4 @@ apiVersion: metering.openshift.io/v1alpha1
 kind: Metering
 metadata:
   name: operator-metering
-spec:
+spec: []
```
chancez: This file is generated, and the value used `[]` is incorrect, as the spec is an object, `{}`.

You'll want to set `cr.spec` to `{}` in https://github.com/operator-framework/operator-metering/blob/master/charts/helm-operator/values.yaml#L3-L4

Eg:

```
cr:
  create: false
  spec: {}
```
Then run `make metering-manifests` and commit your changes.

pruan-rht: @chancez I did that but the generated file is still empty

pruan-rht:
superseded by #615

PR #625, opened 2019-03-12
Update e2e for running against origin CI/4.0 clusters

chancez:
Because 4.0 tests are run against brand new clusters, we can't expect there to
be 2 hours with of data already in monitoring. Instead, use less data, and run
a scheduled report with a more frequent schedule using the cron period
configuration.

This is based on #410 as it changes how tests run dramatically and this leverages the new functionality.


chancez:
/hold

I'm testing this against 4.0 using ci-operator locally currently. Once that passes, I'll remove the hold.

chancez:
Need #628 to merge to fix a bug that this triggers more regularly, but otherwise this is ready.
/assign @tschuy 
/assign @EmilyM1

PR #628, opened 2019-03-13
Dockerfile.src,jenkins: Move checkout of code to happen via jenkins in the container

chancez:
Remove source from Dockerfile.src, so we mount/check it out in builds and tests as needed.

### DIFF
--- a/Dockerfile.src
+++ b/Dockerfile.src
@@ -40,10 +40,13 @@ RUN curl \
 
 RUN helm init --client-only --skip-refresh && helm repo remove stable || true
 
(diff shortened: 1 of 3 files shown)


chancez:
Rebased onto #629 since I hit that bug with this PR.

chancez:
This removes unnecessarily copying the full repo source code into the src image, which gets the source copied in by the CI system in origin CI. As a result, jenkins is also updated to clone the source code in when running this image.

/assign @tschuy 
/assign @EmilyM1

PR #629, opened 2019-03-13
pkg/operator: Fix panic when there are no metrics imported

chancez:
Correctly take length of metrics, not processedTimeRanges to get number
of results imported.

This is because line 628 `firstMetric := results.Metrics[0]` panics when metrics is empty. In PR #610 we changed from using `results.ProcessedTimeRanges` to `results.Metrics` but forgot to change what slice numResultsImported used.


chancez:
Failure is flake I've been seeing all day, overriding.
/override ci/prow/integration-aws-metering

PR #631, opened 2019-03-13
hack/deploy.sh: Fix jq cannot iterate over null errors

chancez:
Fix the ` Cannot iterate over null (null)` that occur in the deploy script.

### DIFF
--- a/hack/deploy.sh
+++ b/hack/deploy.sh
@@ -36,7 +36,7 @@ while true; do
         echo "Namespace is active"
         break
     elif [ "$PHASE" == "Terminating" ]; then
-        echo "Waiting for namespace $METERING_NAMESPACE termination to complete before continuing"
+        echo "Waiting for namespace "$METERING_NAMESPACE" termination to complete before continuing"
     else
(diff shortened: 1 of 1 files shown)


hack/deploy.sh
```diff
@@ -60,20 +60,12 @@ fi
 if [ "$INSTALL_METERING" == "true" ]; then
     echo "Installing metering"
     install_metering "${INSTALL_METHOD}"
-
-    echo
-    echo "Waiting for metering-operator pod to start termination"
-    # we just check until there's a non-ready container then the loop below this will check for readiness
-    until [ "$(kubectl -n $METERING_NAMESPACE get pods -l app=metering-operator -o json | "$FAQ_BIN" -f json -o json -M -c -r '.items | map(.status.containerStatuses[].ready) | all')" == "false" ]; do
```
chancez: This is racy because the operator could become ready before this sees it's unready, so I removed it. It's mostly for continuous upgrade jobs which we'll be removing soon when we swap CI systems anyways. Can revisit later.

PR #632, opened 2019-03-13
Write resourceVersion used to Metering CR status

chancez:
Also updates Dockerfiles to use a newer kubectl by using the `oc` binary as kubectl. Need a newer version to allow patching CRDs. Currently only kubectl 1.5.2 is available in centos/rhel repos.

The idea is this can be used to determine if a Metering operator has reacted to an update in the spec, as currently you cannot tell without looking at what's been deployed.

### DIFF
--- a/images/helm-operator/run-operator.sh
+++ b/images/helm-operator/run-operator.sh
@@ -32,7 +32,6 @@ trap setNeedsExit SIGINT SIGTERM
 CRD="${HELM_RELEASE_CRD_NAME}.${HELM_RELEASE_CRD_API_GROUP}"
 CR_DIRECTORY=/tmp/custom-resources
 UPGRADE_RESULT_DIRECTORY=/tmp/helm-upgrade-result
-RESOURCE_VERSIONS_DIRECTORY=/tmp/resource-versions
 OWNER_PATCH_FILE=/tmp/owner-patch.json
 OWNER_VALUES_FILE=/tmp/owner-values.yaml
 RELEASE_CONFIGMAPS_FILE=/tmp/release-configmaps.json
@@ -137,18 +136,21 @@ helmUpgrade() {
     RELEASE_NAME=$1
     CHART_LOCATION=$2
     NAMESPACE=$3
-    helm upgrade \
+    if helm upgrade \
         --install \
         --namespace "$NAMESPACE" \
-        --wait="$HELM_WAIT" \
         --force \
+        --wait="$HELM_WAIT" \
         --timeout="$HELM_WAIT_TIMEOUT" \
         "$RELEASE_NAME"\
         "$CHART_LOCATION" \
-        "${@:4}"
-    HELM_EXIT_CODE=$?
-    if [ $HELM_EXIT_CODE != 0 ]; then
-        echo "helm upgrade failed, exit code: $HELM_EXIT_CODE"
+        "${@:4}";
+    then
+        echo "Helm upgrade succeeded"
+        return 0
+    else
+        echo "Helm upgrade failed"
+        return 1
     fi
 }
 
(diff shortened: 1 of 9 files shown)


images/helm-operator/run-operator.sh
```diff
@@ -259,38 +261,45 @@ while true; do
...
 
             writeReleaseOwnerValuesFile "$RESOURCE_API_VERSION" "$RESOURCE_KIND" "$RESOURCE_NAME" "$RESOURCE_UID"
             writeReleaseConfigMapOwnerPatchFile "$RESOURCE_API_VERSION" "$RESOURCE_KIND" "$RESOURCE_NAME" "$RESOURCE_UID"
             HELM_ARGS+=("-f" "$OWNER_VALUES_FILE")
 
             mkdir -p "$UPGRADE_RESULT_DIRECTORY"
             echo "Running helm upgrade for release $RELEASE_NAME"
-            if helmUpgrade "$RELEASE_NAME" "$CHART" "$RESOURCE_NAMESPACE" "${HELM_ARGS[@]}" | tee "$UPGRADE_RESULT_DIRECTORY/$RELEASE_NAME.txt"; then
-                echo "Error occurred when processing $FULL_RESOURCE_NAME"
+            if helmUpgrade "$RELEASE_NAME" "$CHART" "$RESOURCE_NAMESPACE" "${HELM_ARGS[@]}"; then
```
chancez: The if conditions here were backwards, so it always errored. I don't know how I managed that.

chancez:
Also resource bump is because modern kubectl/oc are larger and require more memory (even just the oc binary is about 64MB)

images/helm-operator/run-operator.sh
```diff
@@ -259,38 +261,45 @@ while true; do
...
-            HELM_ARGS+=("-f" "$VALUES_FILE")
-        fi
+        echo -E "$RESOURCE_VALUES" > "$NEW_VALUES_FILE"
 
-        mkdir -p "$RESOURCE_VERSIONS_DIRECTORY"
-        RESOURCE_VERSION_FILE="$RESOURCE_VERSIONS_DIRECTORY/$RESOURCE_NAMESPACE-$RESOURCE_NAME.resourceVersion"
-        # If the resource version for this Release CR hasn't changed, we can skip running helm upgrade.
-        if [[ -s "$RESOURCE_VERSION_FILE" && "$(cat "$RESOURCE_VERSION_FILE")" == "$RESOURCE_RESOURCE_VERSION" ]]; then
+        # If the spec for this CR hasn't changed, we can skip running helm upgrade.
+        if [[ -a "$VALUES_FILE" && "$(cat "$VALUES_FILE")" == "$(cat "$NEW_VALUES_FILE")" ]]; then
```
chancez: We now compare the spec of the last one we applied to the current spec, rather than comparing just resourceVersion.

images/helm-operator/run-operator.sh
```diff
@@ -259,38 +261,45 @@ while true; do
...
+            if helmUpgrade "$RELEASE_NAME" "$CHART" "$RESOURCE_NAMESPACE" "${HELM_ARGS[@]}"; then
+                # store the last version that we were able to process so we can
+                # compare new specs against it later
+                mv -f "$NEW_VALUES_FILE" "$VALUES_FILE"
+                echo "Updating $CRD $FULL_RESOURCE_NAME status"
+                kubectl \
+                    -n "$RESOURCE_NAMESPACE" \
+                    patch "$RESOURCE_KIND" "$RESOURCE_NAME" \
+                    --type json \
+                    -p '[{"op": "add", "path": "/status", "value":{}},{"op": "add", "path": "/status/observedVersion", "value":"'"$RESOURCE_RESOURCE_VERSION"'"}]'
```
chancez: patch the status to indicate what version we last observed and deployed.

PR #635, opened 2019-03-15
telemeter-reporting: update chart with new queries

tschuy:
Adds reports using the queries from https://github.com/openshift/cluster-monitoring-operator/pull/281


charts/telemeter-reporting/templates/report-queries/nodes-by-type.yaml
```diff
@@ -0,0 +1,144 @@
+apiVersion: metering.openshift.io/v1alpha1
+kind: ReportGenerationQuery
+metadata:
+  name: "nodes-by-type-raw"
```
chancez: Nit: `nodes-by-instancetype-raw`? I'm thinking there may be other "types" besides instance types (might get confused with role?)

charts/telemeter-reporting/templates/report-queries/nodes-by-type.yaml
```diff
@@ -0,0 +1,144 @@
...
+{{- block "extraMetadata" . }}
+{{- end }}
+spec:
+  reportDataSources:
+  - "nodes-by-type"
+  columns:
+  - name: node_count
+    type: double
+    unit: nodes
+  - name: node_seconds
```
chancez: Do we actually need this column?

charts/telemeter-reporting/templates/report-queries/nodes-by-type.yaml
```diff
@@ -0,0 +1,144 @@
...
+          timestamp,
+          timeprecision,
+          dt
+      FROM {| dataSourceTableName "nodes-by-type" |}
+
+---
+apiVersion: metering.openshift.io/v1alpha1
+kind: ReportGenerationQuery
+metadata:
+  name: "compute-nodes-by-type"
```
chancez: Nit: `-by-instancetype`?

charts/telemeter-reporting/templates/report-queries/nodes-by-type.yaml
tschuy: I don't like that node_count doesn't take into account how long the report period is -- but perhaps this might be better to move into exclusively the non-raw version of the query?

charts/telemeter-reporting/templates/prom-queries/nodes-by-instance-type.yaml
```diff
@@ -1,11 +1,11 @@
 apiVersion: metering.openshift.io/v1alpha1
 kind: ReportPrometheusQuery
 metadata:
-  name: "machine-cpu-cores-by-cluster-id"
+  name: nodes-by-instance-type
```
chancez: node-count-by-instance-type?

PR #643, opened 2019-03-18
Documentation,pkg/operator: Make Report spec.reportingStart optional again.

### LINKED ISSUE #641: allow set "reportingStart" to now
I just updated my installment, and noticed the changes which set reportingStart to be required.

I generate all of the reports via jsonnet - this leaves me with a couple undesirable options:

1 - Set the start time to a random date in the past and wait until the `ReportingPeriodWaiting` eventually catches up, which means for all new clusters a long period of time might pass
2 - Generate a valid RFC3389 date which will continuously mean by reports are being reapplied

Would it be possible to have a the `reportingStart` period be configurable to something like `now` which means it starts immediately? If we can get historical prometheus metrics configured then having the start time configurable is nice, but until then its just a delay in starting the report

chancez:
Automatically determine the next full reportingPeriod after the current
time when spec.reportingStart is unset.

Closes #641

### DIFF
--- a/pkg/operator/reports.go
+++ b/pkg/operator/reports.go
@@ -244,9 +244,6 @@ func (op *Reporting) runReport(logger log.FieldLogger, report *cbTypes.Report) e
 	if report.Spec.GenerationQueryName == "" {
 		return op.setReportStatusInvalidReport(report, "must set spec.generationQuery")
 	}
-	if report.Spec.ReportingStart == nil {
-		return op.setReportStatusInvalidReport(report, "must set spec.reportingStart")
-	}
 	// Validate the reportingStart and reportingEnd make sense and are set when
 	// required.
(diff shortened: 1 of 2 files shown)


pkg/operator/reports.go
```diff
@@ -559,11 +563,11 @@ func getRunOnceReportPeriod(report *cbTypes.Report) (*reportPeriod, error) {
 }
 
 func getNextReportPeriod(schedule reportSchedule, period cbTypes.ReportPeriod, lastScheduled time.Time) *reportPeriod {
-	periodStart := lastScheduled
+	periodStart := lastScheduled.UTC()
 	periodEnd := schedule.Next(periodStart)
 	return &reportPeriod{
-		periodEnd:   periodEnd.Truncate(time.Millisecond).UTC(),
 		periodStart: periodStart.Truncate(time.Millisecond).UTC(),
+		periodEnd:   periodEnd.Truncate(time.Millisecond).UTC(),
```
tschuy: needless diffs are best diffs

chancez: I know I'm terrible and lazy

PR #644, opened 2019-03-18
hack: Update image tag used for origin CI

chancez:
Goes with https://github.com/openshift/release/pull/3140


chancez:
Waiting on #3140 to be merged, then this should be able to pass e2e/integration with the images rebuilt with the proper tags.

PR #655, opened 2019-03-22
Rewrite table and storage location management

chancez:
This updates the reporting-operator APIs to support using HiveTable and PrestoTable resources as the primary way of creating tables in Presto and Hive, and switches many other behaviors to rely on these table resources.


chancez:
/hold

Holding before any approval comes in so we can ensure everyone has tested this before merging.

PR #657, opened 2019-03-25
Make running e2e/integration with local metering/reporting-operator possible

chancez:
Allows running e2e/integration tests locally in a new way: with a local copy of reporting-operator and metering-operator. This means you can immediately run e2e/integration to test changes to charts and reporting-operator without waiting for any image builds.


pkg/operator/pprof.go
```diff
@@ -15,7 +15,7 @@ func newPprofServer() *http.Server {
 	pprofMux.HandleFunc("/debug/pprof/trace", pprof.Trace)
 
 	return &http.Server{
-		Addr:    "127.0.0.1:6060",
+		Addr:    addr,
 		Handler: pprofMux,
 	}
```
EmilyM1: In `start.go` it is at port 6060, which is what the go package uses also. Why change to the string var out of curiosity?

chancez: If you read the commit message for this, it's so that we can configure different ports. It's still defaulted to 6060. We need to be able to configure what ports we listen on so that we can run more than 1 copy of reporting-operator locally, otherwise their ports would conflict. Needed for running e2e and integration at the same time locally.

PR #658, opened 2019-03-25
Refine PVC Used

EmilyM1:
Add period_start and period_end remove phase and associated phase_value, use group by for 1:1 line of data with namespace


charts/openshift-reporting/templates/report-queries/persistentvolumeclaim-usage.yaml
```diff
@@ -170,14 +165,14 @@ spec:
...
     SELECT
+      timestamp '{| default .Report.ReportingStart .Report.Inputs.ReportingStart| prestoTimestamp |}' AS period_start,
+      timestamp '{| default .Report.ReportingEnd .Report.Inputs.ReportingEnd | prestoTimestamp |}' AS period_end,
       namespace,
-      timestamp,
       persistentvolumeclaim,
-      phase,
-      phase_value,
-      persistentvolumeclaim_usage_bytes
+      sum(persistentvolumeclaim_usage_bytes)
```
chancez:
```suggestion
      sum(persistentvolumeclaim_usage_bytes) AS persistentvolumeclaim_usage_bytes
```

chancez: This might not be required, but might be good to add.

chancez:
Looks good to me. Can you do the same for capacity in another PR next?

PR #661, opened 2019-03-28
pkg/operator: Fix metering health check table name

chancez:
Match the health check tablename to it's location

### DIFF
--- a/pkg/operator/reporting/health.go
+++ b/pkg/operator/reporting/health.go
@@ -14,17 +14,19 @@ type PrestoHealthChecker struct {
 	queryer      db.Queryer
 	tableManager TableManager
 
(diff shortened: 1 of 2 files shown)


pkg/operator/reporting/health.go
```diff
@@ -62,25 +63,24 @@ func (checker *PrestoHealthChecker) TestReadFromPresto() bool {
 
 func (checker *PrestoHealthChecker) TestWriteToPresto() bool {
 	logger := checker.logger.WithField("component", "testWriteToPresto")
-	const tableName = "operator_health_check"
```
chancez: The table name was hard coded, and didn't match the location.

PR #669, opened 2019-04-05
reporting-operator: allow custom prometheus CA

tschuy:


charts/reporting-operator/templates/reporting-operator-deployment.yaml
```diff
@@ -243,6 +250,10 @@ spec:
         - name: metrics-tls
           mountPath: /metrics-tls
 {{- end }}
+{{- if .Values.spec.config.prometheusCertificateAuthority.configMap.enabled }}
+        - name: {{ .Values.spec.config.prometheusCertificateAuthority.configMap.name }}
```
chancez: Name should be prometheus-certificate-authority

tschuy:
context on the failing prow tests: I've been told that we're good to go as long as our jenkins tests work, and that we simply need to wait for the Openshift testing infra to get more reliable before we can use that as a merge blocker.

PR #671, opened 2019-04-08
cmd/reporting-operator: Fix port conflict with auth-proxy

chancez:
55a6117a5ae4803d67508a08199c4f9074b05664 changed the default
metrics port accidently, this changes it back to it's old value.
Currently 8081 is used for oauth-proxy when running with auth enabled,
and this causes a port conflict with the current default.


chancez:
CI flake
/override ci/prow/e2e-aws-metering

PR #684, opened 2019-04-16
Use origin images by default and remove extra/unused makefile targets/scripts

chancez:
Basically updates everything to use images built by origin CI, and removes a ton of Makefile targets and variables that are no longer needed as a result of not building images in Jenkins ourselves anymore.


charts/helm-operator/values.yaml
```diff
@@ -15,7 +15,7 @@ crd:
   listKind: HelmReleaseList
 
 image:
-  repository: quay.io/coreosinc/helm-operator
+  repository: "mustbereplaced"
   tag: latest
```
EmilyM1: PR looks good. 
Could `quay.io/openshift/origin-metering-helm-operator` be used as the value here?

chancez: This chart doesn't have any default value that makes sense. We override this always anyways, but this ensures if we didn't, it would be really obvious.

EmilyM1: Good call/check

PR #688, opened 2019-04-18
Use `namespace` label if `exported_namespace` metric label is missing in PVC queries

chancez:
Be compatible with newer versions of kube/prometheus which uses
"namespace" instead of "exported_namespace" in it's
kubelet_volume_stats_* metrics.


chancez:
This passed e2e before and none of our merged changes should effect it. I want QE to validate this sooner than later so overriding.

PR #691, opened 2019-04-22
pkg/operator: Fix small linting errors and remove unneeded code.

timflannagan1:
This would remove an unneeded `else` condition when the accompanied `if` statement returns a value.

Removes any declared type when initializing a variable using the `var` keyword.

### DIFF
--- a/pkg/operator/http.go
+++ b/pkg/operator/http.go
@@ -49,9 +49,9 @@ type server struct {
 	prometheusMetricsRepo prestostore.PrometheusMetricsRepo
 	reportResultsGetter   prestostore.ReportResultsGetter
 
-	reportLister                 listers.ReportLister
-	reportGenerationQuerieLister listers.ReportGenerationQueryLister
-	prestoTableLister            listers.PrestoTableLister
+	reportLister                listers.ReportLister
+	reportGenerationQueryLister listers.ReportGenerationQueryLister
+	prestoTableLister           listers.PrestoTableLister
 }
 
 type requestLogger struct {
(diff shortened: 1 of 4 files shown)


pkg/operator/http.go
```diff
@@ -321,7 +321,7 @@ func writeResultsAsCSV(columns []api.ReportGenerationQueryColumn, results []pres
 func writeResultsResponseAsTabular(logger log.FieldLogger, name string, columns []api.ReportGenerationQueryColumn, results []presto.Row, w http.ResponseWriter, r *http.Request) {
 	w.Header().Set("Content-Type", "text/tab-separated-values")
 	w.Header().Set("Content-Disposition", fmt.Sprintf("attachment;filename=%s.tsv", name))
-	var padding int = 2
+	var padding = 2
```
chancez: This could actually be `padding := 2`

chancez:
Minor nit-pick since you're fixing these things anyways.

pkg/operator/http.go
timflannagan1: Fixed.

timflannagan1:
I saw that in `pkg/operator/http.go` that we have `reportGenerationQuerieLister` field in the server structure and its references five times throughout that file. Is that a typo?

chancez:
Should probably be Queries or Query

pkg/operator/http.go
chancez: I'm not seeing that?

timflannagan1: hmm, that was strange. Updated, again.

timflannagan1:
rebased the commits

PR #693, opened 2019-04-22
Refactor getter logic into test helpers

chancez:


test/testhelpers/helpers.go
```diff
@@ -66,3 +67,66 @@ func NewPrestoTable(name, namespace string, columns []hive.Column) *v1alpha1.Pre
...
+		m[dataSource.Namespace+"/"+dataSource.Name] = dataSource
+	}
+	return &ReportDataSourceStore{m}
+}
+
+func (store *ReportDataSourceStore) GetReportDataSource(namespace, name string) (*v1alpha1.ReportDataSource, error) {
+	dataSource, ok := store.datasources[namespace+"/"+name]
+	if ok {
+		return dataSource, nil
+	} else {
```
timflannagan1: Can drop this `else` clause, and the other ones on line #'s 108 & 129

chancez: Good point.

PR #697, opened 2019-04-24
pkg/operator: Refactor the checks/validations in runReport into individual functions and add their unit tests.

timflannagan1:
Currently, in the `runReport(..)` function in pkg/operator/reports.go, we check the report parameter's running condition. This change would move those checks into a separate function and adds unit tests to verify that the function works as intended.

### DIFF
--- a/pkg/operator/reports.go
+++ b/pkg/operator/reports.go
@@ -1,6 +1,7 @@
 package operator
 
 import (
+	"errors"
 	"fmt"
 	"sort"
 	"strings"
@@ -207,11 +208,8 @@ type reportPeriod struct {
 	periodStart time.Time
 }
 
-// runReport takes a report, and generates reporting data
-// according the report's schedule. If the next scheduled reporting period
-// hasn't elapsed, runReport will requeue the resource for a time when
-// the period has elapsed.
-func (op *Reporting) runReport(logger log.FieldLogger, report *cbTypes.Report) error {
+// isReportFinished checks the running condition of the report parameter and returns true if the report has previously run
+func isReportFinished(logger log.FieldLogger, report *cbTypes.Report) bool {
 	// check if this report was previously finished
 	runningCond := cbutil.GetReportCondition(report.Status, cbTypes.ReportRunning)
 
@@ -222,7 +220,7 @@ func (op *Reporting) runReport(logger log.FieldLogger, report *cbTypes.Report) e
 		// re-processing runOnce reports after they're previously finished
 		if report.Spec.Schedule == nil {
 			logger.Infof("Report %s is a previously finished run-once report, not re-processing", report.Name)
-			return nil
+			return true
 		}
 		// log some messages to indicate we're processing what was a previously finished report
 
@@ -236,43 +234,77 @@ func (op *Reporting) runReport(logger log.FieldLogger, report *cbTypes.Report) e
 		} else {
 			// return without processing because the report is complete
 			logger.Infof("Report %s is already finished: %s", report.Name, runningCond.Message)
-			return nil
+			return true
 		}
 	}
 
+	return false
+}
+
+// validateReport takes a Report structure and checks if it contains valid fields
+func validateReport(
+	report *cbTypes.Report,
+	queryGetter reporting.ReportGenerationQueryGetter,
+	reportDataSourceGetter reporting.ReportDataSourceGetter,
+	reportGetter reporting.ReportGetter, handler *reporting.UninitialiedDependendenciesHandler,
+) (*cbTypes.ReportGenerationQuery, *reporting.ReportGenerationQueryDependencies, error) {
 	// Validate the ReportGenerationQuery is set
 	if report.Spec.GenerationQueryName == "" {
-		return op.setReportStatusInvalidReport(report, "must set spec.generationQuery")
-	}
-	// Validate the reportingStart and reportingEnd make sense and are set when
-	// required.
-	if report.Spec.ReportingStart != nil && report.Spec.ReportingEnd != nil && (report.Spec.ReportingStart.Time.After(report.Spec.ReportingEnd.Time) || report.Spec.ReportingStart.Time.Equal(report.Spec.ReportingEnd.Time)) {
-		return op.setReportStatusInvalidReport(report, fmt.Sprintf("spec.reportingEnd (%s) must be after spec.reportingStart (%s)", report.Spec.ReportingEnd.Time, report.Spec.ReportingStart.Time))
-	}
-	if report.Spec.ReportingEnd == nil && report.Spec.RunImmediately {
-		return op.setReportStatusInvalidReport(report, "spec.reportingEnd must be set if report.spec.runImmediately is true")
+		return nil, nil, errors.New("must set spec.generationQuery")
 	}
 
-	// Validate the ReportGenerationQuery used exists
-	genQuery, err := op.getReportGenerationQueryForReport(report)
+	// Validate the reportingStart and reportingEnd make sense and are set when
+	// required
+	if report.Spec.ReportingStart != nil && report.Spec.ReportingEnd != nil && (report.Spec.ReportingStart.Time.After(report.Spec.ReportingEnd.Time) || report.Spec.ReportingStart.Time.Equal(report.Spec.ReportingEnd.Time)) {
+		return nil, nil, fmt.Errorf("spec.reportingEnd (%s) must be after spec.reportingStart (%s)", report.Spec.ReportingEnd.Time, report.Spec.ReportingStart.Time)
+	}
+	if report.Spec.ReportingEnd == nil && report.Spec.RunImmediately {
+		return nil, nil, errors.New("spec.reportingEnd must be set if report.spec.runImmediately is true")
+	}
+
+	// Validate the ReportGenerationQuery that the Report used exists
+	genQuery, err := GetReportGenerationQueryForReport(report, queryGetter)
 	if err != nil {
-		logger.WithError(err).Errorf("failed to get report generation query")
 		if apierrors.IsNotFound(err) {
-			return op.setReportStatusInvalidReport(report, fmt.Sprintf("ReportGenerationQuery %s does not exist", report.Spec.GenerationQueryName))
+			return nil, nil, fmt.Errorf("ReportGenerationQuery (%s) does not exist", report.Spec.GenerationQueryName)
 		}
-		return err
+		return nil, nil, fmt.Errorf("failed to get report generation query")
 	}
 
 	// Validate the dependencies of this Report's query exist
 	queryDependencies, err := reporting.GetAndValidateGenerationQueryDependencies(
-		reporting.NewReportGenerationQueryListerGetter(op.reportGenerationQueryLister),
-		reporting.NewReportDataSourceListerGetter(op.reportDataSourceLister),
-		reporting.NewReportListerGetter(op.reportLister),
+		queryGetter,
+		reportDataSourceGetter,
+		reportGetter,
 		genQuery,
-		op.uninitialiedDependendenciesHandler(),
+		handler,
 	)
 	if err != nil {
-		return op.setReportStatusInvalidReport(report, fmt.Sprintf("failed to validate ReportGenerationQuery dependencies %s: %v", genQuery.Name, err))
+		return nil, nil, fmt.Errorf("failed to validate ReportGenerationQuery dependencies %s: %v", genQuery.Name, err)
+	}
+
+	return genQuery, queryDependencies, nil
+}
+
+// runReport takes a report, and generates reporting data
+// according the report's schedule. If the next scheduled reporting period
+// hasn't elapsed, runReport will requeue the resource for a time when
+// the period has elapsed.
+func (op *Reporting) runReport(logger log.FieldLogger, report *cbTypes.Report) error {
+	// check if the report was previously finished; store result in bool
+	if reportFinished := isReportFinished(logger, report); reportFinished {
+		return nil
+	}
+
+	runningCond := cbutil.GetReportCondition(report.Status, cbTypes.ReportRunning)
+	queryGetter := reporting.NewReportGenerationQueryListerGetter(op.reportGenerationQueryLister)
+	reportDataSourceGetter := reporting.NewReportDataSourceListerGetter(op.reportDataSourceLister)
+	reportGetter := reporting.NewReportListerGetter(op.reportLister)
+
+	// validate that Report contains valid Spec fields
+	genQuery, queryDependencies, err := validateReport(report, queryGetter, reportDataSourceGetter, reportGetter, op.uninitialiedDependendenciesHandler())
+	if err != nil {
+		return op.setReportStatusInvalidReport(report, err.Error())
 	}
 
 	now := op.clock.Now().UTC()
@@ -672,17 +704,20 @@ func (op *Reporting) setReportStatusInvalidReport(report *cbTypes.Report, msg st
 	return err
 }
 
-func (op *Reporting) getReportGenerationQueryForReport(report *cbTypes.Report) (*cbTypes.ReportGenerationQuery, error) {
-	return op.reportGenerationQueryLister.ReportGenerationQueries(report.Namespace).Get(report.Spec.GenerationQueryName)
+// GetReportGenerationQueryForReport returns the ReportGenerationQuery that was used in the Report parameter
+func GetReportGenerationQueryForReport(report *cbTypes.Report, queryGetter reporting.ReportGenerationQueryGetter) (*cbTypes.ReportGenerationQuery, error) {
+	return queryGetter.GetReportGenerationQuery(report.Namespace, report.Spec.GenerationQueryName)
 }
 
 func (op *Reporting) getReportDependencies(report *cbTypes.Report) (*reporting.ReportGenerationQueryDependencies, error) {
-	genQuery, err := op.getReportGenerationQueryForReport(report)
+	queryGetter := reporting.NewReportGenerationQueryListerGetter(op.reportGenerationQueryLister)
+
+	genQuery, err := GetReportGenerationQueryForReport(report, queryGetter)
 	if err != nil {
 		return nil, err
 	}
 	return reporting.GetGenerationQueryDependencies(
-		reporting.NewReportGenerationQueryListerGetter(op.reportGenerationQueryLister),
+		queryGetter,
 		reporting.NewReportDataSourceListerGetter(op.reportDataSourceLister),
 		reporting.NewReportListerGetter(op.reportLister),
 		genQuery,
--- a/pkg/operator/reports_test.go
+++ b/pkg/operator/reports_test.go
@@ -1,23 +1,33 @@
 package operator
 
 import (
+	"fmt"
 	"testing"
 	"time"
 
+	"github.com/operator-framework/operator-metering/pkg/operator/reporting"
+	"github.com/operator-framework/operator-metering/pkg/operator/reportingutil"
+
 	"github.com/operator-framework/operator-metering/pkg/apis/metering/v1alpha1"
+	metering "github.com/operator-framework/operator-metering/pkg/apis/metering/v1alpha1"
+	"github.com/operator-framework/operator-metering/pkg/apis/metering/v1alpha1/util"
+	"github.com/operator-framework/operator-metering/test/testhelpers"
+	"github.com/sirupsen/logrus"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 func TestGetNextReportPeriod(t *testing.T) {
 	baseTime := time.Date(2018, time.July, 1, 0, 0, 0, 0, time.UTC)
 	tests := map[string]struct {
-		period              v1alpha1.ReportPeriod
+		period              metering.ReportPeriod
 		expectError         bool
 		expectReportPeriods []reportPeriod
 	}{
 		"hourly": {
-			period: v1alpha1.ReportPeriodHourly,
+			period: metering.ReportPeriodHourly,
 			expectReportPeriods: []reportPeriod{
 				{
 					periodStart: baseTime,
@@ -30,7 +40,7 @@ func TestGetNextReportPeriod(t *testing.T) {
 			},
 		},
 		"daily": {
-			period: v1alpha1.ReportPeriodDaily,
+			period: metering.ReportPeriodDaily,
 			expectReportPeriods: []reportPeriod{
 				{
 					periodStart: baseTime,
@@ -43,7 +53,7 @@ func TestGetNextReportPeriod(t *testing.T) {
 			},
 		},
 		"weekly": {
-			period: v1alpha1.ReportPeriodWeekly,
+			period: metering.ReportPeriodWeekly,
 			expectReportPeriods: []reportPeriod{
 				{
 					periodStart: baseTime,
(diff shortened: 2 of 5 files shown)


pkg/operator/reports_test.go
```diff
@@ -92,3 +96,86 @@ func TestGetNextReportPeriod(t *testing.T) {
...
+					Message:            "mock-msg",
+				},
+			},
+			LastReportTime: &v1.Time{Time: reportStart.AddDate(0, 2, 0)},
+		}, schedule), true},
+	}
+
+	for _, testCase := range testTable {
+		var mockLogger = logrus.New()
+		test := testCase.name
```
chancez: `testCase := testCase`

pkg/operator/reports_test.go
```diff
@@ -92,3 +96,86 @@ func TestGetNextReportPeriod(t *testing.T) {
...
+
+		t.Run(test, func(t *testing.T) {
```
chancez: just use `testCase.name`

pkg/operator/reports_test.go
```diff
@@ -92,3 +96,86 @@ func TestGetNextReportPeriod(t *testing.T) {
 		})
 	}
 }
+
+func TestReportRunningCondition(t *testing.T) {
```
chancez: This is testing `isReportRunning` so the test name should be `TestIsReportRunning`

pkg/operator/reports_test.go
```diff
@@ -92,3 +96,86 @@ func TestGetNextReportPeriod(t *testing.T) {
...
+	reportStart := &time.Time{}
+	reportEndTmp := reportStart.AddDate(0, 1, 0)
+	reportEnd := &reportEndTmp
+
+	testTable := []struct {
+		name       string
+		report     *metering.Report
+		expectFlag bool
+	}{
+		{"new report returns false", testhelpers.NewReport(testReportName, namespace, testQueryName, reportStart, reportEnd, v1alpha1.ReportStatus{}, nil), false},
```
chancez: Always use struct fields in composite literals.

pkg/operator/reports_test.go
```diff
@@ -5,19 +5,23 @@ import (
 	"time"
 
 	"github.com/operator-framework/operator-metering/pkg/apis/metering/v1alpha1"
+	metering "github.com/operator-framework/operator-metering/pkg/apis/metering/v1alpha1"
+	"github.com/operator-framework/operator-metering/test/testhelpers"
+	"github.com/sirupsen/logrus"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
chancez: Generally we use `metav1` as the package alias since `v1` usually refers to the `k8s.io/api/core/v1` package.

pkg/operator/reports_test.go
```diff
@@ -92,3 +96,86 @@ func TestGetNextReportPeriod(t *testing.T) {
...
+	}{
+		{"new report returns false", testhelpers.NewReport(testReportName, namespace, testQueryName, reportStart, reportEnd, v1alpha1.ReportStatus{}, nil), false},
+		{"run-once report returns true", testhelpers.NewReport(testReportName, namespace, testQueryName, reportStart, reportEnd, v1alpha1.ReportStatus{
+			Conditions: []metering.ReportCondition{
+				{
+					Type:               metering.ReportRunning,
+					Status:             "Active",
+					LastUpdateTime:     v1.Now(),
+					LastTransitionTime: v1.Now(),
+					Reason:             "Finished",
```
chancez: use the constants in the `https://github.com/operator-framework/operator-metering/blob/master/pkg/apis/metering/v1alpha1/util/report_util.go` package for the value of the `Reason` field in these .

pkg/operator/reports_test.go
```diff
@@ -92,3 +96,86 @@ func TestGetNextReportPeriod(t *testing.T) {
...
+
+	testTable := []struct {
+		name       string
+		report     *metering.Report
+		expectFlag bool
+	}{
+		{"new report returns false", testhelpers.NewReport(testReportName, namespace, testQueryName, reportStart, reportEnd, v1alpha1.ReportStatus{}, nil), false},
+		{"run-once report returns true", testhelpers.NewReport(testReportName, namespace, testQueryName, reportStart, reportEnd, v1alpha1.ReportStatus{
+			Conditions: []metering.ReportCondition{
+				{
```
chancez: You could use `NewReportCondition` from https://github.com/operator-framework/operator-metering/blob/master/pkg/apis/metering/v1alpha1/util/report_util.go  instead of using a struct literal.

pkg/operator/reports_test.go
```diff
@@ -92,3 +96,86 @@ func TestGetNextReportPeriod(t *testing.T) {
...
+					Type:               metering.ReportRunning,
+					Status:             "Active",
```
chancez: Use `k8s.io/api/core/v1`'s `ConditionStatus` (https://github.com/operator-framework/operator-metering/blob/master/vendor/k8s.io/api/core/v1/types.go#L2229:6) for the value here.

pkg/operator/reports.go
```diff
@@ -235,46 +233,85 @@ func (op *Reporting) runReport(logger log.FieldLogger, report *cbTypes.Report) e
...
-		return err
+		return fmt.Errorf("failed to get report generation query")
 	}
 
 	// Validate the dependencies of this Report's query exist
-	queryDependencies, err := reporting.GetAndValidateGenerationQueryDependencies(
-		reporting.NewReportGenerationQueryListerGetter(op.reportGenerationQueryLister),
-		reporting.NewReportDataSourceListerGetter(op.reportDataSourceLister),
-		reporting.NewReportListerGetter(op.reportLister),
+	_, err = reporting.GetAndValidateGenerationQueryDependencies(
```
chancez: This function can return the dependencies

pkg/operator/reports.go
```diff
@@ -235,46 +233,85 @@ func (op *Reporting) runReport(logger log.FieldLogger, report *cbTypes.Report) e
...
+		queryGetter,
+		reportDataSourceGetter,
+		reportGetter,
 		genQuery,
-		op.uninitialiedDependendenciesHandler(),
+		&reporting.UninitialiedDependendenciesHandler{},
```
chancez: You can pass in the handler as an argument of type `reporting.UninitialiedDependendenciesHandler`

pkg/operator/reports.go
```diff
@@ -235,46 +233,85 @@ func (op *Reporting) runReport(logger log.FieldLogger, report *cbTypes.Report) e
...
+		return nil
+	}
+
+	runningCond := cbutil.GetReportCondition(report.Status, cbTypes.ReportRunning)
+	queryGetter := reporting.NewReportGenerationQueryListerGetter(op.reportGenerationQueryLister)
+	reportDataSourceGetter := reporting.NewReportDataSourceListerGetter(op.reportDataSourceLister)
+	reportGetter := reporting.NewReportListerGetter(op.reportLister)
+
+	// validate that Report contains valid Spec fields
+	err := validateReport(logger, report, queryGetter, reportDataSourceGetter, reportGetter)
```
chancez: pass `op.uninitialiedDependendenciesHandler()` as an argument

pkg/operator/reports.go
```diff
@@ -235,46 +233,85 @@ func (op *Reporting) runReport(logger log.FieldLogger, report *cbTypes.Report) e
...
+	if err != nil {
+		return op.setReportStatusInvalidReport(report, err.Error())
+	}
+
+	// we validated these variables in the validateReport, so we know that the errors are nil
```
chancez: Remove this code, return dependencies from `validateReport`

pkg/operator/reports_test.go
```diff
@@ -92,3 +99,224 @@ func TestGetNextReportPeriod(t *testing.T) {
...
+	reportStart := &time.Time{}
+	reportEndTmp := reportStart.AddDate(0, 1, 0)
+	reportEnd := &reportEndTmp
+
+	testTable := []struct {
+		name       string
+		report     *metering.Report
+		expectFlag bool
+	}{
+		{name: "new report returns false", report: testhelpers.NewReport(testReportName, testNamespace, testQueryName, reportStart, reportEnd, v1alpha1.ReportStatus{}, nil, false), expectFlag: false},
```
chancez: Can you put each field onto it's own line so these lines aren't so long? And put `name` onto a separate line from the open `{`

pkg/operator/reports_test.go
```diff
@@ -92,3 +99,224 @@ func TestGetNextReportPeriod(t *testing.T) {
...
+	reportEndTmp := reportStart.AddDate(0, 1, 0)
+	reportEnd := &reportEndTmp
+
+	testTable := []struct {
+		name         string
+		report       *metering.Report
+		expectFlag   bool
+		expectErrMsg string
+	}{
+		{name: "empty spec.generationQuery returns err",
```
chancez: Put `name:` onto the next line, separate from `{`

pkg/operator/reports_test.go
```diff
@@ -92,3 +99,224 @@ func TestGetNextReportPeriod(t *testing.T) {
...
+			expectFlag:   true,
+			expectErrMsg: fmt.Sprintf("failed to validate ReportGenerationQuery dependencies %s: %v", testGenerationQueryName, "stub"),
+		},
+	}
+
+	// hardcore the report spec's generation query name for tests that need that
+	testTable[3].report.Spec.GenerationQueryName = testGenerationQueryName
+	testTable[4].report.Spec.GenerationQueryName = testGenerationQueryName
+
+	ds1 := testhelpers.NewReportDataSource("datasource1", testNamespace)
```
chancez: We already test the validate dependencies code on it's own thoroughly, so you don't need to re-test that here, so you can keep the resources used for testing fairly minimal .

pkg/operator/reports_test.go
```diff
@@ -92,3 +101,154 @@ func TestGetNextReportPeriod(t *testing.T) {
...
+	}
+
+	reportStart := &time.Time{}
+	reportEndTmp := reportStart.AddDate(0, 1, 0)
+	reportEnd := &reportEndTmp
+
+	testTable := []struct {
+		name       string
+		report     *metering.Report
+		expectFlag bool
```
chancez: This would be clearer as `expectRunning`

pkg/operator/reports_test.go
```diff
@@ -92,3 +101,154 @@ func TestGetNextReportPeriod(t *testing.T) {
...
+	}{
+		{
+			name:       "new report returns false",
+			report:     testhelpers.NewReport(testReportName, testNamespace, testQueryName, reportStart, reportEnd, v1alpha1.ReportStatus{}, nil, false),
+			expectFlag: false},
+		{
+			name: "run-once report returns true",
```
chancez: This is more about it being in a finished status than it being `run-once`

pkg/operator/reports_test.go
```diff
@@ -92,3 +101,154 @@ func TestGetNextReportPeriod(t *testing.T) {
...
+		{
+			name: "reportingEnd < lastReportTime returns true",
+			report: testhelpers.NewReport(testReportName, testNamespace, testQueryName, reportStart, reportEnd, v1alpha1.ReportStatus{
+				Conditions: []metering.ReportCondition{
+					*util.NewReportCondition(metering.ReportRunning, v1.ConditionFalse, util.ReportFinishedReason, "mock-msg"),
+				},
+				LastReportTime: &metav1.Time{Time: reportStart.AddDate(0, 2, 0)},
+			}, schedule, false),
+			expectFlag: true},
+	}
```
chancez: You could add some more test cases where the reason isn't finished just to cover the bases.

pkg/operator/reports_test.go
timflannagan1: I also changed the bool `expectFlag` -> `expectErr` in the second test function

pkg/operator/reports_test.go
timflannagan1: Updated to "finished status on run-once report returns true"

pkg/operator/reports_test.go
```diff
@@ -92,3 +101,177 @@ func TestGetNextReportPeriod(t *testing.T) {
...
+		{
+			name: "spec.Reason set to 'InvalidReport' returns false",
+			report: testhelpers.NewReport(testReportName, testNamespace, testQueryName, reportStart, reportEnd, v1alpha1.ReportStatus{
+				Conditions: []metering.ReportCondition{
+					*util.NewReportCondition(metering.ReportRunning, v1.ConditionFalse, util.InvalidReportReason, "mock-msg"),
+				},
+			}, schedule, false),
+			expectRunning: false,
+		},
+	}
```
chancez: We need at least 1 case where running is true (`v1.ConditionTrue`), maybe more.

timflannagan1: Added - wasn't able to think of any other paths that weren't tested.

chancez: I guess our test coverage files would probably help with that. You can verify by running `make test` and it will output `coverage.out` which can be used with the `go tool cover` command: https://blog.golang.org/cover, want to do a check for us?

pkg/operator/reports_test.go
```diff
@@ -92,3 +101,186 @@ func TestGetNextReportPeriod(t *testing.T) {
...
+		{
+			name:         "spec.GenerationQueryName does not exist returns err",
+			report:       testhelpers.NewReport(testReportName, testNamespace, testQueryName, reportStart, reportEnd, v1alpha1.ReportStatus{}, nil, false),
+			expectErr:    true,
+			expectErrMsg: fmt.Sprintf("ReportGenerationQuery %s does not exist", testGenerationQueryName),
+		},
+	}
+
+	// hardcore the report spec's generation query name for tests that need that
+	testTable[3].report.Spec.GenerationQueryName = testGenerationQueryName
```
chancez: perhaps instead you could add a queryName field to each testCase.

pkg/operator/reports_test.go
```diff
@@ -92,3 +101,186 @@ func TestGetNextReportPeriod(t *testing.T) {
...
+	ds1 := testhelpers.NewReportDataSource("datasource1", testNamespace)
```
chancez: All of the queries, datasources and the `.*Stores` setup i think should be above where we define the test cases so that it's done prior to referencing things in the cases.

timflannagan1: That makes sense. Updated.

pkg/operator/reports_test.go
timflannagan1: Added.

pkg/operator/reports_test.go
```diff
@@ -92,3 +101,190 @@ func TestGetNextReportPeriod(t *testing.T) {
...
+			expectRunning: false,
+		},
+		{
+			name: "spec.Status set to v1.ConditionTrue returns false",
+			report: testhelpers.NewReport(testReportName, testNamespace, testQueryName, reportStart, reportEnd, v1alpha1.ReportStatus{
+				Conditions: []metering.ReportCondition{
+					*util.NewReportCondition(metering.ReportRunning, v1.ConditionTrue, util.ReportFinishedReason, "mock-msg"),
+				},
+			}, schedule, false),
+			expectRunning: false,
```
chancez: Hmm I think the `reason` here needs to be something else, or we need another case, because the intent was that we have a test case where `expectRunning` is true, when there's a running condition.

Currently this test case is basically stating: "we don't expect it to be running, when Running is true with reason finished" which doesn't make much sense. In our actual code, this shouldn't ever happen (unless we have a bug of course).

Reason is never finished if the Running condition is true, so this is testing a case we don't currently expect to happen. So this test case might be fine, but it may also be good to figure out what should happen here (perhaps we should be panicing if this happened in our code, and our test should verify we panic).

Anyways, we need another test case where Running is true, and expectRunning is true. Use `ScheduledReason` or `RunImmediatelyReason` to handle the happy path where a report is not finished, and is just running as regular.

pkg/operator/reports_test.go
```diff
@@ -92,3 +101,190 @@ func TestGetNextReportPeriod(t *testing.T) {
...
+			name: "spec.Reason set to 'InvalidReport' returns false",
+			report: testhelpers.NewReport(testReportName, testNamespace, testQueryName, reportStart, reportEnd, v1alpha1.ReportStatus{
+				Conditions: []metering.ReportCondition{
+					*util.NewReportCondition(metering.ReportRunning, v1.ConditionFalse, util.InvalidReportReason, "mock-msg"),
+				},
+			}, schedule, false),
+			expectRunning: false,
+		},
+		{
+			name: "spec.Status set to v1.ConditionTrue returns false",
```
chancez: This name isn't very descriptive. `When status running is true and reason is finished, returns false" might be closer, but read the next comment since you may need to update this test case and/or implementation code.

pkg/operator/reports_test.go
```diff
@@ -92,3 +101,190 @@ func TestGetNextReportPeriod(t *testing.T) {
...
+	reportEnd := &reportEndTmp
+
+	testTable := []struct {
+		name         string
+		report       *metering.Report
+		queryName    string
+		expectErr    bool
+		expectErrMsg string
+	}{
+		{
```
chancez: Need a test case for the happy path when ValidateReport doesn't return an error.

pkg/operator/reports_test.go
chancez: Turns out, the problem aws expectRunning is actually `expectFinished` so this problem is a bit different, mostly a matter of what condition reason is used.

pkg/operator/reports_test.go
```diff
@@ -92,3 +101,199 @@ func TestGetNextReportPeriod(t *testing.T) {
...
+			name: "when status running is false and reason is InvalidReport return false",
+			report: testhelpers.NewReport(testReportName, testNamespace, testQueryName, reportStart, reportEnd, v1alpha1.ReportStatus{
+				Conditions: []metering.ReportCondition{
+					*util.NewReportCondition(metering.ReportRunning, v1.ConditionFalse, util.InvalidReportReason, "mock-msg"),
+				},
+			}, schedule, false),
+			expectFinished: false,
+		},
+		{
+			name: "when status running is true and reason is Scheduled return false",
```
chancez: Nit: could you put this one immediately below the case with name `name: "when status running is false and reason is Scheduled return false"`. Having them closer visually should hopefully help make it easier to distinguish them from each other.

pkg/operator/reports_test.go
chancez: Not seeing this fixed @timflannagan1

pkg/operator/reports_test.go
```diff
@@ -92,3 +101,217 @@ func TestGetNextReportPeriod(t *testing.T) {
...
+		testReportName          = "test-report"
+		testQueryName           = "test-query"
+		testGenerationQueryName = "test-GQN"
+	)
+
+	ds1 := testhelpers.NewReportDataSource("datasource1", testNamespace)
+
+	testQuery := &metering.ReportGenerationQuery{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "test-query",
```
chancez: this could be testQueryName

pkg/operator/reports_test.go
```diff
@@ -92,3 +101,217 @@ func TestGetNextReportPeriod(t *testing.T) {
...
+			Namespace: testNamespace,
+		},
+		Spec: metering.ReportGenerationQuerySpec{
+			DataSources: []string{
+				"datasource1",
```
chancez: Nit: Could be ds1.Name

pkg/operator/reports_test.go
```diff
@@ -92,3 +101,217 @@ func TestGetNextReportPeriod(t *testing.T) {
...
+		})
+	}
+}
+
+func TestValidateReport(t *testing.T) {
+	const (
+		testNamespace           = "default"
+		testReportName          = "test-report"
+		testQueryName           = "test-query"
+		testGenerationQueryName = "test-GQN"
```
chancez: To make this more obvious, perhaps change this to:

```suggestion
		testNonexistentQueryName = "does-not-exist"
```

pkg/operator/reports_test.go
```diff
@@ -92,3 +101,217 @@ func TestGetNextReportPeriod(t *testing.T) {
...
+			name:         "spec.GenerationQueryName does not exist returns err",
+			report:       testhelpers.NewReport(testReportName, testNamespace, testQueryName, reportStart, reportEnd, v1alpha1.ReportStatus{}, nil, false),
+			queryName:    testGenerationQueryName,
+			expectErr:    true,
+			expectErrMsg: fmt.Sprintf("ReportGenerationQuery %s does not exist", testGenerationQueryName),
+		},
+	}
+
+	for _, testCase := range testTable {
+		testCase.report.Spec.GenerationQueryName = testCase.queryName
```
chancez: I dislike this. We should instead just set the queryName we intend to use directly in each test case.

pkg/operator/reports_test.go
```diff
@@ -92,3 +101,217 @@ func TestGetNextReportPeriod(t *testing.T) {
...
+	reportGetter := testhelpers.NewReportStore(nil)
+
+	reportStart := &time.Time{}
+	reportEndTmp := reportStart.AddDate(0, 1, 0)
+	reportEnd := &reportEndTmp
+
+	testTable := []struct {
+		name         string
+		report       *metering.Report
+		queryName    string
```
chancez: Based on my other review comments, this field should be removed. Instead, set the correct query name when you call `testhelper.NewReport`

pkg/operator/reports_test.go
chancez: Also a test case where the error is a dependency error, eg, when a generationQuery depends on a non-existent datasource.

pkg/operator/reports_test.go
timflannagan1: Oh yikes, I misinterpreted the testQueryName field in `NewReport`

pkg/operator/reports_test.go
```diff
@@ -92,3 +102,235 @@ func TestGetNextReportPeriod(t *testing.T) {
...
+	for _, testCase := range testTable {
+		testCase := testCase
+
+		t.Run(testCase.name, func(t *testing.T) {
+			_, _, err := validateReport(testCase.report, queryGetter, dataSourceGetter, reportGetter, &reporting.UninitialiedDependendenciesHandler{})
+
+			if testCase.expectErr {
+				assert.EqualErrorf(t, err, testCase.expectErrMsg, "expected that validateReport would return the correct error message")
+			} else {
+				assert.NoErrorf(t, err, "expected the report '%s' would return no error, but got '%v'", testCase.name, err)
```
chancez: You shouldn't need to print the test case name since it's in the output already. Also, you refer to the test case name as the report name. Probably not necessary either since each test case uses the same `testReportName` as reportName.

pkg/operator/reports_test.go
```diff
@@ -92,3 +102,235 @@ func TestGetNextReportPeriod(t *testing.T) {
...
+		},
+	}
+
+	for _, testCase := range testTable {
+		var mockLogger = logrus.New()
+		testCase := testCase
+
+		t.Run(testCase.name, func(t *testing.T) {
+			runningCond := isReportFinished(mockLogger, testCase.report)
+			assert.Equalf(t, runningCond, testCase.expectFinished, "expected the report '%s' would return '%t', but got '%t'", testCase.name, testCase.expectFinished, runningCond)
```
chancez: This message is repetitive (and the report name isn't the test case name).

timflannagan1: Updated.

pkg/operator/reports_test.go
timflannagan1: Updated.

PR #702, opened 2019-04-29
pkg/operator: Refactor getting a report's reportPeriod into its own function and add unit tests.

timflannagan1:
Currently based off of #697. Same idea as that PR - refactor portions of `runReport` into their own function with the intent of being able to unit tests those functions.

There's a `panic` call when the built-up `reportPeriod` struct contains an invalid start and end period, so I used `assert.Panics()` to test this, but there may be better way to achieve this.

### DIFF
--- a/pkg/operator/reports.go
+++ b/pkg/operator/reports.go
@@ -286,6 +286,57 @@ func validateReport(
 	return genQuery, queryDependencies, nil
 }
 
+// getReportPeriod determines a Report's reporting period based off the report parameter's fields.
+// Returns a pointer to a reportPeriod structure if no error was encountered, else panic or return an error.
+func getReportPeriod(now time.Time, logger log.FieldLogger, report *cbTypes.Report) (*reportPeriod, error) {
+	var reportPeriod *reportPeriod
+
+	// check if the report's schedule spec is set
+	if report.Spec.Schedule != nil {
+		reportSchedule, err := getSchedule(report.Spec.Schedule)
+		if err != nil {
+			return nil, err
+		}
+
+		if report.Status.LastReportTime != nil {
+			reportPeriod = getNextReportPeriod(reportSchedule, report.Spec.Schedule.Period, report.Status.LastReportTime.Time)
+		} else {
+			if report.Spec.ReportingStart != nil {
+				logger.Infof("no last report time for report, using spec.reportingStart %s as starting point", report.Spec.ReportingStart.Time)
+				reportPeriod = getNextReportPeriod(reportSchedule, report.Spec.Schedule.Period, report.Spec.ReportingStart.Time)
+			} else if report.Status.NextReportTime != nil {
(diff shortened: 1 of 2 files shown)


timflannagan1:
/hold 
wait for #697 to be merged first.

pkg/operator/reports.go
```diff
@@ -286,34 +286,17 @@ func validateReport(
...
-		return op.setReportStatusInvalidReport(report, err.Error())
-	}
-
-	now := op.clock.Now().UTC()
-
+// getReportPeriod determines a Report's reporting period based off the report parameter's fields.
+// Returns a pointer to a reportPeriod structure if no error was encountered, else panic or return an error.
+func getReportPeriod(now time.Time, logger log.FieldLogger, report *cbTypes.Report) (*reportPeriod, error) {
 	var reportPeriod *reportPeriod
+	var err error
```
chancez: You only actually need this at line 320/321.

pkg/operator/reports_test.go
```diff
@@ -335,3 +335,105 @@ func TestValidateReport(t *testing.T) {
...
+		name        string
+		report      *metering.Report
+		expectErr   bool
+		expectPanic bool
+	}{
+		{
+			name:        "invalid report with an unset spec.Schedule field returns an error",
+			report:      testhelpers.NewReport(testReportName, testNamespace, testQueryName, nil, nil, v1alpha1.ReportStatus{}, nil, false),
+			expectErr:   true,
+			expectPanic: false,
```
chancez: This can be omitted on anything where it's false.

chancez:
Minor nit, otherwise this LGTM.

timflannagan1:
Updated.

pkg/operator/reports_test.go
```diff
@@ -335,3 +335,97 @@ func TestValidateReport(t *testing.T) {
...
+	for _, testCase := range testTable {
+		var mockLogger = logrus.New()
+		testCase := testCase
+
+		t.Run(testCase.name, func(t *testing.T) {
+			if testCase.expectPanic {
+				assert.Panics(t, func() { getReportPeriod(time.Now(), mockLogger, testCase.report) }, "expected the test case would panic, but it did not")
+			} else {
+				_, err := getReportPeriod(time.Now(), mockLogger, testCase.report)
+				if testCase.expectErr {
```
chancez: Need an else branch to cover when expectErr is false (probably just `assert.IsNil`

pkg/operator/reports_test.go
```diff
@@ -335,3 +335,97 @@ func TestValidateReport(t *testing.T) {
...
+					assert.Errorf(t, err, "expected that the report would return a non-nil error, but got '%v'", testCase.name, err)
```
chancez: "expected that getting the report period would return a non-nil error"`

chancez: No need to include `err` in it either, the errorf already includes it

PR #710, opened 2019-05-09
pkg/operator: Add unit-tests for ReportQuery templating.

timflannagan1:
Currently, I need to add more test cases, clean up the code, and potentially move the test helper functions to a separate file. 

These unit tests would handle the validation of a ReportQuery's templating.

Also added a check for the return of `reportingutil.FullyQualifiedTableName` as I was able to get the function to return `..initialized_datasource`, which would result in invalid SQL (i.e. `SELECT * FROM ..initialized_datasouce`. This could either be moved to a function or have the function do the check to eliminate the code repetition.

### DIFF
--- a/pkg/operator/http.go
+++ b/pkg/operator/http.go
@@ -227,7 +227,12 @@ func (srv *server) getReport(logger log.FieldLogger, name, namespace, format str
 		logger.Debugf("mismatched columns, PrestoTable columns: %v, ReportQuery columns: %v", prestoColumns, queryPrestoColumns)
 	}
 
-	tableName := reportingutil.FullyQualifiedTableName(prestoTable)
+	tableName, err := reportingutil.FullyQualifiedTableName(prestoTable)
+	if err != nil {
+		logger.WithError(err).Errorf("prestoTable contains invalid Status fields")
+		writeErrorResponse(logger, w, r, http.StatusInternalServerError, "invalid prestoTable.Status fields: %v", err)
+		return
+	}
 	results, err := srv.reportResultsGetter.GetReportResults(tableName, prestoColumns)
 	if err != nil {
 		logger.WithError(err).Errorf("failed to perform presto query")
--- a/pkg/operator/promsum.go
+++ b/pkg/operator/promsum.go
@@ -207,7 +207,10 @@ func (op *Reporting) importPrometheusForTimeRange(ctx context.Context, namespace
 				"reportDataSource": reportDataSource.Name,
 				"tableName":        prestoTable.Status.TableName,
 			})
-			importCfg := op.newPromImporterCfg(reportDataSource, reportDataSource.Spec.PrometheusMetricsImporter.Query, prestoTable)
+			importCfg, err := op.newPromImporterCfg(reportDataSource, reportDataSource.Spec.PrometheusMetricsImporter.Query, prestoTable)
+			if err != nil {
+				return err
+			}
 			// ignore any global ImportFrom configuration since this is an
 			// on-demand import
 			importCfg.ImportFromTime = nil
@@ -270,7 +273,7 @@ func (op *Reporting) getQueryIntervalForReportDataSource(reportDataSource *cbTyp
 	return queryInterval
 }
 
-func (op *Reporting) newPromImporterCfg(reportDataSource *cbTypes.ReportDataSource, query string, prestoTable *cbTypes.PrestoTable) prestostore.Config {
+func (op *Reporting) newPromImporterCfg(reportDataSource *cbTypes.ReportDataSource, query string, prestoTable *cbTypes.PrestoTable) (prestostore.Config, error) {
 	chunkSize := op.cfg.PrometheusQueryConfig.ChunkSize.Duration
 	stepSize := op.cfg.PrometheusQueryConfig.StepSize.Duration
 
@@ -301,7 +304,10 @@ func (op *Reporting) newPromImporterCfg(reportDataSource *cbTypes.ReportDataSour
 	// it would take to chunk up our MaxQueryRangeDuration.
 	defaultMaxPromTimeRanges := int64(op.cfg.PrometheusDataSourceMaxQueryRangeDuration / chunkSize)
 
-	tableName := reportingutil.FullyQualifiedTableName(prestoTable)
+	tableName, err := reportingutil.FullyQualifiedTableName(prestoTable)
+	if err != nil {
+		return prestostore.Config{}, err
+	}
 
 	return prestostore.Config{
 		PrometheusQuery:           query,
@@ -312,7 +318,7 @@ func (op *Reporting) newPromImporterCfg(reportDataSource *cbTypes.ReportDataSour
 		MaxQueryRangeDuration:     op.cfg.PrometheusDataSourceMaxQueryRangeDuration,
 		MaxBackfillImportDuration: op.cfg.PrometheusDataSourceMaxBackfillImportDuration,
 		ImportFromTime:            op.cfg.PrometheusDataSourceGlobalImportFromTime,
-	}
+	}, nil
 }
 
 func (op *Reporting) newPromImporter(logger logrus.FieldLogger, reportDataSource *cbTypes.ReportDataSource, prestoTable *cbTypes.PrestoTable, cfg prestostore.Config) (*prestostore.PrometheusImporter, error) {
--- a/pkg/operator/reporting/templates.go
+++ b/pkg/operator/reporting/templates.go
@@ -9,14 +9,16 @@ import (
 	"time"
 
 	"github.com/Masterminds/sprig"
+	metering "github.com/operator-framework/operator-metering/pkg/apis/metering/v1alpha1"
 	"k8s.io/apimachinery/pkg/util/sets"
 
-	metering "github.com/operator-framework/operator-metering/pkg/apis/metering/v1alpha1"
 	"github.com/operator-framework/operator-metering/pkg/operator/prestostore"
 	"github.com/operator-framework/operator-metering/pkg/operator/reportingutil"
 	"github.com/operator-framework/operator-metering/pkg/presto"
 )
 
+// ReportQueryTemplateContext is used to hold all information about a ReportQuery that will be
+// needed when rendering the templating inside of a ReportQuery's query field.
 type ReportQueryTemplateContext struct {
 	Namespace      string
 	Query          string
@@ -28,16 +30,23 @@ type ReportQueryTemplateContext struct {
 	PrestoTables      []*metering.PrestoTable
 }
 
+// TemplateContext is the context passed to each template and contains variables related to the Report
+// and ReportQuery such as inputs and dependencies on other resources.
 type TemplateContext struct {
 	Report ReportTemplateInfo
 }
 
+// ReportTemplateInfo contains the variables for the Reprot being executed with this template.
 type ReportTemplateInfo struct {
 	ReportingStart *time.Time
 	ReportingEnd   *time.Time
 	Inputs         map[string]interface{}
 }
 
+// dataSourceTableName is a receiver method for ReportQueryTemplateContext, which validates that
+// certain fields in the ctx.DataSources are properly set. This returns the name of the Presto Table
+// that the DataSource references (DataSource.Status.TableRef) and nil, or an empty string and an error
+// if the TableRef.Name is unset, or unable to be found in the ctx.PrestoTables.
 func (ctx *ReportQueryTemplateContext) dataSourceTableName(name string) (string, error) {
 	for _, ds := range ctx.ReportDataSources {
 		if ds.Name == name {
@@ -46,7 +55,7 @@ func (ctx *ReportQueryTemplateContext) dataSourceTableName(name string) (string,
 			}
 			for _, prestoTable := range ctx.PrestoTables {
 				if prestoTable.Name == ds.Status.TableRef.Name {
-					return reportingutil.FullyQualifiedTableName(prestoTable), nil
+					return reportingutil.FullyQualifiedTableName(prestoTable)
 				}
 			}
 			return "", fmt.Errorf("tableRef PrestoTable %s not found", ds.Status.TableRef.Name)
@@ -55,6 +64,10 @@ func (ctx *ReportQueryTemplateContext) dataSourceTableName(name string) (string,
 	return "", fmt.Errorf("ReportDataSource %s dependency not found", name)
 }
 
+// reportTableName is a receiver method for ReportQueryTemplateContext, which validates that
+// that certain fields in ctx.Reports are properly set. This returns the name of the Presto Table
+// that the Report references (Report.Status.TableRef) and nil, or an empty string and an error
+// if the TableRef.Name is unset, or unable to be found in the ctx.PrestoTables
 func (ctx *ReportQueryTemplateContext) reportTableName(name string) (string, error) {
 	for _, r := range ctx.Reports {
 		if r.Name == name {
@@ -63,7 +76,7 @@ func (ctx *ReportQueryTemplateContext) reportTableName(name string) (string, err
 			}
 			for _, prestoTable := range ctx.PrestoTables {
 				if prestoTable.Name == r.Status.TableRef.Name {
-					return reportingutil.FullyQualifiedTableName(prestoTable), nil
+					return reportingutil.FullyQualifiedTableName(prestoTable)
 				}
 			}
 			return "", fmt.Errorf("tableRef PrestoTable %s not found", r.Status.TableRef.Name)
@@ -72,6 +85,10 @@ func (ctx *ReportQueryTemplateContext) reportTableName(name string) (string, err
 	return "", fmt.Errorf("Report %s dependency not found", name)
 }
 
+// renderReportQuery takes two parameters: a string parameter referencing a ReportQuery's name, and a TemplateContext
+// parameter, which is typically just `.` in the template. If the tmplCtx.ReportQuery is valid, this returns
+// a string containing the specified ReportQuery in its rendered form, using the second argument as the context
+// for templating rendering, or an error due to an unkown ReportQuery name or the inability to render that query.
 func (ctx *ReportQueryTemplateContext) renderReportQuery(name string, tmplCtx TemplateContext) (string, error) {
 	var reportQuery *metering.ReportQuery
 	for _, q := range ctx.ReportQueries {
@@ -113,6 +130,9 @@ func (ctx *ReportQueryTemplateContext) newQueryTemplate() (*template.Template, e
 	return tmpl, nil
 }
 
+// RenderQuery creates a new query template by calling the ctx parameter's method, newQueryTemplate,
+// and checks if the returned error is nil. If nil, return an empty string and the error, else return
+// the function call to renderTemplate, passing in the new query template and tmplCtx parameter.
 func RenderQuery(ctx *ReportQueryTemplateContext, tmplCtx TemplateContext) (string, error) {
 	requiredInputs := sets.NewString(ctx.RequiredInputs...)
 	givenInputs := sets.NewString()
@@ -130,6 +150,7 @@ func RenderQuery(ctx *ReportQueryTemplateContext, tmplCtx TemplateContext) (stri
 	if err != nil {
 		return "", err
 	}
+
 	return renderTemplate(tmpl, tmplCtx)
 }
 
@@ -142,6 +163,8 @@ func renderTemplate(tmpl *template.Template, tmplCtx TemplateContext) (string, e
 	return buf.String(), nil
 }
 
+// TimestampFormat checks the type of the input interface parameter and returns that parameter in
+// the form specified by the format string parameter, or an error if it's not able to be converted.
 func TimestampFormat(input interface{}, format string) (string, error) {
 	var err error
 	var d time.Time
@@ -161,10 +184,12 @@ func TimestampFormat(input interface{}, format string) (string, error) {
 	return d.Format(format), err
 }
 
+// PrometheusMetricPartitionFormat is a helper function that returns the Prometheus timestamp partition format
 func PrometheusMetricPartitionFormat(input interface{}) (string, error) {
 	return TimestampFormat(input, prestostore.PrometheusMetricTimestampPartitionFormat)
 }
 
+// PrestoTimestamp is a helper function that returns the Presto timestamp format
 func PrestoTimestamp(input interface{}) (string, error) {
 	return TimestampFormat(input, presto.TimestampFormat)
 }
--- a/pkg/operator/reporting/templates.go
+++ b/pkg/operator/reporting/templates.go
@@ -11,7 +11,7 @@ import (
 	"github.com/Masterminds/sprig"
 	"k8s.io/apimachinery/pkg/util/sets"
 
-	cbTypes "github.com/operator-framework/operator-metering/pkg/apis/metering/v1alpha1"
+	metering "github.com/operator-framework/operator-metering/pkg/apis/metering/v1alpha1"
 	"github.com/operator-framework/operator-metering/pkg/operator/prestostore"
 	"github.com/operator-framework/operator-metering/pkg/operator/reportingutil"
 	"github.com/operator-framework/operator-metering/pkg/presto"
@@ -22,10 +22,10 @@ type ReportQueryTemplateContext struct {
 	Query          string
 	RequiredInputs []string
 
-	Reports           []*cbTypes.Report
-	ReportQueries     []*cbTypes.ReportQuery
-	ReportDataSources []*cbTypes.ReportDataSource
-	PrestoTables      []*cbTypes.PrestoTable
+	Reports           []*metering.Report
+	ReportQueries     []*metering.ReportQuery
+	ReportDataSources []*metering.ReportDataSource
+	PrestoTables      []*metering.PrestoTable
 }
 
 type TemplateContext struct {
@@ -73,7 +73,7 @@ func (ctx *ReportQueryTemplateContext) reportTableName(name string) (string, err
 }
 
 func (ctx *ReportQueryTemplateContext) renderReportQuery(name string, tmplCtx TemplateContext) (string, error) {
-	var reportQuery *cbTypes.ReportQuery
+	var reportQuery *metering.ReportQuery
 	for _, q := range ctx.ReportQueries {
 		if q.Name == name {
 			reportQuery = q
--- a/pkg/operator/reporting/templates_test.go
+++ b/pkg/operator/reporting/templates_test.go
@@ -1,0 +1,615 @@
+package reporting
+
+import (
+	"fmt"
+	"testing"
+	"time"
+
+	"github.com/operator-framework/operator-metering/pkg/apis/metering/v1alpha1"
+	metering "github.com/operator-framework/operator-metering/pkg/apis/metering/v1alpha1"
+	"github.com/operator-framework/operator-metering/pkg/operator/prestostore"
+	"github.com/operator-framework/operator-metering/pkg/presto"
+	"github.com/operator-framework/operator-metering/test/testhelpers"
+	"github.com/stretchr/testify/assert"
+	v1 "k8s.io/api/core/v1"
+	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func TestRenderQuery(t *testing.T) {
+	const (
+		testNamespace   = "default"
+		testReportName  = "namespace-cpu-usage"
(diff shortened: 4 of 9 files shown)


pkg/operator/reporting/templates.go
```diff
@@ -42,7 +43,12 @@ func (ctx *ReportQueryTemplateContext) dataSourceTableName(name string) (string,
 			}
 			for _, prestoTable := range ctx.PrestoTables {
 				if prestoTable.Name == ds.Status.TableRef.Name {
-					return reportingutil.FullyQualifiedTableName(prestoTable), nil
+					dsTableName := reportingutil.FullyQualifiedTableName(prestoTable)
+
+					if strings.HasPrefix(dsTableName, ".") || strings.HasSuffix(dsTableName, ".") || strings.Contains(dsTableName, "..") {
```
chancez: You don't need to do string matching, you can just return an error from `FullyQualifiedTableName` instead and return that here.

pkg/operator/reporting/templates_test.go
```diff
@@ -0,0 +1,255 @@
...
+			templateContext: TemplateContext{
+				Report: ReportTemplateInfo{
+					ReportingStart: reportStart,
+					ReportingEnd:   reportEnd,
+					Inputs:         map[string]interface{}{"UnreadyDeploymentReplicas": ""},
+				},
+			},
+			expectErr:    true,
+			expectErrMsg: "error executing template: template: reportQueryTemplate:1:36: executing \"reportQueryTemplate\" at <dataSourceTableName ...>: error calling dataSourceTableName: ReportDataSource  dependency not found",
+			expectOutput: "",
```
chancez: nit: You don't need to set `expectOutput` to "", that's the default, and since we're erroring, it won't be checked anyways

pkg/operator/reporting/templates_test.go
```diff
@@ -0,0 +1,255 @@
...
+}
+
+// NewPrestoTableCustom allows us to insert a custom name, as opposed to testhelpers.NewPrestoTable
+func NewPrestoTableCustom(name, namespace, schema string, columns []presto.Column) *v1alpha1.PrestoTable {
+	return &v1alpha1.PrestoTable{
+		ObjectMeta: meta.ObjectMeta{
+			Name:      name,
+			Namespace: namespace,
+		},
+		Status: v1alpha1.PrestoTableStatus{
```
chancez: Spec should probably match

pkg/operator/reporting/templates_test.go
```diff
@@ -0,0 +1,255 @@
...
+			} else {
+				assert.NoErrorf(t, err, "expected the report would return no error, but got an error.")
+				assert.EqualValuesf(t, testCase.expectOutput, output, "expected that RenderQuery would return the expected output")
+			}
+		})
+	}
+}
+
+// NewPrestoTableCustom allows us to insert a custom name, as opposed to testhelpers.NewPrestoTable
+func NewPrestoTableCustom(name, namespace, schema string, columns []presto.Column) *v1alpha1.PrestoTable {
```
chancez: Yeah let's just put this into testhelpers, remove the `Custom` from the name

pkg/operator/reporting/templates_test.go
```diff
@@ -0,0 +1,255 @@
...
+			Catalog:   "hive",
+			Schema:    schema,
+			TableName: name,
+			Columns:   columns,
+		},
+	}
+}
+
+// NewReportQueryCustomQuery helps create a ReportQuery with a custom ReportDataSource and adds a Query field
+func NewReportQueryCustom(dataSource *metering.ReportDataSource, query string) *metering.ReportQuery {
```
chancez: Probably move to testhelpers, remove the `Custom` at the end

pkg/operator/reporting/templates_test.go
```diff
@@ -0,0 +1,255 @@
...
+		return &v
+	}
+
+	return &metering.ReportQuery{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "unready-deployment-replicas",
+			Namespace: testNamespace,
+		},
+		Spec: metering.ReportQuerySpec{
+			Inputs: []metering.ReportQueryInputDefinition{
```
chancez: This could probably just be passed in as an argument right?

pkg/operator/reporting/templates_test.go
timflannagan1: I did that initially and there was an import cycle w/ the reporting package and testhelpers

pkg/operator/reporting/templates_test.go
chancez: And you could have a function that creates an InputDefinition given a DataSource which you can use to create the input before passing it in.

pkg/operator/reporting/templates_test.go
timflannagan1: Fixed.

pkg/operator/reporting/templates.go
timflannagan1: Fixed.

pkg/operator/reporting/templates_test.go
chancez: Ah. Yeah, our packages need to be reorganized. Well for now you can keep it here, but don't export it, and perhaps put the word test into the function name., eg: `newTestPrestoTable`

timflannagan1: Updated - renamed to include test and unexported the functions.

pkg/operator/reporting/templates_test.go
timflannagan1: Updated - I added the Spec field definition to the `PrestoTable`, which mirrors the Status field definition.

pkg/operator/reporting/templates_test.go
timflannagan1: Updated to include `newTestInputDefinition`. may need to move `newDefault` into its own function or deflate it into the `Default` field

pkg/operator/http.go
```diff
@@ -227,7 +227,7 @@ func (srv *server) getReport(logger log.FieldLogger, name, namespace, format str
 		logger.Debugf("mismatched columns, PrestoTable columns: %v, ReportQuery columns: %v", prestoColumns, queryPrestoColumns)
 	}
 
-	tableName := reportingutil.FullyQualifiedTableName(prestoTable)
+	tableName, _ := reportingutil.FullyQualifiedTableName(prestoTable)
```
chancez: We shouldn't be ignoring these errors.

pkg/operator/promsum.go
```diff
@@ -301,7 +301,7 @@ func (op *Reporting) newPromImporterCfg(reportDataSource *cbTypes.ReportDataSour
 	// it would take to chunk up our MaxQueryRangeDuration.
 	defaultMaxPromTimeRanges := int64(op.cfg.PrometheusDataSourceMaxQueryRangeDuration / chunkSize)
 
-	tableName := reportingutil.FullyQualifiedTableName(prestoTable)
+	tableName, _ := reportingutil.FullyQualifiedTableName(prestoTable)
```
chancez: Don't ignore the err

pkg/operator/reportingutil/util.go
```diff
@@ -35,8 +35,11 @@ func AWSBillingPeriodTimestamp(date time.Time) string {
 	return date.Format(AWSUsagePartitionDateStringLayout)
 }
 
-func FullyQualifiedTableName(prestoTable *cbTypes.PrestoTable) string {
-	return presto.FullyQuaifiedTableName(prestoTable.Status.Catalog, prestoTable.Status.Schema, prestoTable.Status.TableName)
+func FullyQualifiedTableName(prestoTable *cbTypes.PrestoTable) (string, error) {
+	if len(prestoTable.Status.Catalog) == 0 || len(prestoTable.Status.Schema) == 0 || len(prestoTable.Status.TableName) == 0 {
```
chancez: Instead of a single if, use 3 for each case, and give a more specific error message for each missing field. You can even accumulate errors into a slice, and if it's non-empty, join the errors together into one.

eg:
```
...
var errs []string
if len(prestoTable.Status.Catalog) == 0 {
  errs = append(errs, "catalog is unset")
}
if len(prestoTable.Status.Schema) == 0 {
  errs = append(errs, "schema is unset")
}
...

if len(errs) != 0) {
  return fmt.Errorf("PrestoTable status is invalid: %s", strings.Join(errs, ", "))
}
...

File: pkg/operator/reporting/templates_test.go [hunk 10]
```diff
@@ -0,0 +1,525 @@
...
+	ds1.Status.TableRef = v1.LocalObjectReference{Name: "initialized_datasource"}
+	ds2 := testhelpers.NewReportDataSource("datasource2", testNamespace)
+
+	testValidQuery := &metering.ReportQuery{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      testReportQuery,
+			Namespace: testNamespace,
+		},
+		Spec: metering.ReportQuerySpec{
+			Inputs: []metering.ReportQueryInputDefinition{
```
chancez: These inputs aren't used by the query, so remove them

pkg/operator/reporting/templates_test.go
```diff
@@ -0,0 +1,525 @@
...
+			Columns:   columns,
+		},
+	}
+}
+
+// newReportQueryCustomQuery helps create a ReportQuery with a custom ReportDataSource and adds a Query field
+func newTestReportQuery(dataSource *metering.ReportDataSource, query string) *metering.ReportQuery {
+	return &metering.ReportQuery{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      testReportName,
```
chancez: Nit: globals are generally speaking, not good, so maybe these should be arguments to the function

pkg/operator/reporting/templates_test.go
```diff
@@ -0,0 +1,525 @@
...
+	metering "github.com/operator-framework/operator-metering/pkg/apis/metering/v1alpha1"
+	"github.com/operator-framework/operator-metering/pkg/presto"
+	"github.com/operator-framework/operator-metering/test/testhelpers"
+	"github.com/stretchr/testify/assert"
+	v1 "k8s.io/api/core/v1"
+	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
```
chancez: Perhaps these could be moved into `TestRenderQuery` instead of having globals

pkg/operator/reporting/templates_test.go
```diff
@@ -0,0 +1,525 @@
...
+		reportTemplate  *ReportQueryTemplateContext
+		templateContext TemplateContext
+		expectErr       bool
+		expectErrMsg    string
+		expectOutput    string
+	}{
+		{
+			name: "valid report query with no templating returns nil and the expected query output",
+			reportTemplate: &ReportQueryTemplateContext{
+				ReportQuery: newTestReportQuery(ds1, "show tables;"),
```
chancez: Couldn't this use `testValidQuery`?

pkg/operator/reporting/templates_test.go
```diff
@@ -0,0 +1,525 @@
...
+			},
+			templateContext: TemplateContext{
+				Report: ReportTemplateInfo{
+					Inputs: map[string]interface{}{"UnreadyDeploymentReplicas": ds1.Name},
+				},
+			},
+			expectOutput: fmt.Sprintf("SELECT * FROM hive.default.%s", ds1.Status.TableRef.Name),
+		},
+		{
+			name: "invalid report query with invalid templating (missing Inputs field) returns error and the expected error message",
```
chancez: You can probably remove the ` and the expected error message` from all the names

pkg/operator/reporting/templates_test.go
```diff
@@ -0,0 +1,525 @@
...
+					Inputs: map[string]interface{}{"UnreadyDeploymentReplicas": ds1.Name},
+				},
+			},
+			expectErr:    true,
+			expectErrMsg: fmt.Sprintf("error executing template: template: reportQueryTemplate:1:17: executing \"reportQueryTemplate\" at <dataSourceTableName ...>: error calling dataSourceTableName: tableRef PrestoTable %s not found", ds1.Status.TableRef.Name),
+		},
+		{
+			name: "ReportDataSource dependencies was not found (due to empty ds.Name) results in error and the expected error message",
+			reportTemplate: &ReportQueryTemplateContext{
+				ReportQuery:       newTestReportQuery(ds1, "SELECT labels['namespace'], FROM {| dataSourceTableName .Report.Inputs.UnreadyDeploymentReplicas |}"),
```
chancez: Any reason some use `SELECT labels['namespace']` and some use `SELECT *`?  Maybe all of it should be consistent and just be `SELECT *`?

timflannagan1: I thought I had changed all the instances. I'll update.

pkg/operator/promsum.go
timflannagan1: Updated all references.

pkg/operator/reporting/templates_test.go
```diff
@@ -0,0 +1,569 @@
...
+					Inputs: map[string]interface{}{"UnreadyDeploymentReplicas": testReportQuery},
+				},
+			},
+			expectErr:    true,
+			expectErrMsg: fmt.Sprintf("error executing template: template: reportQueryTemplate:1:17: executing \"reportQueryTemplate\" at <dataSourceTableName ...>: error calling dataSourceTableName: ReportDataSource %s dependency not found", testReportQuery),
+		},
+		{
+			name: "ReportDataSource dependencies were not found (due to Report.Inputs referencing an integer) results in error",
+			reportTemplate: &ReportQueryTemplateContext{
+				ReportQuery:       newTestReportQuery(testReportName, testNamespace, ds1, "SELECT * FROM {| dataSourceTableName .Report.Inputs.UnreadyDeploymentReplicas |}"),
```
chancez: I was more looking for new queries that used these different types of inputs, and had different input definitions using these other types.

pkg/operator/reporting/templates_test.go
```diff
@@ -0,0 +1,569 @@
...
+			templateContext: TemplateContext{
+				Report: ReportTemplateInfo{
+					Inputs: map[string]interface{}{"UnreadyDeploymentReplicas": 5},
+				},
+			},
+			expectErr:    true,
+			expectErrMsg: "error executing template: template: reportQueryTemplate:1:44: executing \"reportQueryTemplate\" at <.Report.Inputs.Unrea...>: wrong type for value; expected string; got int",
+		},
+		{
+			name: "ReportDataSource dependencies were not found (due to Report.Inputs referencing a *time.Time object) results in error",
```
chancez: Same here, I was mostly thinking of having a new input definition expecting a `type: time` with a field name that isn't reportingStart/reportingEnd (those are special and are assumed to be time) and testing providing it/not providing it/etc.

pkg/operator/reporting/templates_test.go
```diff
@@ -0,0 +1,569 @@
...
+			templateContext: TemplateContext{
+				Report: ReportTemplateInfo{
+					Inputs: map[string]interface{}{"UnreadyDeploymentReplicas": ds1.Name},
+				},
+			},
+			expectErr:    true,
+			expectErrMsg: "error executing template: template: reportQueryTemplate:1:17: executing \"reportQueryTemplate\" at <dataSourceTableName ...>: error calling dataSourceTableName: PrestoTable status is invalid: prestoTable.Status.Schema is unset",
+		},
+		{
+			name: "Report dependency was not found (due to no reports) returns error",
```
chancez: You need a case where this passes too, and one where the input value is empty

pkg/operator/reporting/templates_test.go
```diff
@@ -0,0 +1,569 @@
...
+			reportTemplate: &ReportQueryTemplateContext{
+				ReportQuery:       newTestReportQuery(testReportName, testNamespace, ds1, "SELECT * FROM {| reportTableName .Report.Inputs.UnreadyDeploymentReplicas |}"),
+				ReportDataSources: []*metering.ReportDataSource{ds1},
```
chancez: You don't need datasources here.

pkg/operator/reporting/templates_test.go
```diff
@@ -0,0 +1,569 @@
...
+					Inputs: map[string]interface{}{"UnreadyDeploymentReplicas": ds1.Name},
+				},
+			},
+			expectErr:    true,
+			expectErrMsg: "error executing template: template: reportQueryTemplate:1:17: executing \"reportQueryTemplate\" at <dataSourceTableName ...>: error calling dataSourceTableName: PrestoTable status is invalid: prestoTable.Status.Schema is unset",
+		},
+		{
+			name: "Report dependency was not found (due to no reports) returns error",
+			reportTemplate: &ReportQueryTemplateContext{
+				ReportQuery:       newTestReportQuery(testReportName, testNamespace, ds1, "SELECT * FROM {| reportTableName .Report.Inputs.UnreadyDeploymentReplicas |}"),
```
chancez: This function should be refactored probably to take inputs, it doesn't need `ds1`

pkg/operator/reporting/templates_test.go
```diff
@@ -0,0 +1,569 @@
...
+			expectErrMsg: fmt.Sprintf("error executing template: template: reportQueryTemplate:1:17: executing \"reportQueryTemplate\" at <reportTableName .Rep...>: error calling reportTableName: Report %s dependency not found", ds1.Name),
+		},
+		{
+			name: "Report dependency was not found (due to Report.Name being empty) returns error",
+			reportTemplate: &ReportQueryTemplateContext{
+				Reports: []*metering.Report{
+					testhelpers.NewReport("", testNamespace, testReportQuery, reportStart, reportEnd, metering.ReportStatus{}, nil, false),
+				},
+				ReportQuery:       newTestReportQuery(testReportName, testNamespace, ds1, "SELECT * FROM {| reportTableName .Report.Inputs.UnreadyDeploymentReplicas |}"),
+				ReportDataSources: []*metering.ReportDataSource{ds1},
```
chancez: No needed

pkg/operator/reporting/templates_test.go
chancez: Also, all of these on reports need a `type: Report` input definition

pkg/operator/reporting/templates_test.go
```diff
@@ -0,0 +1,569 @@
...
+			},
+			templateContext: TemplateContext{
+				Report: ReportTemplateInfo{},
+			},
+			expectOutput: "SELECT * FROM test_table",
+		},
+		{
+			name: "valid report query with valid templating returns nil and the expected query output",
+			reportTemplate: &ReportQueryTemplateContext{
+				ReportQuery:       newTestReportQuery(testReportName, testNamespace, ds1, "SELECT * FROM {| dataSourceTableName .Report.Inputs.UnreadyDeploymentReplicas |}"),
```
chancez: It looks like most of these are using `testReportName` but should be using `testReportQuery` for the first arg

pkg/operator/reporting/templates_test.go
```diff
@@ -0,0 +1,569 @@
...
+	v1 "k8s.io/api/core/v1"
+	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func TestRenderQuery(t *testing.T) {
+	const (
+		testNamespace            = "default"
+		testReportName           = "unready-deployment-replicas"
+		testReportQuery          = "unready-deployment-replicas"
```
chancez: Maybe make `testReportName` and `testReportQuery` have distinct values so the possibility of mixing them up would cause an error.

pkg/operator/reporting/templates_test.go
timflannagan1: Just added some more in the most recent commit. Still might be lacking some test cases using the time type, but without using `ReportingStart` and/or `ReportingEnd`.

pkg/operator/reporting/templates_test.go
```diff
@@ -546,24 +574,22 @@ func newTestPrestoTable(name, namespace, schema, catalog string, columns []prest
...
+			Inputs: inputs,
 			Query:  query,
 		},
 	}
 }
 
-func newTestInputDefinition(dataSource *metering.ReportDataSource) metering.ReportQueryInputDefinition {
+func newTestInputDefinition(inputType string) metering.ReportQueryInputDefinition {
 	return metering.ReportQueryInputDefinition{
-		Name:     dataSource.Name,
```
chancez: Why did you get rid of name and `default`? We need those for testing

pkg/operator/reporting/templates.go
```diff
@@ -129,10 +130,32 @@ func (ctx *ReportQueryTemplateContext) newQueryTemplate() (*template.Template, e
...
+			expectedInputs.Insert(queryDef.Name)
+		}
+	}
+
+	for queryKey := range tmplCtx.Report.Inputs {
+		expectedVals.Insert(queryKey)
+	}
+
+	// check if there's a difference in input definitions and what's provided as an input value, and vice-versa
+	if len(expectedInputs.Difference(expectedVals)) != 0 || len(expectedVals.Difference(expectedInputs)) != 0 {
```
chancez: Here you could actually return an error message including the list of inputs provided that aren't expected and vice-versa. Generally, try to have more specific error messages than generic ones like the one below.

pkg/operator/reporting/templates.go
```diff
@@ -129,10 +130,32 @@ func (ctx *ReportQueryTemplateContext) newQueryTemplate() (*template.Template, e
...
 		return "", err
 	}
+
+	// validate that all required inputs were passed
+	expectedVals := sets.NewString()
+	expectedInputs := sets.NewString()
+
+	for _, queryDef := range ctx.ReportQuery.Spec.Inputs {
+		// check whether the current definition is required
+		if queryDef.Required {
```
chancez: We may need a set of required, and optional inputs and compare the provided values against each, because as-is, I believe we would get errors for providing inputs that aren't required. Tests can confirm though.

pkg/operator/reporting/templates_test.go
```diff
@@ -84,7 +97,7 @@ func TestRenderQuery(t *testing.T) {
 			name: "invalid report query with invalid templating (missing Inputs field) returns error",
 			reportTemplate: &ReportQueryTemplateContext{
 				ReportQuery: newTestReportQuery(testReportQuery, testNamespace, "SELECT * FROM {| dataSourceTableName .Report.Inputs. |}", []metering.ReportQueryInputDefinition{
-					newTestInputDefinition("ReportDataSource"),
+					newTestInputDefinition("", ""),
```
chancez: Why an empty type here? I think this should be `newTestInputDefinition("", "ReportDataSource")`?

chancez: Actually, again, why are any input definitions not specifying both? I think we should be testing where the ReportQuery itself is validated and has a name/type set in all cases.

pkg/operator/reporting/templates_test.go
chancez: This still isnt' removed

pkg/operator/reporting/templates_test.go
```diff
@@ -678,9 +711,11 @@ func newTestReportQuery(testReportName, testNamespace, query string, inputs []me
 	}
 }
 
-func newTestInputDefinition(inputType string) metering.ReportQueryInputDefinition {
+func newTestInputDefinition(inputName, inputType string) metering.ReportQueryInputDefinition {
 	return metering.ReportQueryInputDefinition{
+		Name:     inputName,
 		Type:     inputType,
 		Required: true,
+		Default:  newDefault(`"` + inputName + `"`),
```
chancez: We will want to test with an empty default, so this might not make sense. I'm not sure this function even makes perfect sense at this point

pkg/operator/reporting/templates_test.go
```diff
@@ -171,7 +184,7 @@ func TestRenderQuery(t *testing.T) {
 			name: "ReportDataSource dependencies were not found (due to Report.Inputs referencing a ReportQuery) results in error",
 			reportTemplate: &ReportQueryTemplateContext{
 				ReportQuery: newTestReportQuery(testReportQuery, testNamespace, "SELECT * FROM {| dataSourceTableName .Report.Inputs.UnreadyDeploymentReplicas |}", []metering.ReportQueryInputDefinition{
-					newTestInputDefinition("ReportQuery"),
+					newTestInputDefinition("UnreadyDeploymentReplicas", "ReportQuery"),
```
chancez: The type here doesn't match the SQL query

pkg/operator/reporting/templates.go
```diff
@@ -129,10 +130,32 @@ func (ctx *ReportQueryTemplateContext) newQueryTemplate() (*template.Template, e
...
 // the function call to renderTemplate, passing in the new query template and tmplCtx parameter.
 func RenderQuery(ctx *ReportQueryTemplateContext, tmplCtx TemplateContext) (string, error) {
+	// validate the ctx.ReportQuery.Spec.Query field and create a new template
 	tmpl, err := ctx.newQueryTemplate()
 	if err != nil {
 		return "", err
 	}
+
+	// validate that all required inputs were passed
+	expectedVals := sets.NewString()
```
chancez: rename this to `givenVals`

pkg/operator/reporting/templates_test.go
```diff
@@ -0,0 +1,615 @@
...
+					Inputs: map[string]interface{}{"EmptyPrestoTableName": ""},
+				},
+			},
+			expectErr:    true,
+			expectErrMsg: "error executing template: template: reportQueryTemplate:1:17: executing \"reportQueryTemplate\" at <dataSourceTableName ...>: error calling dataSourceTableName: ReportDataSource  dependency not found",
+		},
+		{
+			name: "Report dependency was not found (due to no reports) returns error",
+			reportTemplate: &ReportQueryTemplateContext{
+				Query:          "SELECT * FROM {| reportTableName .Report.Inputs.EmptyReportsField |}",
```
chancez: A bit confusing since this isn't an empty reports input

pkg/operator/reporting/templates_test.go
```diff
@@ -0,0 +1,615 @@
...
+					Inputs: map[string]interface{}{"ReportNameDoesNotMatchInputName": "this-does-not-match"},
+				},
+			},
+			expectErr:    true,
+			expectErrMsg: "error executing template: template: reportQueryTemplate:1:17: executing \"reportQueryTemplate\" at <reportTableName .Rep...>: error calling reportTableName: Report this-does-not-match dependency not found",
+		},
+		{
+			name: "invalid reportTableName templating (Status.TableRef.Name is empty) results in error and the expected errror output",
+			reportTemplate: &ReportQueryTemplateContext{
+				Query:          "SELECT * FROM {| reportTableName .Report.Inputs.EmptyReportStatusTableRef|}",
```
chancez: This one isn't empty either, just missing

pkg/operator/reporting/templates_test.go
```diff
@@ -0,0 +1,615 @@
...
+			templateContext: TemplateContext{
+				Report: ReportTemplateInfo{
+					Inputs: map[string]interface{}{"MissingTemplateContext": testReportQuery},
+				},
+			},
+			expectErr:    true,
+			expectErrMsg: fmt.Sprintf("error executing template: template: reportQueryTemplate:1:17: executing \"reportQueryTemplate\" at <renderReportQuery .R...>: error calling renderReportQuery: unable to render query %s, err: error executing template: template: reportQueryTemplate:1:42: executing \"reportQueryTemplate\" at <.Report.Inputs.Query...>: invalid value; expected string", testReportQuery),
+		},
+		{
+			name: "stub",
```
chancez: needs a real name

pkg/operator/reporting/templates_test.go
```diff
@@ -0,0 +1,615 @@
...
+					Inputs: map[string]interface{}{"MissingReportStatusTableRef": testReportName},
+				},
+			},
+			expectErr:    true,
+			expectErrMsg: fmt.Sprintf("error executing template: template: reportQueryTemplate:1:17: executing \"reportQueryTemplate\" at <reportTableName .Rep...>: error calling reportTableName: %s tableRef is empty", testReportName),
+		},
+		{
+			name: "query's table ref and presto table don't match up returns error",
+			reportTemplate: &ReportQueryTemplateContext{
+				Query:          "SELECT * FROM {| reportTableName .Report.Inputs.EmptyReportStatusTableRef |}",
```
chancez: Miss this, not empty, just missing a presto table

PR #711, opened 2019-05-09
Remove dependency on tiller and replace deploy logic with scripts using helm template and kubectl

chancez:
Adds a deploy-resources.sh script which deploys all our existing openshift-metering templates using helm template and kubectl apply instead of helm upgrade and tiller.

Updates run-operator.sh to use deploy-resources.sh instead of helm upgrade to deploy without tiller, and removes tiller from the metering-helm-operator chart, as well as the rendered manifests.


chancez:
/hold
Based on top of #709, waiting for that to merge.

PR #713, opened 2019-05-13
Documentation: Fix links that weren't updated after last architecture change.

timflannagan1:


timflannagan1:
@chancez I saw that `Documentation/reportqueries.md` referenced `string` as an input type instead of Presto's varchar. Was this a correct change to make?

chancez:
@timflannagan1 hmm it should use varchar probably. Whoops.

PR #714, opened 2019-05-13
charts: Fix custom Prometheus CA configmap

chancez:
The value must be quoted since it is multiple lines


chancez:
cc @pruan-rht this is the fix

pruan-rht:
@chancez looks good, thanks!

PR #718, opened 2019-05-14
pkg/operator: Fix log warning due to case mismatch in ReportQuery columns and PrestoTable columns

chancez:


chancez:
/hold

Didn't work, still working on this

PR #723, opened 2019-05-17
Replace metering-helm-operator with metering-ansible-operator

chancez:
This contains a few changes:

- Replaces metering-helm-operator with metering-ansible-operator
- Effectively, removes the metering-helm-operator docker image (the Dockerfile for is just `FROM scratch` until we remove it from the CI build configs).
- Updates the metering-ansible-operator to use ansible tasks over shell scripts for deploying resources
- Renames our Metering CRD to MeteringConfig (this was planned and we setup the ansible based operator to use MeteringConfig).
- Replaces the ao-logs script with our own version to enable streaming of playbook logs, and because the ansible-operator base image in CI isn't up to date with the operator-sdk repo, and thus, is missing the script and dependencies for it. Related to this, we install inotify-tools because the ansible-opertor base image in CI isn't updated to contain that package.
- Updates our metering-ansible-operator to use a patched version of the openshift-restclient-python package to work-around a bug that breaks the ability for us to use k8s_facts (items returned sometimes were missing a `kind` field, which we need). Upstream PR to fix is here: https://github.com/openshift/openshift-restclient-python/pull/298
- Fixes a bug in the prune labeling for cluster scoped resources to ensure we don't delete resources not related to a given Metering install.


chancez:
/hold

Ready for review but before we merge I’d like for my PR for openshift-rest-client to merge and I also want to find out how long before that makes its way into pypi and the internal RPM repos.

images/metering-ansible-operator/roles/meteringconfig/tasks/deploy_resource.yml
```diff
@@ -0,0 +1,46 @@
...
+      vars:
+        to_create_resources: "{{ result.stdout | from_yaml_all | list }}"
+      loop: "{{ to_create_resources }}"
+      when: item is not none
+      include_tasks: create_resource.yml
+
+- when: not create
+  block:
+    - name: Query for {{ apis | map(attribute='kind') | join(', ') }} resources with selector {{ label_selector}} to delete
+      vars:
```
timflannagan1: `when` conditionals are typically placed at the bottom of blocks. https://docs.ansible.com/ansible/latest/user_guide/playbooks_blocks.html

images/metering-ansible-operator/roles/meteringconfig/tasks/deploy_resource.yml
```diff
@@ -0,0 +1,46 @@
...
+        api_version: "{{ item.api_version | default(omit) }}"
+        kind: "{{ item.kind }}"
+        namespace: "{{ namespace }}"
+        label_selectors:
+          - "{{ label_selector }}"
+      loop: "{{ apis }} "
+      register: to_delete
+
+    - name: Delete {{ apis | map(attribute='kind') | join(', ') }} resources with selector {{ label_selector}}
+      vars:
```
timflannagan1: this layout is different than the task block: `Deploy resources`, which has a similar structure

images/metering-ansible-operator/roles/meteringconfig/tasks/reconcile.yml
```diff
@@ -0,0 +1,13 @@
+---
+
+- name: Store MeteringConfig spec into values file
+  copy: content="{{ meteringconfig_spec }}" dest=/tmp/metering-values.yaml
+
+- include_tasks: "{{ item }}"
```
timflannagan1: what about something like this:
```
(defaults/main.yml)
custom_resources: ['monitoring', 'hdfs', 'hive', 'presto', 'reporting_operator', 'reporting']
...
- name: reconcile resources
  include_tasks: "reconcile_{{ resource_type }}.yml"
  loop: "{{ custom_resources }}"
  loop_control:
    loop_var: resource_type
```

PR #725, opened 2019-05-21
metering-ansible-operator: Run deploy-resources.sh in meteringconfig role

chancez:
Manually tested by editing the deployment to use my own custom image of metering-ansible-operator.
This updates metering-ansible-operator to run the deploy-resources.sh script that metering-helm-operator currently uses to deploy the metering stack.

As a follow up, this script can be translated into ansible roles and tasks that accomplish the same logic directly.

### DIFF
--- a/images/metering-ansible-operator/roles/meteringconfig/tasks/main.yml
+++ b/images/metering-ansible-operator/roles/meteringconfig/tasks/main.yml
@@ -1,4 +1,13 @@
 ---
-- name: start metering
-  debug:
-    msg: "Hello World (non default)"
+
+- name: Extract MeteringConfig spec
+  copy: content="{{ _metering_openshift_io_meteringconfig.spec }}" dest=/tmp/metering-values.yaml
+
+- name: Install Metering
+  shell: |
+    /opt/ansible/scripts/deploy-resources.sh -f /tmp/metering-values.yaml {{ meta.name }} {{ lookup('env','HELM_CHART_PATH') }} {{ meta.namespace }} metering.openshift.io meteringconfig {{ meta.name }} {{ _metering_openshift_io_meteringconfig.metadata.uid }} metering.openshift.io/prune false 2>&1 | tee /tmp/deploy-resources.txt
+  register: install_result
+
+- debug: msg="{{item}}"
+  with_items: "{{install_result.stdout_lines}}"
+  when: install_result.stdout_lines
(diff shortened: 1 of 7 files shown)


images/metering-ansible-operator/roles/meteringconfig/tasks/main.yml
```diff
@@ -1,4 +1,13 @@
 ---
-- name: start metering
-  debug:
-    msg: "Hello World (non default)"
+
+- name: Extract MeteringConfig spec
+  copy: content="{{ _metering_openshift_io_meteringconfig.spec }}" dest=/tmp/metering-values.yaml
+
+- name: Install Metering
+  shell: |
```
timflannagan1: `command` is typically the preferred module

images/metering-ansible-operator/roles/meteringconfig/tasks/main.yml
```diff
@@ -1,4 +1,13 @@
 ---
-- name: start metering
-  debug:
-    msg: "Hello World (non default)"
+
+- name: Extract MeteringConfig spec
+  copy: content="{{ _metering_openshift_io_meteringconfig.spec }}" dest=/tmp/metering-values.yaml
```
timflannagan1: might want to register the return code in the case where this fails

chancez: If it fails, it'll just stop the role here and return the error.

images/metering-ansible-operator/roles/meteringconfig/tasks/main.yml
chancez: I'm using a pipe here `|`, so I need `shell`

chancez: And I'm also redirect both stdout and stderr to the pipe (normally it's just stdout) using `2>&1`

images/metering-ansible-operator/roles/meteringconfig/tasks/main.yml
timflannagan1: I would also move the `content={{...` and `dest=...` to their own lines. i.e.
```
copy:
  content: "{{ _metering_openshift_io_meteringconfig.spec }}"
  dest: /tmp/metering-values.yaml
```

chancez: But why? Inline is shorter and not very long.

PR #728, opened 2019-05-21
Support configuring default component images from environment variables to metering-ansible-operator

chancez:
The marketplace publishing pipeline managed by RCM/ART does string replacement of the image values in the CSV manifest, and currently we expose those as environment variables to the metering-ansible-operator for this exact purpose.
This updates the role to use those variables as the default values for our component images.


chancez:
/hold cancel

This is ready. I tried to minimize it down to the essentials. I have some follow up PRs to update some scripts/env vars to allow adjusting these when running metering locally also.

PR #729, opened 2019-05-23
charts/openshift-metering: Update the reporting-operator and Presto StatefulSets to use TLS when specified

timflannagan1:
### Overview:
Currently, the presto coordinator and worker pods communicate using http. With this change, the presto components can communicate using https when specified in the configurations, or by default, http.

You would need to specify key/cert pem files, which are then used to create a java keystore file and provided to presto.

### DIFF
--- a/charts/openshift-metering/templates/presto/presto-coordinator-config.yaml
+++ b/charts/openshift-metering/templates/presto/presto-coordinator-config.yaml
@@ -13,13 +13,29 @@ data:
     node.environment=production
 
   config.properties: |
+    coordinator=true
+    node-scheduler.include-coordinator={{ .Values.presto.spec.presto.config.nodeSchedulerIncludeCoordinator }}
+    discovery-server.enabled=true
+
+{{- if .Values.presto.spec.presto.config.tls.enabled }}
+    http-server.http.enabled=false
+    http-server.https.enabled=true
+    http-server.https.port=8080
+    http-server.https.keystore.path=/opt/presto/tls/keystore.jks
+    http-server.https.keystore.key=password
+    node.internal-address-source=FQDN
+    discovery.uri=https://presto:8080
+    internal-communication.https.required=true
+    internal-communication.https.keystore.path=/opt/presto/tls/keystore.jks
+    internal-communication.https.keystore.key=password
+{{- else }}
     http-server.http.port=8080
+    discovery.uri=http://presto:8080
+{{- end }}
+
     jmx.rmiserver.port=8081
     jmx.rmiregistry.port=8081
-    coordinator=true
-    discovery-server.enabled=true
-    discovery.uri={{ .Values.presto.spec.presto.config.discoveryURI }}
-    node-scheduler.include-coordinator={{ .Values.presto.spec.presto.config.nodeSchedulerIncludeCoordinator }}
+
 {{- if .Values.presto.spec.presto.config.maxQueryLength }}
     query.max-length={{ .Values.presto.spec.presto.config.maxQueryLength }}
 {{- end }}
--- a/charts/openshift-metering/templates/presto/presto-coordinator-statefulset.yaml
+++ b/charts/openshift-metering/templates/presto/presto-coordinator-statefulset.yaml
@@ -1,17 +1,17 @@
 apiVersion: apps/v1
 kind: StatefulSet
 metadata:
-  name: presto-worker
+  name: presto-coordinator
   labels:
     app: presto
-    presto: worker
+    presto: coordinator
 spec:
-  replicas: {{ .Values.presto.spec.presto.worker.replicas }}
+  replicas: 1
   serviceName: presto-nodes
   selector:
     matchLabels:
       app: presto
-      presto: worker
+      presto: coordinator
 {{- if .Values.presto.spec.presto.labels }}
 {{ toYaml .Values.presto.spec.presto.labels | indent 6 }}
 {{- end }}
@@ -19,12 +19,12 @@ spec:
     metadata:
       labels:
         app: presto
-        presto: worker
+        presto: coordinator
 {{- if .Values.presto.spec.presto.labels }}
 {{ toYaml .Values.presto.spec.presto.labels | indent 8 }}
 {{- end }}
       annotations:
-        presto-worker-config-hash: {{ include (print $.Template.BasePath "/presto/presto-worker-config.yaml") . | sha256sum }}
+        presto-coordinator-config-hash: {{ include (print $.Template.BasePath "/presto/presto-coordinator-config.yaml") . | sha256sum }}
         presto-common-config-hash: {{ include (print $.Template.BasePath "/presto/presto-common-config.yaml") . | sha256sum }}
         presto-catalog-config-hash: {{ include (print $.Template.BasePath "/presto/presto-catalog-config-secret.yaml") . | sha256sum }}
         presto-jmx-config-hash: {{ include (print $.Template.BasePath "/presto/presto-jmx-config.yaml") . | sha256sum }}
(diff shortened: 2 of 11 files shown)


charts/openshift-metering/templates/presto/presto-worker-config.yaml
```diff
@@ -13,12 +13,29 @@ data:
     node.environment=production
 
   config.properties: |
+    coordinator=false
+    node-scheduler.include-coordinator={{ .Values.presto.spec.presto.config.nodeSchedulerIncludeCoordinator }}
+    discovery-server.enabled=true
+
+    {{- if .Values.presto.spec.presto.config.useTLS }}
```
chancez: Indent is off

charts/openshift-metering/templates/presto/presto-worker-config.yaml
```diff
@@ -13,12 +13,29 @@ data:
...
+    http-server.http.enabled=false
+    http-server.https.enabled=true
+    http-server.https.port=8080
+    http-server.https.keystore.path=/opt/presto/tls/keystore.jks
+    http-server.https.keystore.key=password
+    node.internal-address=presto
```
chancez: This isn't correct, each worker will have a different address, this is the service name.

chancez: Unfortunately, A records are not created for pod hostnames though, so we may need to create a headless service to make this work,

chancez: https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#pod-s-hostname-and-subdomain-fields has details on this. I think we want to set subdomain + create headless service so we can get an A record per presto pod. Then we can set `node.internal-address=FQDN`

charts/openshift-metering/templates/presto/presto-worker-config.yaml
timflannagan1: It also references an invalid field. I updated both in the most recent commit.

charts/openshift-metering/templates/presto/presto-coordinator-config.yaml
```diff
@@ -13,13 +13,29 @@ data:
...
+    node-scheduler.include-coordinator={{ .Values.presto.spec.presto.config.nodeSchedulerIncludeCoordinator }}
     discovery-server.enabled=true
+
+{{- if .Values.presto.spec.presto.config.useTLS }}
+    http-server.http.enabled=false
+    http-server.https.enabled=true
+    http-server.https.port=8080
+    http-server.https.keystore.path=/opt/presto/tls/keystore.jks
+    http-server.https.keystore.key=password
+    node.internal-address=presto
```
chancez: I think we want to do this the same way that we do the workers, using the podIP. It shouldn't need to match the discovery URI.

chancez: Update: we've decided to go with hostnames using statefulsets. So this can use `FQDN` instead.

charts/openshift-metering/templates/presto/presto-worker-config.yaml
chancez: Update: we've decided to go with hostnames using statefulsets. So this can use `FQDN` instead.

timflannagan1:
I added a commit to rename the Presto StatefulSet files after we merged the commit that changed them from a Deployment to a StatefulSet, and updated the `images/helm-operator/deploy-resources.sh` file to reflect that change. 

I removed the WIP tag, as the tests had previously passed, and this PR feels like it's in a good place now.

charts/openshift-metering/templates/presto/presto-worker-config.yaml
```diff
@@ -13,12 +13,28 @@ data:
...
+    http-server.https.keystore.path=/opt/presto/tls/keystore.jks
+    http-server.https.keystore.key=password
+    node.internal-address-source=FQDN
+    discovery.uri=https://presto:8080
+    internal-communication.https.required=true
+    internal-communication.https.keystore.path=/opt/presto/tls/keystore.jks
+    internal-communication.https.keystore.key=password
+{{- else }}
     http-server.http.port=8080
+    discovery.uri={{ .Values.presto.spec.presto.config.discoveryURI }}
```
chancez: Honestly this could probably be hardcoded and removed from `values.yaml`

charts/openshift-metering/templates/presto/presto-coordinator-config.yaml
```diff
@@ -13,13 +13,29 @@ data:
     node.environment=production
 
   config.properties: |
-    http-server.http.port=8080
-    jmx.rmiserver.port=8081
-    jmx.rmiregistry.port=8081
     coordinator=true
+    node-scheduler.include-coordinator={{ .Values.presto.spec.presto.config.nodeSchedulerIncludeCoordinator }}
```
chancez: Honestly this could probably be hardcoded and removed from `values.yaml`

charts/openshift-metering/templates/presto/presto-coordinator-statefulset.yaml
```diff
@@ -52,10 +52,21 @@ spec:
...
+{{- if .Values.presto.spec.presto.config.tls.enabled }}
+        - name: PRESTO_KEYFILE
+          value: /presto/tls.key
+        - name: PRESTO_CRTFILE
+          value: /presto/tls.crt
+        - name: PRESTO_KEYSTORE_PKCS12
+          value: /opt/presto/tls/keystore.pkcs12
+        - name: PRESTO_KEYSTORE_JKS
+          value: /opt/presto/tls/keystore.jks
+        - name: PASSWORD
```
chancez: `PRESTO_KEYSTORE_PASSWORD`

chancez:
This looks great. Can you write some brief instructions so I could test it out locally? Basically just the commands you would use to create the certs, secrets, and then what your custom Metering CR looks like.

charts/openshift-metering/values.yaml
```diff
@@ -418,14 +422,15 @@ presto:
...
+          enabled: false
+          secretName: ""
         connectors:
           extraConnectorFiles: []
-        discoveryURI: http://presto:8080
         environment: production
         hiveMetastoreURI: thrift://hive-metastore:9083
         maxQueryLength: "10000000"
         metastoreTimeout: null
-        nodeSchedulerIncludeCoordinator: true
```
chancez: We still want this

charts/openshift-metering/templates/presto/presto-coordinator-config.yaml
chancez: Shoot my bad. I meant to put this on `discovery.uri`, sorry.

timflannagan1: Updated in the last force-push

charts/openshift-metering/templates/reporting-operator/reporting-operator-config.yaml
```diff
@@ -15,6 +15,9 @@ data:
   prometheus-url: {{ required "a valid reporting-operator.spec.config.prometheusURL must be set" $operatorValues.spec.config.prometheusURL | quote}}
   disable-prometheus-metrics-importer: {{ $operatorValues.spec.config.disablePrometheusMetricsImporter | quote}}
   enable-finalizers: {{ $operatorValues.spec.config.enableFinalizers | quote}}
+{{- if $operatorValues.spec.config.prestoTLS.enabled }}
+  presto-ca-file: "{{ $operatorValues.spec.config.prestoTLS.mountLocation }}/certificate.crt"
```
chancez: I feel like we should use ca.crt or tls.crt here for consistency. How about we just use `tls.crt` for now. 

Also, I'm not entirely sure why we have `$operatorValues.spec.config.prestoTLS.mountLocation` at all? I don't think any of this needs to be configurable. Additionally, maybe don't use `/var/run/secrets/java.io/keystores`, that's a bit confusing since we're not using a java keystore file in this secret, just a CA certificate and not a java program. Maybe use `/var/run/secrets/presto-tls/tls.crt`

charts/openshift-metering/templates/reporting-operator/reporting-operator-deployment.yaml
```diff
@@ -257,6 +269,10 @@ spec:
 {{ toYaml $operatorValues.spec.livenessProbe | indent 10 }}
 {{- end }}
         volumeMounts:
+{{- if $operatorValues.spec.config.prestoTLS.enabled }}
+        - name: presto-keystore-volume
```
chancez: `name: presto-tls` is probably more concise

charts/openshift-metering/templates/presto/presto-worker-statefulset.yaml
```diff
@@ -96,6 +115,10 @@ spec:
           mountPath: /var/presto/data
         - name: presto-logs
           mountPath: /var/presto/logs
+{{- if .Values.presto.spec.presto.config.tls.enabled }}
+        - name: presto-tls-volume
```
chancez: I don't think you need to put `-volume` on the end here, it's kinda implied.

chancez:
@timflannagan1  Can you squash this up a bit?

timflannagan1:
@chancez rebased

PR #732, opened 2019-05-29
charts/openshift-metering: Update the Presto components to use a StatefulSet instead of a Deployment.

timflannagan1:


chancez:
😂 we forgot to rename the files.

timflannagan1:
yeah, I removed that commit (renaming the presto deployment files) after the hack scripts were failing. I'll have to include that in the WIP Presto w/ TLS PR.

PR #739, opened 2019-06-10
Add support for configuring TLS authentication for Presto components by providing pre-existing secrets containing server, client, and CA resources.

timflannagan1:
There are a couple more things that need to be fleshed out - most notably the ordering of the TLS/auth (`.Values.presto.spec.presto.config.*.enabled` in `values.yaml`) and what we want the default behavior to be when TLS is enabled and auth is not, and vice-versa. 

The current state of this PR is was constructed under the assumption that we want to be able to configure authentication only if TLS is also enabled (which is reflected as a nested helm if-conditional  templating). It's also important to note that we recently upgraded our Presto fork to a much newer version. This version allows us to specify PEM formatted certificates/keys to Presto, where before we needed to provide JKS/PKCS files (and had to convert the PEM files in the pre-existing secrets, to JKS keystores/truststores.) 

See: https://gist.github.com/timflannagan1/5b9cd5425eb5c4b85c804f034f7da9ea for help setting up locally.

### DIFF
--- a/charts/openshift-metering/templates/presto/presto-common-config.yaml
+++ b/charts/openshift-metering/templates/presto/presto-common-config.yaml
@@ -8,8 +8,14 @@ data:
     set -e
 
 {{- if .Values.presto.spec.presto.config.tls.enabled }}
-    openssl pkcs12 -export -inkey $PRESTO_KEYFILE -in $PRESTO_CRTFILE -out $PRESTO_KEYSTORE_PKCS12 -password pass:$PRESTO_KEYSTORE_PASSWORD
-    keytool -importkeystore -noprompt -srckeystore $PRESTO_KEYSTORE_PKCS12 -srcstoretype pkcs12 -destkeystore $PRESTO_KEYSTORE_JKS -storepass $PRESTO_KEYSTORE_PASSWORD -srcstorepass $PRESTO_KEYSTORE_PASSWORD
+    # copy all necessary tls.crt/tls.key from read-only secret volumes
+    cat $PRESTO_SERVER_KEYFILE $PRESTO_SERVER_CRTFILE > $PRESTO_SERVER_PEM
+    cat $PRESTO_SERVER_PEM $PRESTO_CA_CRTFILE > $PRESTO_KEYSTORE_PEM
+    cat $PRESTO_SERVER_PEM $PRESTO_CA_CRTFILE > $PRESTO_TRUSTSTORE_PEM
+{{- end }}
+{{- if .Values.presto.spec.presto.config.auth.enabled }}
+    # append the client key/certificate to the beginning of the truststore PEM
+    sed -i -e '1 e cat $PRESTO_CLIENT_KEYFILE $PRESTO_CLIENT_CRTFILE' $PRESTO_TRUSTSTORE_PEM
 {{- end }}
 
     cp -v -L -r -f /presto-etc/* /opt/presto/presto-server/etc/
--- a/charts/openshift-metering/templates/presto/presto-common-config.yaml
+++ b/charts/openshift-metering/templates/presto/presto-common-config.yaml
@@ -7,15 +7,14 @@ data:
     #!/bin/bash
     set -e
 
-{{- if .Values.presto.spec.presto.config.tls.enabled }}
-    # copy all necessary tls.crt/tls.key from read-only secret volumes
-    cat $PRESTO_SERVER_KEYFILE $PRESTO_SERVER_CRTFILE > $PRESTO_SERVER_PEM
-    cat $PRESTO_SERVER_PEM $PRESTO_CA_CRTFILE > $PRESTO_KEYSTORE_PEM
-    cat $PRESTO_SERVER_PEM $PRESTO_CA_CRTFILE > $PRESTO_TRUSTSTORE_PEM
-{{- end }}
 {{- if .Values.presto.spec.presto.config.auth.enabled }}
-    # append the client key/certificate to the beginning of the truststore PEM
-    sed -i -e '1 e cat $PRESTO_CLIENT_KEYFILE $PRESTO_CLIENT_CRTFILE' $PRESTO_TRUSTSTORE_PEM
+    cat $PRESTO_CLIENT_KEYFILE $PRESTO_CLIENT_CRTFILE > $PRESTO_TRUSTSTORE_PEM
+
+{{- if .Values.presto.spec.presto.config.tls.enabled }}
+    # copy all necessary tls.crt/tls.key from read-only secret volumes - append server bundle and CA cert to truststore
+    cat $PRESTO_SERVER_KEYFILE $PRESTO_SERVER_CRTFILE $PRESTO_CA_CRTFILE > $PRESTO_KEYSTORE_PEM
+    cat $PRESTO_SERVER_KEYFILE $PRESTO_SERVER_CRTFILE $PRESTO_CA_CRTFILE >> $PRESTO_TRUSTSTORE_PEM
+{{- end }}
 {{- end }}
 
     cp -v -L -r -f /presto-etc/* /opt/presto/presto-server/etc/
--- a/charts/openshift-metering/templates/reporting-operator/reporting-operator-config.yaml
+++ b/charts/openshift-metering/templates/reporting-operator/reporting-operator-config.yaml
@@ -18,7 +18,7 @@ data:
 {{- if $operatorValues.spec.config.prestoTLS.enabled }}
   presto-ca-file: "/var/run/secrets/presto-tls/tls.crt"
 {{- if $operatorValues.spec.config.prestoAuth.enabled }}
-  presto-server-file: "/var/run/secrets/presto-auth/"
+  presto-client-cert-file: "/var/run/secrets/presto-auth/"
 {{- end }}
 {{- end }}
 {{- if $operatorValues.spec.config.prometheusMetricsImporterPollInterval }}
--- a/charts/openshift-metering/templates/reporting-operator/reporting-operator-config.yaml
+++ b/charts/openshift-metering/templates/reporting-operator/reporting-operator-config.yaml
@@ -18,7 +18,8 @@ data:
 {{- if $operatorValues.spec.config.prestoTLS.enabled }}
   presto-ca-file: "/var/run/secrets/presto-tls/tls.crt"
 {{- if $operatorValues.spec.config.prestoAuth.enabled }}
-  presto-client-cert-file: "/var/run/secrets/presto-auth/"
+  presto-client-cert-file: "/var/run/secrets/presto-auth/tls.crt"
+  presto-client-key-file: "/var/run/secrets/presto-auth/tls.key"
 {{- end }}
 {{- end }}
 {{- if $operatorValues.spec.config.prometheusMetricsImporterPollInterval }}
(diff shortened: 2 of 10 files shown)


charts/openshift-metering/templates/presto/presto-common-config.yaml
```diff
@@ -8,8 +8,14 @@ data:
...
-    openssl pkcs12 -export -inkey $PRESTO_KEYFILE -in $PRESTO_CRTFILE -out $PRESTO_KEYSTORE_PKCS12 -password pass:$PRESTO_KEYSTORE_PASSWORD
-    keytool -importkeystore -noprompt -srckeystore $PRESTO_KEYSTORE_PKCS12 -srcstoretype pkcs12 -destkeystore $PRESTO_KEYSTORE_JKS -storepass $PRESTO_KEYSTORE_PASSWORD -srcstorepass $PRESTO_KEYSTORE_PASSWORD
+    # copy all necessary tls.crt/tls.key from read-only secret volumes
+    cat $PRESTO_SERVER_KEYFILE $PRESTO_SERVER_CRTFILE > $PRESTO_SERVER_PEM
+    cat $PRESTO_SERVER_PEM $PRESTO_CA_CRTFILE > $PRESTO_KEYSTORE_PEM
+    cat $PRESTO_SERVER_PEM $PRESTO_CA_CRTFILE > $PRESTO_TRUSTSTORE_PEM
+{{- end }}
+{{- if .Values.presto.spec.presto.config.auth.enabled }}
+    # append the client key/certificate to the beginning of the truststore PEM
+    sed -i -e '1 e cat $PRESTO_CLIENT_KEYFILE $PRESTO_CLIENT_CRTFILE' $PRESTO_TRUSTSTORE_PEM
```
chancez: I don't understand the sed expression here. Wan't we just concat things in the correct order?

cmd/reporting-operator/start.go
```diff
@@ -79,7 +79,9 @@ func init() {
 
 	startCmd.Flags().StringVar(&cfg.PrestoHost, "presto-host", defaultPrestoHost, "the hostname:port for connecting to Presto")
 	startCmd.Flags().BoolVar(&cfg.PrestoUseTLS, "presto-use-tls", false, "If true, enables TLS when connecting to Presto")
+	startCmd.Flags().BoolVar(&cfg.PrestoUseAuth, "presto-use-auth", false, "If true, enables TLS authentication when connecting to Presto")
 	startCmd.Flags().StringVar(&cfg.PrestoCAFile, "presto-ca-file", "", "The path to the certificate authority to use to connect to Presto.")
+	startCmd.Flags().StringVar(&cfg.PrestoServerFile, "presto-server-file", "", "The path to the certificate authority to use to connect to Presto.")
```
chancez: `presto-client-cert-file` is probably more accurate

cmd/reporting-operator/start.go
```diff
@@ -79,7 +79,9 @@ func init() {
 
 	startCmd.Flags().StringVar(&cfg.PrestoHost, "presto-host", defaultPrestoHost, "the hostname:port for connecting to Presto")
 	startCmd.Flags().BoolVar(&cfg.PrestoUseTLS, "presto-use-tls", false, "If true, enables TLS when connecting to Presto")
+	startCmd.Flags().BoolVar(&cfg.PrestoUseAuth, "presto-use-auth", false, "If true, enables TLS authentication when connecting to Presto")
```
chancez: in the description: `Enables TLS Client Certificate Authentication`

cmd/reporting-operator/start.go
chancez: Description could also be updated to reflect it's a client certificate, not a CA

charts/openshift-metering/values.yaml
```diff
@@ -423,7 +426,11 @@ presto:
       config:
         tls:
           enabled: false
-          secretName: ""
+          serverSecretName: ""
+          rootCASecretName: ""
+        auth:
+          enabled: false
+          clientSecretName: ""
```
chancez: `clientCertificateSecretName`?

charts/openshift-metering/templates/reporting-operator/reporting-operator-deployment.yaml
```diff
@@ -197,6 +197,18 @@ spec:
...
           value: "false"
+{{- end }}
+{{- if $operatorValues.spec.config.prestoAuth.enabled }}
+        - name: REPORTING_OPERATOR_PRESTO_USE_AUTH
+          value: "true"
+        - name: REPORTING_OPERATOR_PRESTO_SERVER_FILE
+          valueFrom:
+            configMapKeyRef:
+              name: reporting-operator-config
+              key: presto-server-file
```
chancez: presto-client-cert-file

charts/openshift-metering/templates/reporting-operator/reporting-operator-deployment.yaml
```diff
@@ -197,6 +197,18 @@ spec:
 {{- else }}
         - name: REPORTING_OPERATOR_PRESTO_USE_TLS
           value: "false"
+{{- end }}
+{{- if $operatorValues.spec.config.prestoAuth.enabled }}
+        - name: REPORTING_OPERATOR_PRESTO_USE_AUTH
+          value: "true"
+        - name: REPORTING_OPERATOR_PRESTO_SERVER_FILE
```
chancez: `REPORTING_OPERATOR_PRESTO_CLIENT_CERT_FILE`

charts/openshift-metering/templates/reporting-operator/reporting-operator-config.yaml
```diff
@@ -17,6 +17,9 @@ data:
   enable-finalizers: {{ $operatorValues.spec.config.enableFinalizers | quote}}
 {{- if $operatorValues.spec.config.prestoTLS.enabled }}
   presto-ca-file: "/var/run/secrets/presto-tls/tls.crt"
+{{- if $operatorValues.spec.config.prestoAuth.enabled }}
+  presto-server-file: "/var/run/secrets/presto-auth/"
```
chancez: This is a directory, but the reporting operator seems to be expecting a file at this location?

charts/openshift-metering/values.yaml
```diff
@@ -423,7 +426,11 @@ presto:
       config:
         tls:
           enabled: false
-          secretName: ""
+          serverSecretName: ""
```
chancez: `serverCertificateSecretName`?

charts/openshift-metering/values.yaml
```diff
@@ -423,7 +426,11 @@ presto:
...
+          rootCASecretName: ""
```
chancez: Was there a reason this couldn't go into the same secret as the `serverCertificateSecretName`?

pkg/operator/operator.go
```diff
@@ -105,9 +105,11 @@ type Config struct {
...
 	HiveCAFile string
 
-	PrestoHost   string
-	PrestoUseTLS bool
-	PrestoCAFile string
+	PrestoHost       string
+	PrestoUseTLS     bool
+	PrestoUseAuth    bool
+	PrestoCAFile     string
+	PrestoServerFile string
```
chancez: `PrestoClientCertFile`

charts/openshift-metering/templates/reporting-operator/reporting-operator-config.yaml
timflannagan1: The current implementation expects a directory, see `operator.go`:
```
serverCert, err := tls.LoadX509KeyPair(op.cfg.PrestoServerFile+"tls.crt", op.cfg.PrestoServerFile+"tls.key")
```

This could be removed and we could instead use separate variables for the cert/key location

pkg/operator/operator.go
```diff
@@ -105,9 +105,11 @@ type Config struct {
 	HiveUseTLS bool
 	HiveCAFile string
 
-	PrestoHost   string
-	PrestoUseTLS bool
-	PrestoCAFile string
+	PrestoHost       string
+	PrestoUseTLS     bool
+	PrestoUseAuth    bool
```
chancez: Perhaps `PrestoUseClientCertAuth` since there's more than 1 form of auth, and we're using client certs specifically.

pkg/operator/operator.go
```diff
@@ -445,19 +447,39 @@ func (op *Reporting) Run(ctx context.Context) error {
...
+			RootCAs: rootCertPool,
+		}
+
+		if op.cfg.PrestoUseAuth {
+			serverCert, err := tls.LoadX509KeyPair(op.cfg.PrestoServerFile+"tls.crt", op.cfg.PrestoServerFile+"tls.key")
+			if err != nil {
+				return fmt.Errorf("presto: Error loading SSL Server cert/key file: %v", err)
+			}
+			// override the default customClientConfig with server/client certificates
+			customClientConfig = &tls.Config{
```
chancez: Instead of overwriting the existing variable, why not just assign to the fields that need to be set?

pkg/operator/operator.go
```diff
@@ -445,19 +447,39 @@ func (op *Reporting) Run(ctx context.Context) error {
...
 		}
-		certPool := x509.NewCertPool()
-		certPool.AppendCertsFromPEM(cert)
 
-		// build up the http client structure to use the provided CA cert
+		rootCertPool := x509.NewCertPool()
+		rootCertPool.AppendCertsFromPEM(rootCert)
+
+		// if we are using only TLS and auth is disabled, then we only need the root CA cert pool
+		customClientConfig = &tls.Config{
```
chancez: A better name to call this might be `tlsConfig` or `prestoTLSConfig`.

cmd/reporting-operator/start.go
chancez: Also indicate `presto-use-tls` must be true also.

charts/openshift-metering/templates/presto/presto-common-config.yaml
timflannagan1: The sed expression above appends the values stored in `$PRESTO_CLIENT_*` to the beginning of the truststore key in order to maintain the certificate chain. I just used this for its brevity over the usage of `cat to re-construct the truststore, but I can change this to using the latter if you prefer the readability.

charts/openshift-metering/templates/reporting-operator/reporting-operator-config.yaml
chancez: Yes please do that.

pkg/operator/operator.go
```diff
@@ -445,19 +448,36 @@ func (op *Reporting) Run(ctx context.Context) error {
 
 	// check if the PrestoUseTLS flag is set to true
 	if op.cfg.PrestoUseTLS {
-		cert, err := ioutil.ReadFile(op.cfg.PrestoCAFile)
+		prestoTLSConfig := &tls.Config{}
```
chancez: Why not just initialize this at the same time as you set the RootCAs? Eg remove this line, change https://github.com/operator-framework/operator-metering/pull/739/files#diff-554de5523753fda8c93e7008c9bd947fR461 to use `:=`

timflannagan1: yeah, my mistake I should've changed that after making the previous changes.

PR #740, opened 2019-06-11
charts,hack: Set presto resource requests lower and remove limits for CI tests

chancez:
I noticed CI fail once because presto coordinator was stuck pending due to not enough CPU resources. removing limits and lowering request may potentially help.

### DIFF
--- a/charts/metering-ci/values.yaml
+++ b/charts/metering-ci/values.yaml
@@ -14,4 +14,6 @@ hdfsNamenodeMemory: null
 hiveMetastoreStorageSize: null
 hiveMetastoreMemory: null
 hiveMetastoreCpu: null
+prestoMemory: null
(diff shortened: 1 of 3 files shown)


charts/metering-ci/values.yaml
```diff
@@ -14,4 +14,6 @@ hdfsNamenodeMemory: null
 hiveMetastoreStorageSize: null
 hiveMetastoreMemory: null
 hiveMetastoreCpu: null
+prestoMemory: null
+prestoCpu: null
```
timflannagan1: would `prestoCPU` read better?

chancez: Mostly going for consistency with the other fields. Not really worried about that since it's only used by CI.

PR #753, opened 2019-06-14
Update references to hdfs configuration in MeteringConfig

chancez:


Documentation/configuring-storage.md
```diff
@@ -50,9 +50,9 @@ Uncomment the following sections and replace the `null` in `class: null` value w
 
 Use [custom-storage.yaml][custom-storage-config] as a template and adjust the `size: "5Gi"` value to the desired capacity for the following sections:
 
-- `presto.spec.hive.metastore.storage.size`
-- `hdfs.spec.datanode.storage.size`
-- `hdfs.spec.namenode.storage.size`
+- `spec.presto.spec.hive.metastore.storage.size`
+- `spec.hadoop.spec.hdfs.datanode.storage.size`
```
timflannagan1: shouldn't this, and the others be `hadoop.spec.hdfs...`?

timflannagan1: or is the first `spec` referencing the `spec` field in the metering config?

chancez: These ones are actually correct, but the ones in the section above are wrong. Fixed.

chancez: @timflannagan1 the first `spec` is the meteringconfig `spec`. We were a bit inconsistent on that previously.

PR #757, opened 2019-06-17
pkg/operator: Never stop re queuing StorageLocations

chancez:
I noticed sometimes with the new setup that the reporting-operator stops processing storageLocations because hive-server/hive-metastore isn't reachable when reporting-operator starts, and it errors enough times to stop processing storageLocations. This fixes that by requeuing storageLocation resources forever. These are fairly important and if they don't get processed the reporting-operator is effectively stuck until restarted.

### DIFF
--- a/pkg/operator/queues.go
+++ b/pkg/operator/queues.go
@@ -446,7 +446,7 @@ func (op *Reporting) handleErr(logger log.FieldLogger, err error, objType string
 
 	// This controller retries up to maxRequeues times if something goes wrong.
 	// After that, it stops trying.
-	if queue.NumRequeues(obj) < maxRequeues {
+	if maxRequeues == -1 || queue.NumRequeues(obj) < maxRequeues {
 		logger.WithError(err).Errorf("error syncing %s %q, adding back to queue", objType, obj)
 		queue.AddRateLimited(obj)
 		return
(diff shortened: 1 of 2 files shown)


pkg/operator/queues.go
```diff
@@ -446,7 +446,7 @@ func (op *Reporting) handleErr(logger log.FieldLogger, err error, objType string
 
 	// This controller retries up to maxRequeues times if something goes wrong.
 	// After that, it stops trying.
-	if queue.NumRequeues(obj) < maxRequeues {
+	if maxRequeues == -1 || queue.NumRequeues(obj) < maxRequeues {
```
timflannagan1: Is there a point to the right side of this conditional? Seems like it's always going to evaluate to true.

pkg/operator/storagelocations.go
```diff
@@ -22,28 +22,11 @@ const (
 func (op *Reporting) runStorageLocationWorker() {
 	logger := op.logger.WithField("component", "storageLocationWorker")
 	logger.Infof("StorageLocation worker started")
-	const maxRequeues = 10
+	const maxRequeues = -1
 	for op.processResource(logger, op.syncStorageLocation, "StorageLocation", op.storageLocationQueue, maxRequeues) {
 	}
 }
 
-func (op *Reporting) processStorageLocation(logger log.FieldLogger) bool {
```
chancez: This simply was dead/unused code I noticed while updating the above.

pkg/operator/queues.go
chancez: Yep, maxRequeues is normally something like 5 or 10, but in this case we're setting it to -1.

PR #760, opened 2019-06-17
Disable HDFS by default and wait for user to configure storage

chancez:
- Add new top-level storage config under spec.storage to centralize
  storage related configuration.
- openshift-reporting: move storageLocation into
  openshift-reporting.spec and rename to defaultStorageLocation. This
  move is to aligns with the new top-level storage option so that there
  isn't confusion between the two.
- hive: Don't create shared-pvc by default and stop creating empty
  warehouse volume. This has been removed from the hive image, and only
  makes it possible to start hive without storage which isn't wanted.
- *: Don't create awsCredentialsSecrets by default and don't make
  optional env var. Ensures that when the secret is missing that pods do
  not start.
- Rename hdfs-config to hadoop-config to generalize it more.
- Update installation documentation and configuration documentation to
  explicitly mention that storage is no longer configured by default,
  and that HDFS is an unsupported, developer only storage option.

Currently CI is still configured to use HDFS since we have no easily available alternative.

### DIFF
--- a/Documentation/configuring-hdfs.md
+++ b/Documentation/configuring-hdfs.md
@@ -1,0 +1,33 @@
+# Configuring HDFS
+
+By default, HDFS is not used or installed with Metering.
+However, if you set the `spec.unsupportedFeatures.enableHDFS` unsupported features toggle to true, then you can install a non-production grade HDFS cluster alongside metering for storage.
+
+We are not HDFS experts, and thus we will not directly support HDFS, and it should only be used for testing or development.
+Currently there is no way to secure the communications with HFDS, meaning all communications are done in plain text, without authentication.
+
+Proceed at your own risk.
+
+## Persistent Volumes created by HDFS
+
+- `hdfs-namenode-data-hdfs-namenode-0`
+  Used by the hdfs-namenode pod to store metadata about the files and blocks stored in the hdfs-datanodes.
+- One `hdfs-datanode-data-hdfs-datanode-$i` PV per hdfs-datanode replica.
+  Used by each hdfs-datanode pod to store blocks for files in the HDFS cluster.
+
+## Configuring the Storage Class for HDFS
+
+Use [hdfs-storage.yaml][hdfs-storage-config] as a template and adjust the `class: null` value to name of the `StorageClass` to use.
+Leaving the value `null` will cause Metering to use the default StorageClass for the cluster.
+
+- `spec.hadoop.spec.hdfs.datanode.storage.class`
+- `spec.hadoop.spec.hdfs.namenode.storage.class`
+
+## Configuring the volume sizes for HDFS
+
+Use [hdfs-storage.yaml][hdfs-storage-config] as a template and adjust the `size: "5Gi"` value to the desired capacity for the following sections:
+
+- `spec.hadoop.spec.hdfs.datanode.storage.size`
+- `spec.hadoop.spec.hdfs.namenode.storage.size`
+
+[hdfs-storage-config]: ../manifests/metering-config/hdfs-storage.yaml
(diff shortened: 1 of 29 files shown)


Documentation/configuring-hdfs.md
```diff
@@ -0,0 +1,33 @@
+# Configuring HDFS
+
+By default, HDFS is not used or installed with Metering.
+However, if you set the `spec.unsupportedFeatures.enableHDFS` unsupported features toggle to true, then you can install a non-production grade HDFS cluster along side metering for storage.
```
timflannagan1: nit: I believe alongside is the correct spelling.

charts/openshift-metering/templates/reporting-operator/reporting-operator-aws-credentials-secrets.yaml
```diff
@@ -3,7 +3,7 @@
 apiVersion: v1
 kind: Secret
 metadata:
-  name: reporting-operator-aws-credentials-secrets
+  name: reporting-operator-aws-credentials
```
timflannagan1: I like this better - I'll have to update the Presto and reporting-operator (api, metrics, presto) to remove the `-secrets` suffix from the secret template name.

chancez: Yeah, I basically changed it so the name of the secret created isn't controllable by the user. If you look at the deployment you'll see how I use `default` to go to use this secret name if the user didn't specify an `awsCredentialSecretName`.

Documentation/configuring-hive-metastore.md
```diff
@@ -1,11 +1,38 @@
...
-Generally the default configuration of Hive metastore works for small clusters, but users may wish to improve performance or move storage requirements out of cluster by using a dedicated database for storing table metadata for Presto and Hive server.
+Hive metastore is responsible for storing all the metadata about the database tables we create in Presto and Hive.
+By default, the metastore stores this information in a local embedded Derby database in a PersistentVolume attached to the pod.
 
-## Use MySQL or Postgresql for the Hive Metastore database
+Generally the default configuration of Hive metastore works for small clusters, but users may wish to improve performance or move storage requirements out of cluster by using a dedicated SQL database for storing the Hive metastore data.
+
+## Configuring PersistentVolumes
+
+Hive by default requires 1 Persistent Volume to operate.
```
timflannagan1: 1 -> one

timflannagan1: Hive, by default, requires one Persistent Volume to operate.

Documentation/configuring-hive-metastore.md
```diff
@@ -1,11 +1,38 @@
...
+
+`hive-metastore-db-data` is the main PVC required by default.
+It is used by Hive metastore to retain information about the location of where data is stored, which Presto and Hive server use.
```
timflannagan1: Maybe change this to "This is used..." or "This PVC is used..."

Documentation/configuring-hive-metastore.md
```diff
@@ -1,11 +1,38 @@
...
+In practice, it is possible to remove this requirement by using [MySQL](#use-mysql-for-the-hive-metastore-database) or [PostgreSQL](#use-postgresql-for-the-hive-metastore-database) for the Hive metastore database.
+
+To install Hive metastore requires that dynamic volume provisioning be enabled via a Storage Class, a persistent volumes of the correct size must be manually pre-created, or that you use a pre-existing MySQL or PostgreSQL database.
```
timflannagan1: "To install, Hive metastore ..."

chancez: Isn't it actually "To install Hive metastore, requires..."?

chancez: Actually, I could see as yours, and as it is currently, my previous suggestion seems wrong though when I say it outloud.

timflannagan1:
I'm going to assume this is still ready as it looks like you just rebased and updated the documentation/resource limits.
/lgtm

PR #761, opened 2019-06-18
Update typo in OpenShift example

rbaumgar:


chancez:
This is already covered by https://github.com/operator-framework/operator-metering/pull/761/files#diff-acc1813502153f043ffc79e97341de19R85

Do you feel that the instructions aren't clear enough? In the future when we have docs added to the upstream openshift-docs repo, they'll be specific to Openshift which will help streamline stuff like this that differs between Openshift and vanilla Kube.

PR #762, opened 2019-06-18
Fix lookup of DataSource table name in Prometheus store/fetch API

chancez:
Also updates e2e/integration tests to ensure the PrestoTable referenced by
ReportDataSources exist and have a status.tableName.


pkg/operator/reportingutil/util.go
```diff
@@ -51,7 +51,7 @@ func FullyQualifiedTableName(prestoTable *metering.PrestoTable) (string, error)
 		return "", fmt.Errorf("PrestoTable status is invalid: %s", strings.Join(errs, ", "))
 	}
 
-	return presto.FullyQuaifiedTableName(prestoTable.Status.Catalog, prestoTable.Status.Schema, prestoTable.Status.TableName), nil
```
timflannagan1: wow I forgot this was still here

PR #765, opened 2019-06-18
charts,hack: Bump pod resources limits

chancez:
Keep lower resource limits in CI, but bump limits to reduce amount of CPU cgroup
throttling. Overall it seems most pods are being throttled at some
level, meaning our CPU resources are too low. To start, try to increase
the CPU limits and keep CPU request low for scheduleability.

Additionally, increase memory limits of Java based components to
increase the JVM heap size which should also reduce CPU usage by having
to do GCs less frequently.


charts/metering-ci/templates/metering.yaml
```diff
@@ -63,6 +63,13 @@ spec:
         # or less
         prometheusMetricsImporterStepSize: "60s"
 
+      {{- if or .Values.reportingOperatorMemory .Values.reportingOperatorCpu }}
```
timflannagan1: Does it matter where this line is aligned? Seems like the CI metering file is pretty inconsistent in that regard.

timflannagan1: At the same time, it's the CI metering file, so it's probably not something that needs to be worried about.

charts/openshift-metering/values.yaml
```diff
@@ -528,8 +528,8 @@ presto:
         nodeSelector: {}
         resources:
           limits:
-            cpu: "2"
-            memory: 2Gi
+            cpu: "4"
```
timflannagan1: I see the majority of the values.yaml doesn't use quotes around the integer values

charts/metering-ci/templates/metering.yaml
chancez: No, because of the `-` in the `{{-`, but I'm going to fix this up. Indenting a bit helps since it's hard to find which if/ends go together.

charts/openshift-metering/values.yaml
chancez: True. Fixed.

PR #768, opened 2019-06-20
charts/openshift-metering: Update the annotations used in the Presto coordinator/worker statefulsets to use the secret template hashes.

timflannagan1:
This was left out after we made the switch from deployments to statefulsets.

### DIFF
--- a/charts/openshift-metering/templates/presto/presto-coordinator-statefulset.yaml
+++ b/charts/openshift-metering/templates/presto/presto-coordinator-statefulset.yaml
@@ -28,6 +28,12 @@ spec:
         presto-common-config-hash: {{ include (print $.Template.BasePath "/presto/presto-common-config.yaml") . | sha256sum }}
         presto-catalog-config-hash: {{ include (print $.Template.BasePath "/presto/presto-catalog-config-secret.yaml") . | sha256sum }}
         presto-jmx-config-hash: {{ include (print $.Template.BasePath "/presto/presto-jmx-config.yaml") . | sha256sum }}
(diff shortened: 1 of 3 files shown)


timflannagan1:
@chancez I was under the impression I had already added those, but it appears that only the `reporting-operator-presto-tls-secrets.yaml` was included in the annotations. Updated.

charts/openshift-metering/templates/presto/presto-coordinator-statefulset.yaml
```diff
@@ -28,6 +28,12 @@ spec:
         presto-common-config-hash: {{ include (print $.Template.BasePath "/presto/presto-common-config.yaml") . | sha256sum }}
         presto-catalog-config-hash: {{ include (print $.Template.BasePath "/presto/presto-catalog-config-secret.yaml") . | sha256sum }}
         presto-jmx-config-hash: {{ include (print $.Template.BasePath "/presto/presto-jmx-config.yaml") . | sha256sum }}
+{{- if .Values.presto.spec.presto.config.tls.enabled }}
```
chancez: This also requires `createSecret` to be true right? Otherwise we're using the users secret and we won't be using this one so hashing it isn't useful.

maybe `if and .Values.presto.spec.presto.config.tls.enabled .Values.presto.spec.presto.config.tls.createSecret`

timflannagan1: Yeah, that needed to be updated. There was also a poor copy-paste job done in the reporting-operator deployment that I updated also.

PR #769, opened 2019-06-21
Enable TLS and client-side authentication in Presto and reporting-operator by default.

timflannagan1:
Currently seeing if the images build correctly.

This needs to add some more validation of relevant fields and clarify some of the behavior we want.

### DIFF
--- a/images/metering-ansible-operator/roles/meteringconfig/defaults/main.yml
+++ b/images/metering-ansible-operator/roles/meteringconfig/defaults/main.yml
@@ -104,7 +104,70 @@ _storage_overrides:
 meteringconfig_storage_hive_storage_type: "{{ meteringconfig_default_values | combine(meteringconfig_spec_overrides, recursive=True) | json_query('storage.hive.type') }}"
 meteringconfig_storage_overrides: "{{ _storage_overrides[meteringconfig_storage_hive_storage_type] | default({}, true) }}"
 
-meteringconfig_spec: "{{ meteringconfig_default_values | combine(meteringconfig_default_image_overrides, meteringconfig_storage_overrides, meteringconfig_spec_overrides, recursive=True) }}"
+#
+# Override the default values with _tls_overrides when spec.tls.enabled is set to true, else use values.yaml and meteringconfig_spec_overrides
+#
+meteringconfig_reporting_operator_presto_server_ca_certificate: ""
+meteringconfig_reporting_operator_presto_server_secret_name: "reporting-operator-presto-server-tls"
+
+meteringconfig_reporting_operator_presto_client_secret_name: "reporting-operator-presto-client-tls"
+meteringconfig_reporting_operator_client_cert: ""
+meteringconfig_reporting_operator_presto_client_ca_certificate: ""
+meteringconfig_reporting_operator_client_key: ""
+
+meteringconfig_presto_server_secret_name: "presto-server-tls"
+meteringconfig_presto_server_ca_certificate: ""
+meteringconfig_presto_server_certificate: ""
+meteringconfig_presto_server_key: ""
+
+meteringconfig_presto_client_secret_name: "presto-client-tls"
+meteringconfig_presto_client_ca_certificate: ""
+meteringconfig_presto_client_certificate: ""
+meteringconfig_presto_client_key: ""
+
+_tls_overrides:
+  reporting-operator:
+    spec:
+      config:
+        prestoTLS:
+          enabled: true
+          createSecret: true
+          secretName: "{{ meteringconfig_reporting_operator_presto_server_secret_name }}"
+          caCertificate: "{{ meteringconfig_reporting_operator_presto_server_ca_certificate }}"
+
+        prestoAuth:
+          enabled: true
+          createSecret: true
+          secretName: "{{ meteringconfig_reporting_operator_presto_client_secret_name }}"
+          certificate: "{{ meteringconfig_reporting_operator_client_cert }}"
+          key: "{{ meteringconfig_reporting_operator_client_key }}"
+          caCertificate: "{{ meteringconfig_reporting_operator_presto_client_ca_certificate }}"
+  presto:
+    spec:
+      presto:
+        config:
+          tls:
+            certificate: "{{  meteringconfig_presto_server_certificate }}"
+            key: "{{ meteringconfig_presto_server_key }}"
+            caCertificate: "{{ meteringconfig_presto_server_ca_certificate }}"
+
+            enabled: true
+            createSecret: true
+            secretName: "{{ meteringconfig_presto_server_secret_name }}"
+          auth:
+            certificate: "{{  meteringconfig_presto_client_certificate }}"
+            key: "{{ meteringconfig_presto_client_key }}"
+            caCertificate: "{{ meteringconfig_presto_client_ca_certificate }}"
+
+            enabled: true
+            createSecret: true
+            secretName: "{{ meteringconfig_presto_client_secret_name }}"
+
+# check if the user's spec contains the top-level tls.enabled field. if true, use the value stored in that field, else default to the value stored in values.yaml
+meteringconfig_tls_enabled: "{{ meteringconfig_default_values | combine(meteringconfig_spec_overrides, recursive=True) | json_query('tls.enabled') }}"
+meteringconfig_tls_overrides: "{{ meteringconfig_tls_enabled | ternary(_tls_overrides, {}) }}"
+# combine the meteringconfig_tls_overrides dictionary last to enforce when spec.tls.enabled is specified and set to true
+meteringconfig_spec: "{{ meteringconfig_default_values | combine(meteringconfig_default_image_overrides, meteringconfig_storage_overrides, meteringconfig_spec_overrides, meteringconfig_tls_overrides, recursive=True) }}"
 
 meteringconfig_storage_s3_create_bucket: "{{ meteringconfig_spec | json_query('storage.hive.s3.createBucket') | default(true, true) }}"
 # Intentionally an invalid default name to ensure that this is overridden, and obvious when it's not.
@@ -137,9 +200,18 @@ meteringconfig_create_hadoop_aws_credentials: "{{ _hadoop_spec.config.aws.create
 meteringconfig_create_hive_metastore_pvc: "{{ _presto_spec.hive.metastore.storage.create | default(true) }}"
 meteringconfig_create_presto_shared_volume_pvc: "{{ _presto_spec.config.sharedVolume.enabled and _presto_spec.config.sharedVolume.createPVC | default(false) }}"
 meteringconfig_create_presto_aws_credentials: "{{ _presto_spec.config.createAwsCredentialsSecret | default(false) }}"
+
+#
+# Enabling Presto TLS/Auth by Default
+#
+meteringconfig_template_presto_tls_secret: "{{ _presto_spec.presto.config.tls.enabled and _presto_spec.presto.config.tls.createSecret and _presto_spec.presto.config.tls.secretName != 'presto-server-tls' }}"
+meteringconfig_template_presto_auth_secret: "{{ _presto_spec.presto.config.auth.enabled and _presto_spec.presto.config.auth.createSecret and _presto_spec.presto.config.auth.secretName != 'presto-client-tls' }}"
 meteringconfig_create_presto_tls_secrets: "{{ _presto_spec.presto.config.tls.enabled and _presto_spec.presto.config.tls.createSecret | default(false) }}"
 meteringconfig_create_presto_auth_secrets: "{{ _presto_spec.presto.config.auth.enabled and _presto_spec.presto.config.auth.createSecret | default(false) }}"
 
+#
+# Reporting Operator
+#
 meteringconfig_create_reporting_operator_auth_proxy_cookie_secret: "{{ _reporting_op_spec.authProxy.enabled and _reporting_op_spec.authProxy.createCookieSecret | default(true) }}"
 meteringconfig_create_reporting_operator_auth_proxy_htpasswd_secret: "{{ _reporting_op_spec.authProxy.enabled and _reporting_op_spec.authProxy.createHtpasswdSecret | default(true) }}"
 meteringconfig_create_reporting_operator_auth_proxy_authenticated_emails_secret: "{{ _reporting_op_spec.authProxy.enabled and _reporting_op_spec.authProxy.createAuthenticatedEmailsSecret | default(true) }}"
--- a/images/metering-ansible-operator/roles/meteringconfig/tasks/configure_presto_tls.yml
+++ b/images/metering-ansible-operator/roles/meteringconfig/tasks/configure_presto_tls.yml
@@ -1,0 +1,84 @@
+---
+
+- name: Validate user's Presto TLS configuration
+  include_tasks: validate_presto_tls.yml
+  when: not meteringconfig_tls_enabled
+
+#
+# Check for Presto TLS and auth secret existence (to avoid re-generating/overwriting the secret data if that secret name already exists)
+#
+- name: Check for the existence of the Presto TLS secret
+  k8s_facts:
+    api_version: v1
+    kind: Secret
+    name: "{{ meteringconfig_spec.presto.spec.presto.config.tls.secretName }}"
+    namespace: "{{ meta.namespace }}"
+  no_log: true
+  register: presto_secret_tls_buf
+  when: meteringconfig_tls_enabled
+
+- name: Check for the existence of the Presto Auth secret
+  k8s_facts:
+    api_version: v1
+    kind: Secret
+    name: "{{ meteringconfig_spec.presto.spec.presto.config.auth.secretName }}"
+    namespace: "{{ meta.namespace }}"
+  no_log: true
+  register: presto_secret_auth_buf
+  when: meteringconfig_tls_enabled
+
+#
+# Generate server and client certificates using the Ansible OpenSSL modules when top-level spec.tls.enabled is set to true
+#
+- name: Configure TLS and client-side authentication for Presto and reporting-operator
+  block:
+  - name: Create temporary directory to store all the necessary certificates/keys
+    tempfile:
+      suffix: certificates
+      state: directory
+    register: certificates_dir
+
+  - name: Generate presto server and client TLS certificates and keys
+    include_tasks: configure_presto_openssl.yml
+
+  - name: Configure Presto to use generated TLS files
+    set_fact:
+      meteringconfig_reporting_operator_presto_server_ca_certificate: "{{ lookup('file', '{{ certificates_dir.path }}/ca.crt') + '\n' }}"
+
+      meteringconfig_presto_server_ca_certificate: "{{ lookup('file', '{{ certificates_dir.path }}/ca.crt') + '\n' }}"
+      meteringconfig_presto_server_certificate: "{{ lookup('file', '{{ certificates_dir.path }}/server.crt') + '\n' }}"
+      meteringconfig_presto_server_key: "{{ lookup('file', '{{ certificates_dir.path }}/server.key') + '\n' }}"
+
+      meteringconfig_reporting_operator_presto_client_ca_certificate: "{{ lookup('file', '{{ certificates_dir.path }}/ca.crt') + '\n' }}"
+      meteringconfig_reporting_operator_client_cert: "{{ lookup('file', '{{ certificates_dir.path }}/client.crt') + '\n' }}"
+      meteringconfig_reporting_operator_client_key: "{{ lookup('file', '{{ certificates_dir.path }}/client.key') + '\n' }}"
+
+      meteringconfig_presto_client_ca_certificate: "{{ lookup('file', '{{ certificates_dir.path }}/ca.crt') + '\n' }}"
+      meteringconfig_presto_client_certificate: "{{ lookup('file', '{{ certificates_dir.path }}/client.crt') + '\n' }}"
+      meteringconfig_presto_client_key: "{{ lookup('file', '{{ certificates_dir.path }}/client.key') + '\n' }}"
+    no_log: true
+  always:
+  - name: Cleanup the temporary directory which held the certificates and keys
+    file:
+      path: "{{ certificates_dir.path }}"
+      state: absent
+  when: meteringconfig_tls_enabled and (presto_secret_tls_buf.resources | length == 0 or presto_secret_auth_buf.resources | length == 0)
+
+#
+# Update the MeteringConfig values to use the pre-existing Presto TLS/auth secret data (when the secret already exists)
+#
+- name: Configure Presto to use existing server TLS secret data
+  set_fact:
+    meteringconfig_presto_server_ca_certificate: "{{ presto_secret_tls_buf.resources[0].data['ca.crt'] | b64decode }}"
+    meteringconfig_presto_server_certificate: "{{ presto_secret_tls_buf.resources[0].data['tls.crt'] | b64decode }}"
+    meteringconfig_presto_server_key: "{{ presto_secret_tls_buf.resources[0].data['tls.key'] | b64decode }}"
+  no_log: true
+  when: meteringconfig_tls_enabled and presto_secret_tls_buf.resources | length > 0
+
+- name: Configure Presto to use existing client TLS secret data
+  set_fact:
+    meteringconfig_presto_client_ca_certificate: "{{ presto_secret_auth_buf.resources[0].data['ca.crt'] | b64decode }}"
+    meteringconfig_presto_client_certificate: "{{ presto_secret_auth_buf.resources[0].data['tls.crt'] | b64decode }}"
+    meteringconfig_presto_client_key: "{{ presto_secret_auth_buf.resources[0].data['tls.key'] | b64decode }}"
+  no_log: true
+  when: meteringconfig_tls_enabled and presto_secret_auth_buf.resources | length > 0
--- a/images/metering-ansible-operator/roles/meteringconfig/tasks/configure_reporting_operator_tls.yml
+++ b/images/metering-ansible-operator/roles/meteringconfig/tasks/configure_reporting_operator_tls.yml
@@ -1,0 +1,39 @@
+---
+
+#
+# Presto and reporting-operator TLS
+#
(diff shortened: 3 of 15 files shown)


images/metering-ansible-operator/roles/meteringconfig/defaults/main.yml
```diff
@@ -103,7 +103,68 @@ _storage_overrides:
...
+            certificate: "{{  meteringconfig_presto_client_certificate }}"
+            key: "{{ meteringconfig_presto_client_key }}"
+            caCertificate: "{{ meteringconfig_presto_client_ca_certificate }}"
+
+            enabled: true
+            createSecret: "{{ meteringconfig_presto_auth_create_secret }}"
+            secretName: "{{ meteringconfig_presto_client_secret_name }}"
+
+# what happens if CR is unset
+meteringconfig_tls_overrides: "{{ meteringconfig_default_values.tls.enabled | ternary(_tls_overrides, {}) }}"
```
chancez: I think we should probably also check `meteringconfig_spec_overrides.tls.enabled` probably, otherwise I don't think setting `spec.tls.enabled` in your MeteringConfig will have any effect. It's a bit tricky though, since the user may no have a `spec.tls` section, it's easiest to use `json_query` to get that value, and then use the `default` filter to fallback to the default as defined in `values.yaml`.

Eg:

```suggestion
meteringconfig_tls_enabled: "{{ meteringconfig_spec_overrides | json_query('tls.enabled') | default(meteringconfig_default_values.tls.enabled, true) }}"
meteringconfig_tls_overrides: "{{ meteringconfig_tls_enabled | ternary(_tls_overrides, {}) }}"
```

images/metering-ansible-operator/roles/meteringconfig/tasks/reconcile.yml
```diff
@@ -12,6 +12,78 @@
     hiveStorageType: "{{ meteringconfig_spec_overrides | json_query('storage.hive.type') }}"
   when: hiveStorageType is undefined or hiveStorageType not in ['s3', 'sharedPVC', 'hdfs']
 
+- name: Check if we need to create a Presto TLS secret using the user-populated fields
+  set_fact:
+    meteringconfig_reporting_operator_server_ca_certificate: "{{ _presto_spec.presto.config.tls.caCertificate }}"
```
chancez: So, none of our roles should need to use the `_` variables, they should generally just all consuming the `meteringconfig_spec` variable, and we should be setting the variables within `meteringconfig_spec` such that it's all we need.

These `_` variables are only really to make things less verbose within the default variables file, not for use outside that file. So in this case, I'd expect you to have:

```suggestion
    meteringconfig_reporting_operator_server_ca_certificate: "{{ meteringconfig_spec.presto.spec.presto.config.tls.caCertificate }}"
```

images/metering-ansible-operator/roles/meteringconfig/defaults/main.yml
```diff
@@ -103,7 +103,68 @@ _storage_overrides:
 
 meteringconfig_storage_overrides: "{{ _storage_overrides[meteringconfig_default_values | combine(meteringconfig_spec_overrides) | json_query('storage.hive.type')] | default({}, true) }}"
 
-meteringconfig_spec: "{{ meteringconfig_default_values | combine(meteringconfig_default_image_overrides, meteringconfig_storage_overrides, meteringconfig_spec_overrides, recursive=True) }}"
+#
+# Enable TLS by default
+#
+meteringconfig_reporting_operator_server_ca_certificate: ""
```
chancez: The name of this variable, and the ones below for reporting_operator probably need the word `presto` in them, because it's confusing since reporting-operator has it's own server CA cert for it's own API.

images/metering-ansible-operator/roles/meteringconfig/tasks/reconcile.yml
```diff
@@ -12,6 +12,78 @@
     hiveStorageType: "{{ meteringconfig_spec_overrides | json_query('storage.hive.type') }}"
   when: hiveStorageType is undefined or hiveStorageType not in ['s3', 'sharedPVC', 'hdfs']
 
+- name: Check if we need to create a Presto TLS secret using the user-populated fields
```
chancez: Everything you've added to this file should go into it's own task file.

chancez: Maybe rename `configure_tls.yml` to `configure_presto_tls.yml` and put these tasks into a `configure_tls.yml` task file.

timflannagan1: Good idea, I'll update the conditional on the `configure_tls.yml` to rely on the top-level `meteringconfig_spec.tls.enabled` flag.

images/metering-ansible-operator/roles/meteringconfig/tasks/reconcile.yml
```diff
@@ -12,6 +12,60 @@
...
+  k8s_facts:
+    api_version: v1
+    kind: Secret
+    name: "{{ meteringconfig_ansible_generate_presto_auth_secret_name }}"
+    namespace: "{{ meta.namespace }}"
+  register: presto_secret_client_buf
+  when: meteringconfig_template_presto_auth_secret or meteringconfig_ansible_generate_presto_auth_secret
+
+- name: Configure Presto TLS
+  block:
```
chancez: This block doesn't seem necessary.

images/metering-ansible-operator/roles/meteringconfig/tasks/reconcile.yml
```diff
@@ -12,6 +12,60 @@
     hiveStorageType: "{{ meteringconfig_spec_overrides | json_query('storage.hive.type') }}"
   when: hiveStorageType is undefined or hiveStorageType not in ['s3', 'sharedPVC', 'hdfs']
 
+- name: Template the user-provided certificates and key as needed
+  include_tasks: configure_tls.yml
+  when: meteringconfig_spec.tls.enabled
+
+- name: Check for the existence of the Presto TLS secret
```
chancez: I feel like this logic should happen before we template any openssl content?

timflannagan1:
The image CI test failed due to PR 760 being merged to master, and this branch not being updated.

images/metering-ansible-operator/roles/meteringconfig/tasks/configure_presto_tls.yml
```diff
@@ -0,0 +1,150 @@
...
+      csr_path: "{{ certificates_dir.path }}/client.csr"
+      provider: ownca
+      ownca_path: "{{ certificates_dir.path }}/ca.crt"
+      ownca_privatekey_path: "{{ certificates_dir.path }}/ca.key"
+      selfsigned_digest: sha256
+    when: meteringconfig_spec.presto.spec.presto.config.auth.certificate == ""
+  when: meteringconfig_ansible_generate_presto_auth_secret
+
+- name: Configure TLS for Metering
+  block:
```
chancez: I don't think we need this block

images/metering-ansible-operator/roles/meteringconfig/tasks/configure_presto_tls.yml
```diff
@@ -0,0 +1,150 @@
...
+      meteringconfig_presto_server_key: "{{ lookup('file', '{{ certificates_dir.path }}/server.key') + '\n' }}"
+
+      meteringconfig_create_reporting_operator_presto_tls_secret: true
+      meteringconfig_create_presto_tls_secrets: true
+      meteringconfig_presto_tls_create_secret: true
+  no_log: true
+  when: meteringconfig_ansible_generate_presto_tls_secret
+
+- name: Configure Auth for Metering
+  block:
```
chancez: I don't think we need this block

images/metering-ansible-operator/roles/meteringconfig/tasks/configure_tls.yml
```diff
@@ -0,0 +1,60 @@
...
+- name: Check for the existence of the Presto Auth secret
+  k8s_facts:
+    api_version: v1
+    kind: Secret
+    name: "{{ meteringconfig_ansible_generate_presto_auth_secret_name }}"
+    namespace: "{{ meta.namespace }}"
+  register: presto_secret_client_buf
+  when: meteringconfig_template_presto_auth_secret or meteringconfig_ansible_generate_presto_auth_secret
+
+- name: Manually generate certificates/keys using the openssl modules
```
chancez: Remove "manually"

images/metering-ansible-operator/roles/meteringconfig/tasks/reconcile.yml
```diff
@@ -12,6 +12,10 @@
     hiveStorageType: "{{ meteringconfig_spec_overrides | json_query('storage.hive.type') }}"
   when: hiveStorageType is undefined or hiveStorageType not in ['s3', 'sharedPVC', 'hdfs']
 
+- name: Configure TLS communication between resources
```
chancez: Could probably just say "Configure TLS"

images/metering-ansible-operator/roles/meteringconfig/tasks/configure_tls.yml
```diff
@@ -0,0 +1,60 @@
+---
+
+- name: Check if we need to create a Presto TLS secret using the user-populated fields
```
chancez: The name here doesn't match the logic. It's just doing validation on the Presto Server TLS configuration.

images/metering-ansible-operator/roles/meteringconfig/tasks/configure_tls.yml
```diff
@@ -0,0 +1,60 @@
...
+      msg: "presto.spec.presto.config.tls.caCertificate cannot be empty if createSecret: true and secretName != ''"
+    when: meteringconfig_spec.presto.spec.presto.config.tls.caCertificate == ""
+
+  - name: Validate that the user provided a non-empty TLS key value
+    fail:
+      msg: "presto.spec.presto.config.tls.key cannot be empty if createSecret: true and secretName != ''"
+    when: meteringconfig_spec.presto.spec.presto.config.tls.key == ""
+  when: meteringconfig_template_presto_tls_secret
+
+- name: Check if we need to create a Presto TLS secret using the user-populated fields
```
chancez: The name here doesn't match the logic. It's just doing validation on the Presto Auth TLS configuration.

images/metering-ansible-operator/roles/meteringconfig/defaults/main.yml
```diff
@@ -140,6 +201,22 @@ meteringconfig_create_presto_aws_credentials: "{{ _presto_spec.config.createAwsC
...
 meteringconfig_create_presto_auth_secrets: "{{ _presto_spec.presto.config.auth.enabled and _presto_spec.presto.config.auth.createSecret | default(false) }}"
 
+#
+# Enabling Presto TLS/Auth by Default
+#
+meteringconfig_ansible_generate_presto_tls_secret: "{{ meteringconfig_spec.tls.enabled and _presto_spec.presto.config.tls.enabled and not _presto_spec.presto.config.tls.createSecret and _presto_spec.presto.config.tls.secretName == '' | default(false) }}"
+meteringconfig_ansible_generate_presto_auth_secret: "{{ meteringconfig_spec.tls.enabled and _presto_spec.presto.config.auth.enabled and not _presto_spec.presto.config.auth.createSecret and _presto_spec.presto.config.auth.secretName == '' | default(false) }}"
+meteringconfig_ansible_generate_presto_secrets: "{{ meteringconfig_ansible_generate_presto_tls_secret or meteringconfig_ansible_generate_presto_auth_secret }}"
+meteringconfig_ansible_generate_presto_tls_secret_name: "{{ presto-server-tls if meteringconfig_ansible_generate_presto_tls_secret else _presto_spec.presto.config.tls.secretName }}"
```
chancez: Are these variables needed since we set them in https://github.com/operator-framework/operator-metering/pull/769/files#diff-0e59c4148c558b24a0bdbbd4229bccd6R134 above?

chancez: Why do we have `meteringconfig_ansible_generate_presto_tls_secret_name`, `meteringconfig_ansible_generate_presto_auth_secret_name`, `meteringconfig_presto_server_secret_name` and `meteringconfig_presto_client_secret_name`?

images/metering-ansible-operator/roles/meteringconfig/defaults/main.yml
```diff
@@ -104,7 +104,68 @@ _storage_overrides:
...
+    spec:
+      presto:
+        config:
+          tls:
+            certificate: "{{  meteringconfig_presto_server_certificate }}"
+            key: "{{ meteringconfig_presto_server_key }}"
+            caCertificate: "{{ meteringconfig_presto_server_ca_certificate }}"
+
+            enabled: true
+            createSecret: "{{ meteringconfig_presto_tls_create_secret }}"
```
chancez: I'm wondering if we can just set this to `true` directly now. Any reason not to?

images/metering-ansible-operator/roles/meteringconfig/defaults/main.yml
```diff
@@ -104,7 +104,68 @@ _storage_overrides:
...
+            secretName: "{{ meteringconfig_presto_server_secret_name }}"
+          auth:
+            certificate: "{{  meteringconfig_presto_client_certificate }}"
+            key: "{{ meteringconfig_presto_client_key }}"
+            caCertificate: "{{ meteringconfig_presto_client_ca_certificate }}"
+
+            enabled: true
+            createSecret: "{{ meteringconfig_presto_auth_create_secret }}"
```
chancez: I'm wondering if we can just set this to `true` directly now. Any reason not to?

images/metering-ansible-operator/roles/meteringconfig/defaults/main.yml
timflannagan1: The reason I defaulted this to false was that I use the variable `meteringconfig_ansible_generate_presto_tls_secret` (which relies on `createSecret: false`) as the conditional on a couple of these tasks. If by default, I want ansible to generate the TLS data for me, then I need `createSecret: false`. Else, anytime the user wants us to create the secret for them, they're going to specify `createSecret: true` inside their metering config, which will override the default values.yaml.

chancez: What if we just use `meteringconfig_tls_enabled` as the source everywhere for "do all TLS stuff for the user" to simplify it a bit and and if the user wants to do TLS themselves then they have to set `spec.tls.enabled` to false, and then they would half to do all TLS stuff themselves.

images/metering-ansible-operator/roles/meteringconfig/tasks/configure_presto_tls.yml
```diff
@@ -0,0 +1,146 @@
...
+      type: RSA
+      path: "{{ certificates_dir.path }}/client.key"
+    when: meteringconfig_spec.presto.spec.presto.config.auth.key == ""
+
+  - name: Generate the Presto client CSR
+    openssl_csr:
+      path: "{{ certificates_dir.path }}/client.csr"
+      privatekey_path: "{{ certificates_dir.path }}/client.key"
+      common_name: "coordinator.presto-nodes"
+      subject_alt_name: "DNS:coordinator.presto-nodes.{{ meta.namespace }}.svc.cluster.local,DNS:presto,DNS:coordinator.presto-nodes"
```
chancez: I don't believe we need a SAN for client certs.

images/metering-ansible-operator/roles/meteringconfig/tasks/configure_presto_tls.yml
```diff
@@ -0,0 +1,146 @@
...
+      size: 2048
+      type: RSA
+      path: "{{ certificates_dir.path }}/client.key"
+    when: meteringconfig_spec.presto.spec.presto.config.auth.key == ""
+
+  - name: Generate the Presto client CSR
+    openssl_csr:
+      path: "{{ certificates_dir.path }}/client.csr"
+      privatekey_path: "{{ certificates_dir.path }}/client.key"
+      common_name: "coordinator.presto-nodes"
```
chancez: Technically this is currently being used by all clients connecting to presto. Let's call this "presto-clients", until we move to using a client cert for each component talking to Presto.

images/metering-ansible-operator/roles/meteringconfig/tasks/deploy_resources.yml
```diff
@@ -46,6 +46,7 @@
     namespace: "{{ meta.namespace }}"
     definition: "{{ updated_template_results | flatten }}"
     merge_type: ['merge', 'strategic-merge']
+  no_log: true
```
chancez: Without this, are the secrets being printed? If so..this will make it really difficult to debug other errors.

timflannagan1: That was a debug artifact I forgot to delete - Updated

images/metering-ansible-operator/roles/meteringconfig/tasks/configure_tls.yml
```diff
@@ -0,0 +1,98 @@
...
+
+- name: Update meteringconfig values for enabling TLS in Presto
+  set_fact:
+    meteringconfig_presto_server_ca_certificate: "{{ presto_secret_tls_buf.resources[0].data['ca.crt'] | b64decode }}"
+    meteringconfig_presto_server_certificate: "{{ presto_secret_tls_buf.resources[0].data['tls.crt'] | b64decode }}"
+    meteringconfig_presto_server_key: "{{ presto_secret_tls_buf.resources[0].data['tls.key'] | b64decode }}"
+  no_log: true
+  when: presto_secret_tls_buf is defined and (presto_secret_tls_buf.resources | length > 0)
+
+- name: Update meteringconfig values for enabling TLS in reporting-operator
```
chancez: Does it make sense for this to go into `configure_reporting_operator_tls.yml`?

images/metering-ansible-operator/roles/meteringconfig/tasks/configure_tls.yml
```diff
@@ -0,0 +1,98 @@
...
+
+- name: Update meteringconfig values for enabling auth in Presto
+  set_fact:
+    meteringconfig_presto_client_ca_certificate: "{{ presto_secret_auth_buf.resources[0].data['ca.crt'] | b64decode }}"
+    meteringconfig_presto_client_certificate: "{{ presto_secret_auth_buf.resources[0].data['tls.crt'] | b64decode }}"
+    meteringconfig_presto_client_key: "{{ presto_secret_auth_buf.resources[0].data['tls.key'] | b64decode }}"
+  no_log: true
+  when: presto_secret_auth_buf is defined and (presto_secret_auth_buf.resources | length > 0) 
+
+- name: Update meteringconfig values for enabling auth in reporting-operator
```
chancez: Does it make sense for this to go into `configure_reporting_operator_tls.yml`?

images/metering-ansible-operator/roles/meteringconfig/tasks/configure_tls.yml
```diff
@@ -0,0 +1,98 @@
...
+  no_log: true
+  when: presto_secret_tls_buf is defined and (presto_secret_tls_buf.resources | length > 0)
+
+- name: Update meteringconfig values for enabling TLS in reporting-operator
+  set_fact:
+    meteringconfig_reporting_operator_presto_server_ca_certificate: "{{ reporting_operator_tls_secret_buf.resources[0].data['ca.crt'] | b64decode }}"
+  no_log: true
+  when: reporting_operator_tls_secret_buf is defined and (reporting_operator_tls_secret_buf.resources | length > 0)
+
+- name: Update meteringconfig values for enabling auth in Presto
```
chancez: Does it make sense for this to go into `configure_presto_tls.yml`?

images/metering-ansible-operator/roles/meteringconfig/tasks/configure_presto_tls.yml
```diff
@@ -0,0 +1,125 @@
+---
+#
+# Setup
+#
+- name: Create temporary directory to store all the necessary certificates/keys
+  tempfile:
+    suffix: certificates
+    state: directory
+  register: certificates_dir
```
chancez: We never clean this up

images/metering-ansible-operator/roles/meteringconfig/tasks/configure_presto_tls.yml
```diff
@@ -0,0 +1,119 @@
...
+  - name: Create temporary directory to store all the necessary certificates/keys
+    tempfile:
+      suffix: certificates
+      state: directory
+    register: certificates_dir
+
+  - name: Generate server and client certificiates/keys using openssl
+    include_tasks: configure_presto_openssl.yml
+
+  - name: Update the meteringconfig_spec to use the certificates and keys created in the task file configure_presto_openssl.yml
```
chancez: This name feels a bit verbose. Perhaps something like:
```
- name: Configure components to use generated TLS files"
```

images/metering-ansible-operator/roles/meteringconfig/tasks/configure_presto_tls.yml
```diff
@@ -0,0 +1,119 @@
...
+  - name: Cleanup the temporary directory which held the certificates and keys
+    file:
+      path: "{{ certificates_dir.path }}"
+      state: absent
+  when: meteringconfig_tls_enabled and (presto_secret_tls_buf.resources | length == 0 or presto_secret_auth_buf.resources | length == 0)
+
+#
+# Update the MeteringConfig values to use the pre-existing Presto TLS/auth secret data (when the secret already exists)
+#
+- name: Update meteringconfig values for enabling TLS in Presto
```
chancez:
```
- name: "Configure Presto to use existing TLS data"
```

images/metering-ansible-operator/roles/meteringconfig/tasks/configure_presto_tls.yml
```diff
@@ -0,0 +1,119 @@
...
+  set_fact:
+    meteringconfig_presto_server_ca_certificate: "{{ presto_secret_tls_buf.resources[0].data['ca.crt'] | b64decode }}"
+    meteringconfig_presto_server_certificate: "{{ presto_secret_tls_buf.resources[0].data['tls.crt'] | b64decode }}"
+    meteringconfig_presto_server_key: "{{ presto_secret_tls_buf.resources[0].data['tls.key'] | b64decode }}"
+  no_log: true
+  when: meteringconfig_spec.presto.spec.presto.config.tls.enabled and presto_secret_tls_buf is defined and (presto_secret_tls_buf.resources | length > 0)
+
+- name: Update meteringconfig values for enabling auth in Presto
```
chancez:
```
- name: "Configure Presto to use existing client TLS data"
```

images/metering-ansible-operator/roles/meteringconfig/tasks/configure_reporting_operator_tls.yml
```diff
@@ -0,0 +1,39 @@
...
+  k8s_facts:
+    api_version: v1
+    kind: Secret
+    name: "{{ meteringconfig_spec['reporting-operator'].spec.config.prestoAuth.secretName }}"
+    namespace: "{{ meta.namespace }}"
+  no_log: true
+  register: reporting_operator_auth_secret_buf
+  when: meteringconfig_tls_enabled or meteringconfig_template_presto_auth_secret
+
+- name: Update meteringconfig values for enabling auth in reporting-operator
```
chancez:
```
- name: "Configure reporting-operator to use existing presto client auth TLS secret data"
```

images/metering-ansible-operator/roles/meteringconfig/tasks/configure_reporting_operator_tls.yml
```diff
@@ -0,0 +1,39 @@
...
+  set_fact:
+    meteringconfig_reporting_operator_presto_server_ca_certificate: "{{ reporting_operator_auth_secret_buf.resources[0].data['ca.crt'] | b64decode }}"
+    meteringconfig_reporting_operator_presto_client_ca_certificate: "{{ reporting_operator_auth_secret_buf.resources[0].data['ca.crt'] | b64decode }}"
+    meteringconfig_reporting_operator_client_cert: "{{ reporting_operator_auth_secret_buf.resources[0].data['tls.crt'] | b64decode }}"
+    meteringconfig_reporting_operator_client_key: "{{ reporting_operator_auth_secret_buf.resources[0].data['tls.key'] | b64decode }}"
+  no_log: true
+  when: reporting_operator_auth_secret_buf is defined and (reporting_operator_auth_secret_buf.resources | length > 0)
+
+- name: Update meteringconfig values for enabling TLS in reporting-operator
```
chancez:
```
- name: "Configure reporting-operator to use existing presto server TLS secret data"
```

images/metering-ansible-operator/roles/meteringconfig/tasks/configure_presto_tls.yml
```diff
@@ -0,0 +1,119 @@
...
+#
+- name: Configure TLS and client-side authentication for Presto and reporting-operator
+  block:
+  - name: Create temporary directory to store all the necessary certificates/keys
+    tempfile:
+      suffix: certificates
+      state: directory
+    register: certificates_dir
+
+  - name: Generate server and client certificiates/keys using openssl
```
chancez: "Generate presto server and client TLS certificates and keys"

images/metering-ansible-operator/roles/meteringconfig/tasks/validate_presto_tls.yml
```diff
@@ -0,0 +1,45 @@
...
+    fail:
+      msg: "presto.spec.presto.config.auth.caCertificate cannot be empty if createSecret: true and secretName != ''"
+    when: meteringconfig_spec.presto.spec.presto.config.auth.caCertificate == ""
+
+  - name: Validate that the user provided a non-empty auth key value
+    fail:
+      msg: "presto.spec.presto.config.auth.key cannot be empty if createSecret: true and secretName != ''"
+    when: meteringconfig_spec.presto.spec.presto.config.auth.key == ""
+
+  - name: Validate that the user enabled either TLS by itself or both TLS and authentication
```
chancez: "Validate that TLS is enabled when auth is enabled"

PR #771, opened 2019-06-24
charts,images,manifests: Update reporting-operator configuration schema

chancez:
Updates the MeteringConfig spec.reporting-operator section to have more
consistency in configuring options. This is the first of many changes,
where I will be updating all components to have consistency across
components and consistency within components.

### DIFF
--- a/charts/openshift-metering/templates/reporting-operator/reporting-operator-deployment.yaml
+++ b/charts/openshift-metering/templates/reporting-operator/reporting-operator-deployment.yaml
@@ -1,6 +1,6 @@
 {{- $operatorValues :=  index .Values "reporting-operator" -}}
 {{/* When using TLS but set probes to use https */}}
-{{- if $operatorValues.spec.config.tls.enabled -}}
+{{- if $operatorValues.spec.config.tls.api.enabled -}}
 {{- $_ := set $operatorValues.spec.readinessProbe.httpGet "scheme" "HTTPS" -}}
 {{- $_ := set $operatorValues.spec.livenessProbe.httpGet "scheme" "HTTPS" -}}
 {{- end }}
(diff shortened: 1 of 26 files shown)


charts/openshift-metering/templates/reporting-operator/reporting-operator-deployment.yaml
```diff
@@ -365,60 +372,62 @@ spec:
...
           name: cookie-secret
         - mountPath: /etc/proxy/htpasswd
           name: htpasswd-secret
-{{- if $operatorValues.spec.authProxy.authenticatedEmailsEnabled  }}
+{{- if $operatorValues.spec.authProxy.authenticatedEmails.enabled  }}
         - mountPath: /etc/proxy/authenticated-emails
           name: authenticated-emails-secret
 {{- end }}
-{{- end }}
+{{- end }}{{/* end of authProxy.enabled */}}
```
timflannagan1: nice, I had a hard time following this when I was initially looking at the reporting operator deployment.

charts/openshift-metering/values.yaml
```diff
@@ -212,146 +212,206 @@ reporting-operator:
...
-      disablePrometheusMetricsImporter: "false"
-      enableFinalizers: "false"
-      hiveHost: hive-server:10000
-      leaderLeaseDuration: 60s
-      logDDLQueries: "false"
-      logDMLQueries: "false"
+      enableFinalizers: false
+      leaderLeaseDuration: "60s"
+
+      logDDLQueries: false
```
timflannagan1: I still prefer having a `logging` field that would store these, but if you think unnecessary, that's fine with me.

chancez: I just feel like most components may not have many logging options, and they're all kind app specific so it seems generalizing this isn't as useful.

PR #773, opened 2019-06-25
Add TLS to Hive Metastore

EmilyM1:


images/metering-ansible-operator/roles/meteringconfig/tasks/deploy_resources.yml
```diff
@@ -7,7 +7,6 @@
     loop_var: resource
     label: "{{ resource.template_file }}"
   when: resource.create | default(true)
-  no_log: true
```
timflannagan1: we still need the `no_log: true` on the helm templating task

charts/openshift-metering/templates/presto/hive-metastore-statefulset.yaml
```diff
@@ -164,3 +172,6 @@ spec:
         persistentVolumeClaim:
           claimName: {{ .Values.presto.spec.config.sharedVolume.claimName }}
 {{- end}}
+      - name: hive-metastore-secret
+        secret:
+          secretName: metastoresecret
```
timflannagan1: you could store the `secretName` in the `charts/openshift-metering/values.yaml` to avoid hardcoding the name in the hive metastore statefulset

charts/openshift-metering/templates/presto/hive-metastore-statefulset.yaml
```diff
@@ -120,6 +120,14 @@ spec:
...
         resources:
 {{ toYaml .Values.presto.spec.hive.metastore.resources | indent 10 }}
+      - name: ghostunnel-server
+        command: ["ghostunnel"]
+        args: ["server", "--listen", "localhost:8443", "--target", "localhost:9083", "--key", "/opt/hive/tls/tls.key", "--cert", "/opt/hive/tls/tls.crt", "--cacert", "/opt/hive/tls/ca.crt", "--allow-cn", "Any"]
+        image: quay.io/openshift/origin-ghostunnel
+        volumeMounts:
+        - name: hive-metastore-secret
+          mountPath: /opt/hive/tls
+          readOnly: true
```
timflannagan1: secret volume mounts are `readOnly: true` by default. See: https://github.com/kubernetes/kubernetes/issues/62099#issuecomment-378809922

charts/openshift-metering/templates/presto/hive-metastore-statefulset.yaml
timflannagan1: for example, you could create a mapping in `presto.spec.hive.config` that would handle the TLS-related values. We typically use the format in other TLS sections (like reporting-operator, Presto):

```
tls:
  enabled: false
  createSecret: false
  secretName: ""
```

as the defaults values

charts/openshift-metering/templates/presto/hive-metastore-statefulset.yaml
```diff
@@ -120,6 +120,14 @@ spec:
 {{- end }}
         resources:
 {{ toYaml .Values.presto.spec.hive.metastore.resources | indent 10 }}
+      - name: ghostunnel-server
+        command: ["ghostunnel"]
+        args: ["server", "--listen", "localhost:8443", "--target", "localhost:9083", "--key", "/opt/hive/tls/tls.key", "--cert", "/opt/hive/tls/tls.crt", "--cacert", "/opt/hive/tls/ca.crt", "--allow-cn", "Any"]
```
timflannagan1: we could also use an environment variable to control the location of the key, cert, CA cert instead of hardcoding the values in the argument array

chancez: Also prefer a hyphen based list for long list of args. Eg:

```
args:
- server
- --listen
- localhhost:8443
- --target
- localhost:9083
....
```
etc
```

File: charts/openshift-metering/templates/presto/hive-metastore-statefulset.yaml [hunk 5]
```diff
@@ -120,6 +120,14 @@ spec:
...
+        image: quay.io/openshift/origin-ghostunnel
```
chancez: Your container definition exposes no `ports`. It should at least expose 8443. Also, the hive-metastore service could probably be updated to include the new port so we can connect to ghostunnel.

charts/openshift-metering/templates/presto/hive-metastore-statefulset.yaml
```diff
@@ -120,6 +120,14 @@ spec:
...
+        volumeMounts:
+        - name: hive-metastore-secret
```
chancez: Probably rename the volume and volumeMount names to "hive-tls-secrets" to better represent what the volume/mounts are.

charts/openshift-metering/templates/presto/hive-metastore-statefulset.yaml
```diff
@@ -164,3 +172,6 @@ spec:
         persistentVolumeClaim:
           claimName: {{ .Values.presto.spec.config.sharedVolume.claimName }}
 {{- end}}
+      - name: hive-metastore-secret
```
chancez: "hive-tls-secrets"

charts/openshift-metering/templates/presto/hive-metastore-statefulset.yaml
EmilyM1: When I looked at it by way of kubectl get secrets it was read write

charts/openshift-metering/templates/presto/hive-metastore-statefulset.yaml
EmilyM1: Can do, but hive-server will have tls secrets as well. What will the hive server secret name be to keep them distinct?

chancez: `hive-metastore-tls-secrets`?

charts/openshift-metering/templates/presto/hive-metastore-statefulset.yaml
```diff
@@ -120,6 +120,14 @@ spec:
 {{- end }}
         resources:
 {{ toYaml .Values.presto.spec.hive.metastore.resources | indent 10 }}
+      - name: ghostunnel-server
```
chancez: We should have a conditional around this new container section so that we can have it disabled by default, with an option to enable it.

Eg:

```
{{- if .Values.presto.spec.hive.config.tls.enabled }}
      - name: ghostunnel-server
        command: ["ghostunnel"]
....
{{- end }}
```

charts/openshift-metering/templates/presto/hive-metastore-statefulset.yaml
chancez: +1 to the condition. Same one should be used around the `ghostunnel-server` container too: https://github.com/operator-framework/operator-metering/pull/773/files#r297341285

charts/openshift-metering/templates/presto/hive-metastore-statefulset.yaml
chancez: Note, that this is required for CI to pass.

charts/openshift-metering/templates/presto/hive-metastore-statefulset.yaml
chancez: Also the condition is necessary for CI to pass since we don't have any secret created by default.

charts/openshift-metering/templates/presto/hive-metastore-statefulset.yaml
```diff
@@ -120,6 +120,36 @@ spec:
 {{- end }}
         resources:
 {{ toYaml .Values.presto.spec.hive.metastore.resources | indent 10 }}
+{{- if .Values.presto.spec.hive.metastore.config.tls.enabled }}
+      - name: ghostunnel-server
+        env:
+        - name: HIVE_METASTORE_CA_CRTFILE
+          value: opt/hive/tls/ca.crt
```
chancez: These should be absolute paths.

charts/openshift-metering/templates/presto/hive-metastore-statefulset.yaml
```diff
@@ -120,6 +120,36 @@ spec:
...
+        - name: HIVE_METASTORE_KEYFILE
+          value: opt/hive/tls/tls.key
```
chancez: These should be absolute paths.

charts/openshift-metering/templates/presto/hive-metastore-statefulset.yaml
```diff
@@ -120,6 +120,36 @@ spec:
...
+        - name: HIVE_METASTORE_CRTFILE
+          value: opt/hive/tls/tls.crt
```
chancez: These should be absolute paths.

charts/openshift-metering/templates/presto/hive-metastore-statefulset.yaml
```diff
@@ -120,6 +120,36 @@ spec:
...
+        command: ["ghostunnel"]
+        args:
+        - server
+        - --listen
+        - localhost:8443
+        - --target
+        - localhost:9083
+        - --key
+        -  HIVE_METASTORE_KEYFILE
```
timflannagan1: does this need to be prefixed w/ `$` as it's an environment variable?

chancez: Env vars need $, and to actually have it evaluated by Kubernetes, we need https://kubernetes.io/docs/tasks/inject-data-application/define-environment-variable-container/#using-environment-variables-inside-of-your-config

charts/openshift-metering/templates/presto/hive-metastore-statefulset.yaml
```diff
@@ -164,3 +194,8 @@ spec:
         persistentVolumeClaim:
           claimName: {{ .Values.presto.spec.config.sharedVolume.claimName }}
 {{- end}}
+{{- if .Values.presto.spec.hive.metastore.config.tls.enabled }}
+      - name: hive-metastore-tls-secret
+        secret:
+          secretName: {{ .Values.presto.spec.config.tls.secretName }}
```
timflannagan1: this appears to be undefined, is it not supposed to be `secretName: "{{ .Values.presto.spec.hive.metastore.tls.secretName }}"`

chancez: Actually it should be `"{{ .Values.presto.spec.hive.metastore.config.tls.secretName }}"` I believe. Note the `config`.

charts/openshift-metering/templates/presto/hive-metastore-statefulset.yaml
```diff
@@ -120,6 +120,36 @@ spec:
...
+        - --key
+        -  HIVE_METASTORE_KEYFILE
+        - --cert
+        - HIVE_METASTORE_CRTFILE
+        - --cacert
+        -  HIVE_METASTORE_CA_CRTFILE
+        - --allow-cn
+        - Any
+        image: quay.io/openshift/origin-ghostunnel
+        containerPort: 8443
```
chancez: This needs to be in a `ports` config.

charts/openshift-metering/templates/presto/hive-metastore-statefulset.yaml
```diff
@@ -120,6 +120,36 @@ spec:
...
+        - localhost:9083
+        - --key
+        -  HIVE_METASTORE_KEYFILE
+        - --cert
+        - HIVE_METASTORE_CRTFILE
+        - --cacert
+        -  HIVE_METASTORE_CA_CRTFILE
+        - --allow-cn
+        - Any
+        image: quay.io/openshift/origin-ghostunnel
```
chancez: A few more things:
- Can you move `image:` to be immediate after `name:` to match other containers.
- Use a variable for the image value, eg `image: "{{ .Values.presto.spec.ghostunnel.image.repository }}:{{ .Values.presto.spec.ghostunnel.image.tag }}"`.
- Also add `imagePullPolicy` to the container, also using a variable.

charts/openshift-metering/values.yaml
```diff
@@ -458,6 +463,10 @@ presto:
       metastore:
         affinity: {}
         config:
+          tls:
+            enabled: false
```
timflannagan1: We will need a follow-up PR that adds the `certificate`, `key`, and `caCertificate` fields so the user can specify those directly.

timflannagan1:
I think this looks much better. We'll need to create a template for these secrets next, and add that template secret hash to the `hive-metastore-statefulset` in a follow-up PR.

PR #780, opened 2019-06-26
Enable the ansible-operator to store Root CA data in a secret by default

timflannagan1:


timflannagan1:
Reporting operator Presto secrets are being overwritten. 
/hold

timflannagan1:
Realized that I committed a commented-out `include_task: configure_reporting_operator_tls.yml`
/hold remove

timflannagan1:
Realized that I committed a commented-out include_task: configure_reporting_operator_tls.yml
/hold cancel

images/metering-ansible-operator/roles/meteringconfig/defaults/main.yml
```diff
@@ -192,6 +200,7 @@ meteringconfig_create_metering_default_storage: "{{ _openshift_reporting_spec.de
 
 meteringconfig_create_metering_monitoring_resources: "{{ _monitoring_conf.enabled | default(true) }}"
 meteringconfig_create_metering_monitoring_rbac: "{{ _monitoring_conf.enabled and _monitoring_conf.enabled and _monitoring_conf.createRBAC | default(true) }}"
+# meteringconfig_create_metering_root_ca_secret: "{{ meteringconfig_spec.tls.enabled }}"
```
chancez: ?

images/metering-ansible-operator/roles/meteringconfig/tasks/configure_root_ca.yml
```diff
@@ -0,0 +1,70 @@
+---
+
+- name: Check if Root CA secret already exists
+  k8s_facts:
+    api_version: v1
+    kind: Secret
+    name: "{{ meteringconfig_spec.tls.secretName }}"
+    namespace: "{{ meta.namespace }}"
+  no_log: true
+  register: root_ca_secret_buf
```
chancez: Should we only be checking if tls is enabled?

images/metering-ansible-operator/roles/meteringconfig/tasks/configure_tls.yml
```diff
@@ -1,7 +1,21 @@
...
+- name: Generate root certificate authority
+  include_tasks: configure_root_ca.yml
+
 - name: Configure TLS and client-side authentication in Presto
   include_tasks: configure_presto_tls.yml
 
 - name: Configure TLS and authentication in the reporting-operator
   include_tasks: configure_reporting_operator_tls.yml
+
+- name: Cleanup the temporary directory which held the certificates and keys
```
chancez: This will be skipped if the previous tasks fail. The other includes need to be in a block with this as the `always:`

images/metering-ansible-operator/roles/meteringconfig/tasks/reconcile_metering.yml
```diff
@@ -11,3 +11,6 @@
       - template_file: templates/metering/metering-rolebindings.yaml
         apis: [ {kind: rolebindings, api_version: 'rbac.authorization.k8s.io/v1'} ]
         prune_label_value: openshift-metering-rolebindings
+      - template_file: templates/metering/metering-root-ca-secret.yaml
+        apis: [ {kind: Secret} ]
+        prune_label_value: openshift-metering-root-ca-secret
```
chancez: maybe we want `create: "{{ meteringconfig_enable_tls }}"` since we only really need this if we're generating the cert right?

timflannagan1: I was under the impression that we were always going to create this after our discussion this morning. I had the `create: ` value conditioned on `meteringconfig_enable_tls` initially. 

No problem, I can update this.

images/metering-ansible-operator/roles/meteringconfig/tasks/configure_root_ca.yml
```diff
@@ -0,0 +1,70 @@
...
+  - name: Use generate cert/key when secret doesn't already exist
+    set_fact:
+      meteringconfig_tls_root_ca_certificate: "{{ lookup('file', '{{ certificates_dir.path }}/ca.crt') + '\n' }}"
+      meteringconfig_tls_root_ca_key: "{{ lookup('file', '{{ certificates_dir.path }}/ca.key') + '\n' }}"
+
+  vars:
+    root_ca_secret_exists: "{{ root_ca_secret_buf.resources | length > 0 }}"
+  when: not root_ca_secret_exists
+
+- name: Use pre-existing secret cert/key when secret already exists
```
chancez: Should refer to which cert/key; ie the "CA" cert/key.
```suggestion
- name: Use pre-existing CA secret cert/key when secret already exists
```

charts/openshift-metering/values.yaml
```diff
@@ -31,7 +31,11 @@ storage:
   #     mountPath: "/"
 
 tls:
-  enabled: false
+  enabled: true
+  # the name of the secret containing root CA key/cert
+  certificate: ""
+  key: ""
+  secretName: "metering-operator-root-ca-secret"
```
chancez: Maybe drop the `-secret` suffix

images/metering-ansible-operator/roles/meteringconfig/tasks/reconcile_metering.yml
chancez: Well I was trying to think what we would put into it when tls is disabled, and it seems like nothing, so that's why I was thinking this.

images/metering-ansible-operator/roles/meteringconfig/tasks/configure_tls.yml
timflannagan1: good catch

images/metering-ansible-operator/roles/meteringconfig/tasks/configure_tls.yml
```diff
@@ -1,7 +1,24 @@
...
+  - name: Configure TLS and client-side authentication in Presto
+    include_tasks: configure_presto_tls.yml
+
+  - name: Configure TLS and authentication in the reporting-operator
+    include_tasks: configure_reporting_operator_tls.yml
+
+  always:
+  - name: Cleanup the temporary directory which held the certificates and keys
+    file:
+      path: "{{ certificates_dir.path }}"
```
chancez: Since `certificates_dir` is in the block, does the scope here work? I guess since it's not evaluated until later it might be okay?

timflannagan1: Scoping should be a problem as I also do this:
```
  vars:
    root_ca_secret_exists: "{{ root_ca_secret_buf and root_ca_secret_buf.resources | length > 0 }}"
  when: meteringconfig_tls_enabled and not root_ca_secret_exists
```

which evaluates properly

PR #781, opened 2019-06-26
Allow running reporting-operator locally against TLS secured Presto

chancez:
Necessary for running e2e/integration tests locally with reporting-operator.

### DIFF
--- a/hack/run-reporting-operator-local.sh
+++ b/hack/run-reporting-operator-local.sh
@@ -22,8 +22,31 @@ source "${ROOT_DIR}/hack/common.sh"
 : "${METERING_HIVE_HOST:="127.0.0.1:${METERING_HIVE_PORT_FORWARD_PORT}"}"
 : "${METERING_PROMETHEUS_HOST:="127.0.0.1:${METERING_PROMETHEUS_PORT_FORWARD_PORT}"}"
 
+: "${METERING_PRESTO_USE_TLS:=true}"
+
+TMPDIR="$(mktemp -d)"
+
+function cleanup() {
+    set +e +o pipefail
+    exit_status=$?
+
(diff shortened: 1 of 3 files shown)


hack/run-reporting-operator-local.sh
```diff
@@ -42,6 +65,44 @@ else
...
+        sleep 5
+    done
+
+    export REPORTING_OPERATOR_PRESTO_USE_TLS=true
+    export REPORTING_OPERATOR_PRESTO_CA_FILE="$TMPDIR/reporting-operator-presto-server-ca.crt"
+    export REPORTING_OPERATOR_PRESTO_USE_AUTH=true
+    export REPORTING_OPERATOR_PRESTO_CLIENT_CERT_FILE="$TMPDIR/reporting-operator-presto-client-tls.crt"
+    export REPORTING_OPERATOR_PRESTO_CLIENT_KEY_FILE="$TMPDIR/reporting-operator-presto-client-tls.key"
+    export REPORTING_OPERATOR_PRESTO_CLIENT_CA_CERT_FILE="$TMPDIR/reporting-operator-presto-client-ca.crt"
+    export REPORTING_OPERATOR_PRESTO_TLS_INSECURE_SKIP_VERIFY=true
```
timflannagan1: Do we need the client CA file anymore? We deleted that in the most recent update to the Presto stuff.

Also, where else is `REPORTING_OPERATOR_PRESTO_TLS_INSECURE_SKIP_VERIFY` used?

chancez: Reporting-operator reads the env vars natively, these map to the flags. And yes, `REPORTING_OPERATOR_PRESTO_CLIENT_CA_CERT_FILE` can be removed

PR #786, opened 2019-06-27
Update openshift-reporting MeteringConfig schema

chancez:
Primarily moves defaultReportDataSources into a list, and supports disabling
the creation of these resources.

Additionally, the first two commits fix issues with pruning resources I
discovered while testing this.

### DIFF
--- a/charts/openshift-metering/templates/openshift-reporting/_reporting_helpers.tpl
+++ b/charts/openshift-metering/templates/openshift-reporting/_reporting_helpers.tpl
@@ -1,0 +1,10 @@
+{{- define "new-report-datasource" }}
+apiVersion: metering.openshift.io/v1
+kind: ReportDataSource
+metadata:
+  name: "{{ .name }}"
+  labels:
+    operator-metering: "true"
+spec:
+{{ toYaml .spec | indent 2 }}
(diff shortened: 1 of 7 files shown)


charts/openshift-metering/templates/openshift-reporting/_reporting_helpers.tpl
```diff
@@ -0,0 +1,10 @@
+{{- define "new-report-datasource" }}
+apiVersion: metering.openshift.io/v1alpha1
```
chancez: Self reminder that this will need updating after #788 merges

images/metering-ansible-operator/roles/meteringconfig/tasks/reconcile_reporting.yml
```diff
@@ -6,55 +6,56 @@
     values_file: /tmp/metering-values.yaml
     resources:
       - template_file: templates/metering/default-storage-location.yaml
-        apis: [ {kind: storagelocation} ]
+        apis: [ {kind: storagelocation, api_version: 'metering.openshift.io/v1alpha1'} ]
```
chancez: These will also need updating after #788 merges

PR #791, opened 2019-06-28
images/metering-ansible-operator: Fix root_ca_secret_exists variable when meteringconfig_tls_enabled is false.

timflannagan1:


images/metering-ansible-operator/roles/meteringconfig/tasks/configure_root_ca.yml
```diff
@@ -44,7 +44,7 @@
       meteringconfig_tls_root_ca_key: "{{ lookup('file', '{{ certificates_dir.path }}/ca.key') + '\n' }}"
 
   vars:
-    root_ca_secret_exists: "{{ root_ca_secret_buf and root_ca_secret_buf.resources | length > 0 }}"
+    root_ca_secret_exists: "{{ root_ca_secret_buf and root_ca_secret_buf.resources is defined and root_ca_secret_buf.resources | length > 0 }}"
```
chancez: Shouldn't it be  `root_ca_secret_buf  is defined and root_ca_secret_buf.resources | length > 0`?

timflannagan1: That line is definitely off, but I was still getting errors with `root_ca_secret_buf is defined`, so I changed it to `root_ca_secret_buf.resources is defined`.

I'm not sure why it wasn't working with the former.

timflannagan1: In the case where you want to reproduce this, all you would need to do is update `configure_root_ca.yml` with the `root_ca_secret_buf is defined ...`, configure a meteringconfig with `tls.enabled: false`, and it will result in an error.

chancez: Ok

PR #800, opened 2019-07-01
charts,images: Support using both pre/post-kubernetes 1.14 metrics

chancez:
Updates our Prometheus queries to be prepared for https://github.com/kubernetes/enhancements/blob/master/keps/sig-instrumentation/20181106-kubernetes-metrics-overhaul.md.
Sometime in the future the old metric labels will be removed, so switch
our queries to use the new metric labels.

To support existing users on Kubernetes versions prior to 1.14, they can
set `spec.openshift-reporting.spec.defaultReportDataSources.postKube_1_14`
to false.

Ex:

```
spec:
  openshift-reporting:
    spec:
      defaultReportDataSources:
        postKube_1_14:
          enabled: false
```

### DIFF
--- a/charts/openshift-metering/values.yaml
+++ b/charts/openshift-metering/values.yaml
@@ -67,151 +67,172 @@ openshift-reporting:
       enabled: false
 
     defaultReportDataSources:
-      enabled: true
-      items:
-        - name: cluster-cpu-capacity-raw
-          spec:
-            reportQueryView:
-              queryName: cluster-cpu-capacity-raw
-        - name: cluster-cpu-usage-raw
-          spec:
-            reportQueryView:
-              queryName: cluster-cpu-usage-raw
-        - name: cluster-memory-capacity-raw
-          spec:
-            reportQueryView:
-              queryName: cluster-memory-capacity-raw
-        - name: cluster-memory-usage-raw
-          spec:
-            reportQueryView:
-              queryName: cluster-memory-usage-raw
-        - name: node-allocatable-cpu-cores
-          spec:
-            prometheusMetricsImporter:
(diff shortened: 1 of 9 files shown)


chancez:
Just added support to auto-detect kubernetes version and correctly set the default reportDataSources automatically. I felt that this would be easier than fielding questions about this, and I felt like it was a good opportunity to figure out how to do basic version detection, which we can use in the future.

charts/openshift-metering/values.yaml
```diff
@@ -67,151 +67,172 @@ openshift-reporting:
...
+          - name: pod-usage-cpu-cores
+            spec:
+              prometheusMetricsImporter:
+                query: |
+                  label_replace(sum(rate(container_cpu_usage_seconds_total{container_name!="POD",container_name!="",pod_name!=""}[1m])) BY (pod_name, namespace), "pod", "$1", "pod_name", "(.*)") + on (pod, namespace) group_left(node) (sum(kube_pod_info{pod_ip!="",node!="",host_ip!=""}) by (pod, namespace, node) * 0)
+          - name: pod-usage-memory-bytes
+            spec:
+              prometheusMetricsImporter:
+                query: |
+                  sum(label_replace(container_memory_usage_bytes{container_name!="POD", container_name!="",pod_name!=""}, "pod", "$1", "pod_name", "(.*)")) by (pod, namespace) + on (pod, namespace) group_left(node) (sum(kube_pod_info{pod_ip!="",node!="",host_ip!=""}) by (pod, namespace, node) * 0)
```
EmilyM1: So these 2 would have been effected by the metrics name change --- interesting.

PR #787, opened 2019-06-27
images/metering-ansible-operator: Enable reporting-operator authProxy by default.

timflannagan1:
This would configure the `reporting-operator.spec.authProxy` data by default. Like enabling TLS in Presto/reporting-operator by default, this is contingent on the value of the `spec.tls.enabled` key, which is set to true by default.

If the user wants to configure authProxy themselves or disable it entirely, they would need to specify `spec.tls.enabled: false` in their meteringconfig CR, and either elect to enable certain fields like `authProxy.cookie` or `authProxy.htpasswd` or use the default values.

PR #802, opened 2019-07-02
images/metering-ansible-operator: Prefix TLS-related default variables with an underscore.

timflannagan1:
This is based on #787 

~This would still print things like `meteringconfig_tls_enabled`, `meteringconfig_tls_overrides`, `meteringconfig_root_ca_overrides`, but they won't contain any useful information.~


timflannagan1:
Wait for #787 to merge first.
/hold

timflannagan1:
#797 was merged, removing hold
/hold cancel

chancez:
@timflannagan1 I don't think we want to print meteringconfig_tls_overrides since that contains the other data in it

PR #795, opened 2019-06-28
[WIP] Support TLS from  Hive Server to Hive metastore using ghostunnel

EmilyM1:

PR #803, opened 2019-07-02
Support TLS from Hive Server to Hive metastore using ghostunnel

chancez:
Supercedes #795

This adds support for hive-server to communicate with hive-metastore using ghostunnel to provide a secured TLS connection to communicate over.
Also adds the ability to create the hive TLS secrets by providing configuration in your MeteringConfig.

### DIFF
--- a/charts/openshift-metering/templates/hive/hive-server-statefulset.yaml
+++ b/charts/openshift-metering/templates/hive/hive-server-statefulset.yaml
@@ -48,6 +48,32 @@ spec:
       affinity:
 {{ toYaml .Values.hive.spec.server.affinity | indent 8 }}
 {{- end }}
+{{- if .Values.hive.spec.server.config.metastoreTLS.enabled }}
+      initContainers:
+      - name: copy-hive-tls
+        image: {{ .Values.__ghostunnel.image.repository }}:{{ .Values.__ghostunnel.image.tag }}
+        imagePullPolicy: {{ .Values.__ghostunnel.image.pullPolicy }}
+        env:
+        - name: GHOSTUNNEL_CLIENT_CA_CRTFILE
+          value: /opt/hive/tls/ca.crt
+        - name: GHOSTUNNEL_CLIENT_TLS_CERT
+          value: /opt/hive/tls/tls.crt
+        - name: GHOSTUNNEL_CLIENT_TLS_KEY
+          value: /opt/hive/tls/tls.key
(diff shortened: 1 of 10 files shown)


charts/openshift-metering/values.yaml
```diff
@@ -669,6 +678,34 @@ hive:
...
+          caCertificate: ""
+
+          enabled: false
+          createSecret: false
+          secretName: ""
+
+        auth:
+          enabled: false
+
+        metastoreTLS:
```
timflannagan1: I wonder if we can rename this field to something else now that we've deflated the `reporting-operator.spec.config.prestoTLS` into `reporting-operator.spec.config.presto.tls` (and like-wise for `presto.auth`). 

That being said, I'm not sure you need a `metastore` field itself if we're only configuring TLS for metastore.

charts/openshift-metering/templates/hive/hive-server-statefulset.yaml
```diff
@@ -124,6 +152,67 @@ spec:
...
+          value: /opt/hive/tls/ca.crt
+        - name: GHOSTUNNEL_CLIENT_KEYSTORE
+          value: /opt/hive/tls/keystore-combined.pem
+        command: ["ghostunnel"]
+        args:
+        - client
+        - --listen
+        - localhost:8443
+        - --target
+        - hive-metastore:8443
```
timflannagan1: This could potentially go in a `.spec.hive.spec.server.config.metastore` field, if you want to go that route.

charts/openshift-metering/values.yaml
chancez: Yeah, it's basically that. And it doesn't make much since to put into `spec.hive.metastore` either, since this needs to be done for each client of metastore, including presto. For Presto, I put it's TLS config into `spec.presto.config.connectors.hive.tls`.

charts/openshift-metering/templates/hive/hive-server-statefulset.yaml
chancez: It's going to be static anyways, and I'm moving away from fields for `hosts` that are mostly unchanging, and using hardcoded values with conditions instead since it makes it easier to write template logic for.

timflannagan1: Fair enough, this looks fine to me.

chancez:
Seems like tide is stuck so I'm manually merging.

PR #810, opened 2019-07-05
Support configuring hive-server with authentication via ghostunnel

chancez:
- Add support for configuring hive-server to use ghostunnel to handle TLS and authentication via client certificates for authenticating to Hive.
- Updates reporting-operator to support passing client certificates when connecting to Hive via TLS.
- Add support configuring reporting-operator to use TLS client certificates for authentication to hive server in MeteringConfig.

### DIFF
--- a/pkg/operator/operator.go
+++ b/pkg/operator/operator.go
@@ -101,18 +101,23 @@ type Config struct {
 	MetricsListen string
 	PprofListen   string
 
-	HiveHost   string
-	HiveUseTLS bool
-	HiveCAFile string
+	HiveHost                  string
+	HiveUseTLS                bool
+	HiveCAFile                string
+	HiveTLSInsecureSkipVerify bool
+
+	HiveUseClientCertAuth bool
+	HiveClientCertFile    string
+	HiveClientKeyFile     string
 
 	PrestoHost                  string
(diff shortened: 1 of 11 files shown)


pkg/operator/operator.go
```diff
@@ -410,16 +415,29 @@ func (op *Reporting) Run(ctx context.Context) error {
...
-		certPool.AppendCertsFromPEM(cert)
+
 		dialer = hive.TLSDialer{
-			Config: &tls.Config{
-				RootCAs: certPool,
-			},
+			Config: hiveTLSConfig,
 		}
 	} else {
 		dialer = hive.DialWrapper{}
```
timflannagan1: Couldn't we just eliminate this `else` block by initializing `dialer := hive.DialWrapper{}` and override that variable if `op.cfg.HiveUseTLS` evaluates to true?

chancez: These are two different types, and `dialer` is a `hive.Dialer` which is an interface. Setting it to `hive.DialWrapper{}` would make it a concrete type. 

That said, I think I can eliminate the branch by specifying the type when initializing:
```
	var dialer hive.Dialer = hive.DialWrapper{}
```

PR #816, opened 2019-07-08
images/metering-ansible-operator: Rename TLS override variables to avoid being printed.

timflannagan1:
This somehow was reverted when I rebased my original PR to pull in merged master changes.

### DIFF
--- a/images/metering-ansible-operator/roles/meteringconfig/defaults/main.yml
+++ b/images/metering-ansible-operator/roles/meteringconfig/defaults/main.yml
@@ -149,7 +149,7 @@ _meteringconfig_presto_client_key: ""
 _meteringconfig_tls_root_ca_certificate: ""
 _meteringconfig_tls_root_ca_key: ""
 
-meteringconfig_root_ca_overrides:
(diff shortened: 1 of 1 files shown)


images/metering-ansible-operator/roles/meteringconfig/defaults/main.yml
```diff
@@ -148,7 +148,7 @@ _meteringconfig_presto_client_key: ""
 _meteringconfig_tls_root_ca_certificate: ""
 _meteringconfig_tls_root_ca_key: ""
 
-meteringconfig_root_ca_overrides:
+metering_root_ca_overrides:
```
chancez: Maybe just `_meteringconfig_root_ca_overrides`?

timflannagan1: I believe when I initially tried that, I was having a hard time getting that to work. 

I'll try to use this again though.

timflannagan1: Update: seems like that's not the case, and this works as expected.

PR #826, opened 2019-07-10
Standardize CPU_Core_Seconds

EmilyM1:
When going thru report queries I noted most of the time we use `CPU_Core_Seconds` as the unit but there were 1-3 units involving cpu measures that were at `core_seconds` so switching to standardize.


EmilyM1:
Ok, did a final check of all files... with the above commit they will all be at `cpu_core_seconds`.

chancez:
@EmilyM1 Can you fix `rollup-reports.md` line 50 also?

PR #829, opened 2019-07-10
Azure settings overide

commanderchewbacca:
added settings to Ansible, allowing setting the azure credentials in one place to apply to all the different nodes.

### DIFF
--- a/Documentation/configuring-storage.md
+++ b/Documentation/configuring-storage.md
@@ -34,6 +34,22 @@ If you left `spec.storage.hive.s3.createBucket` set to true, or unset, then you 
 Please note that this must be done before installation.
 Changing these settings after installation will result in broken and unexpected behavior.
 
+## Storing data in Azure
+
+You can also store your data in Azure blob storage, and to do so, you must use an existing container.
+Edit the `spec.storage` section in the example [azure-blob-storage.yaml][azure-blob-storage-config]] configuration.
+Set the `spec.storage.hive.azure.container` and `spec.storage.hive.azure.secretName` with the secret following this format:
+```
+apiVersion: v1
+kind: Secret
+metadata:
+  name: your-azure-secret
+data:
+  azure-storage-account-name: "dGVzdAo="
+  azure-secret-access-key: "c2VjcmV0Cg=="
+```
+The `spec.storage.hive.azure.container` should be the container you wish to store metering data at, and the optional `spec.storage.hive.azure.rootDirectory` should be the folder you want your data in, inside the container.
+
 ## Using shared volumes for storage
 
 Metering has no storage by default, but can use any ReadWriteMany PersistentVolume or [StorageClass][storage-classes] that provisions a ReadWriteMany PersistentVolume.
@@ -59,6 +75,7 @@ For more details read [configuring HDFS][configuring-hdfs].
 
 [storage-classes]: https://kubernetes.io/docs/concepts/storage/storage-classes/
 [s3-storage-config]: ../manifests/metering-config/s3-storage.yaml
+[azure-blob-storage-config]: ../manifests/metering-config/azure-blob-storage.yaml
 [shared-storage-config]: ../manifests/metering-config/shared-storage.yaml
 [hdfs-storage-config]: ../manifests/metering-config/hdfs-storage.yaml
 [configuring-hive-metastore]: configuring-hive-metastore.md
--- a/charts/openshift-metering/templates/hadoop/hadoop-azure-credentials.yaml
+++ b/charts/openshift-metering/templates/hadoop/hadoop-azure-credentials.yaml
@@ -6,8 +6,8 @@ metadata:
 data:
 {{- if .Values.hadoop.spec.config.azure.storageAccountName }}
   azure-storage-account-name: {{ .Values.hadoop.spec.config.azure.storageAccountName | b64enc | quote}}
-{{- end}}
+{{- end }}
 {{- if .Values.hadoop.spec.config.azure.secretAccessKey }}
   azure-secret-access-key: {{ .Values.hadoop.spec.config.azure.secretAccessKey | b64enc | quote}}
-{{- end}}
+{{- end }}
 {{- end -}}
--- a/charts/openshift-metering/templates/hive/hive-server-statefulset.yaml
+++ b/charts/openshift-metering/templates/hive/hive-server-statefulset.yaml
@@ -103,7 +103,7 @@ spec:
       - name: hive-scripts
         configMap:
           name: hive-scripts
-          defaultMode: 0555
+          defaultMode: 0775
       containers:
       - name: hiveserver2
         command: ["/hive-scripts/entrypoint.sh"]
@@ -258,7 +258,7 @@ spec:
       - name: hive-scripts
         configMap:
           name: hive-scripts
-          defaultMode: 0555
+          defaultMode: 0775
 {{- if .Values.hive.spec.config.useHadoopConfig }}
       - name: hadoop-config
         emptyDir: {}
--- a/images/metering-ansible-operator/roles/meteringconfig/defaults/main.yml
+++ b/images/metering-ansible-operator/roles/meteringconfig/defaults/main.yml
@@ -3,7 +3,6 @@
 meteringconfig_chart_path: "{{ lookup('env','HELM_CHART_PATH') }}"
 meteringconfig_prune_label_key: 'metering.openshift.io/prune'
 meteringconfig_prune_namespace_label_key: 'metering.openshift.io/ns-prune'
-
 meteringconfig_default_values_file: "{{ meteringconfig_chart_path + '/values.yaml' }}"
 meteringconfig_default_values: "{{ lookup('file', meteringconfig_default_values_file) | from_yaml }}"
 
@@ -53,7 +52,7 @@ meteringconfig_default_image_overrides:
         repository: "{{ lookup('env','METERING_HADOOP_IMAGE').split(':')[0] | default(meteringconfig_hadoop_default_image_repo, true) }}"
         tag: "{{ lookup('env','METERING_HADOOP_IMAGE').split(':')[1] | default(meteringconfig_hadoop_default_image_tag, true) }}"
 
-_storage_overrides:
+_base_storage_overrides:
   s3:
     storage:
       hive:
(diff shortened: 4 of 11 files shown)


charts/openshift-metering/templates/hive/hive-server-statefulset.yaml
```diff
@@ -77,15 +78,32 @@ spec:
...
+              key: azure-secret-access-key
+{{- end }}
         volumeMounts:
         - name: hadoop-config
           mountPath: /hadoop-config
         - name: hadoop-starting-config
           mountPath: /hadoop-starting-config
+        - name: hive-scripts
+          mountPath: /hive-scripts
+      volumes:
```
timflannagan1: I wonder if we should move this section (`volumes`) to be under `containers:`, which is what we do in the other statefulsets/deployments (`initContainers` -> `containers` -> `volumes`)

images/metering-ansible-operator/roles/meteringconfig/defaults/main.yml
```diff
@@ -93,6 +93,48 @@ _storage_overrides:
...
+      hive:
+        # azure storage defaults
+        azure:
+          # Intentionally an invalid default name to ensure that this is overridden, and obvious when it's not.
+          container: "-unconfigured-metering-container-name-"
+          secretName: ""
+          storageAccountName: ""
+          secretAccessKey: ""
+          createSecret: true
+          secretName: ""
```
timflannagan1: you have two `secretName` fields (line 103 and 107)

images/metering-ansible-operator/roles/meteringconfig/tasks/reconcile.yml
```diff
@@ -7,10 +7,10 @@
 
 - name: Validate Hive storage configuration
   fail:
```
timflannagan1: we should update this to use the `assert` module, which is already extensively used throughout this role:
```
- name: Validate Hive storage configuration
  assert:
    that:
      - hiveStorageType is defined
      - hiveStorageType in ['s3', 'sharedPVC', 'hdfs', 'azure']
    msg: "Invalid spec.storage.hive.type: '{{ hiveStorageType }}', must be one of 's3', 'azure', or 'sharedPVC'"
```

timflannagan1: This isn't a big deal, just something that I noticed, and can be changed in future PRs.

charts/openshift-metering/templates/hadoop/hadoop-azure-credentials.yaml
```diff
@@ -0,0 +1,13 @@
...
+kind: Secret
+metadata:
+  name: hadoop-azure-credentials
+data:
+{{- if .Values.hadoop.spec.config.azure.storageAccountName }}
+  azure-storage-account-name: {{ .Values.hadoop.spec.config.azure.storageAccountName | b64enc | quote}}
+{{- end}}
+{{- if .Values.hadoop.spec.config.azure.secretAccessKey }}
+  azure-secret-access-key: {{ .Values.hadoop.spec.config.azure.secretAccessKey | b64enc | quote}}
+{{- end}}
```
timflannagan1: this should be `{{- end }}`

charts/openshift-metering/templates/hadoop/hadoop-azure-credentials.yaml
```diff
@@ -0,0 +1,13 @@
...
+{{- end -}}
```
timflannagan1: missing newline at end of file

images/metering-ansible-operator/roles/meteringconfig/tasks/configure_storage.yml
```diff
@@ -34,3 +35,36 @@
       when: operator_s3_credentials_secret.resources | length > 0
 
   when: meteringconfig_storage_hive_storage_type == 's3' and meteringconfig_storage_s3_create_bucket
+
+- name: Azure account name secret extraction
```
chancez:
```suggestion
- name: Get azure storage account name
```

images/metering-ansible-operator/roles/meteringconfig/tasks/configure_storage.yml
```diff
@@ -34,3 +35,36 @@
...
+  block:
+    - name: Check for secret
```
chancez:
```suggestion
    - name: Get Azure storage credentials secret
```

images/metering-ansible-operator/roles/meteringconfig/tasks/configure_storage.yml
```diff
@@ -34,3 +35,36 @@
...
+      k8s_facts:
+        api_version: v1
+        kind: Secret
+        name: "{{ meteringconfig_spec.storage.hive.azure.secretName }}"
+        namespace: "{{ meta.namespace }}"
+      register: azure_secret_buf
+
+    - name: azure accoutname extraction
```
chancez:
```suggestion
    - name: Extract Azure storage account name
```

images/metering-ansible-operator/roles/meteringconfig/tasks/reconcile.yml
chancez: We can do this in a separate PR

images/metering-ansible-operator/roles/meteringconfig/defaults/main.yml
```diff
@@ -92,6 +92,48 @@ _storage_overrides:
...
+    storage:
+      hive:
+        # azure storage defaults
+        azure:
+          # Intentionally an invalid default name to ensure that this is overridden, and obvious when it's not.
+          container: "-unconfigured-metering-container-name-"
+          secretName: ""
+          storageAccountName: ""
+          secretAccessKey: ""
+          createSecret: true
```
chancez: We don't want to create the secret by default, so this should be false.

manifests/metering-config/azure-blob-storage.yaml
```diff
@@ -0,0 +1,17 @@
...
+  name: "operator-metering"
+spec:
+  storage:
+    type: "hive"
+    hive:
+      type: "azure"
+      azure:
+        container: "bucket1"
+        storageAccountName: "meteringazure"
+        secretAccessKey: "your-secret-here"
```
chancez: Our default example should instruct the user to create a secret and reference it, instead of putting their credentials in the MeteringConfig

manifests/metering-config/azure-blob-storage.yaml
```diff
@@ -0,0 +1,17 @@
...
+        #set to false if you want to reference an exting secret, and set the secretName to the filename
+        createSecret: true
```
chancez: Remove

manifests/metering-config/azure-blob-storage.yaml
```diff
@@ -0,0 +1,17 @@
...
+        secretName: ""
```
chancez: Set to existing secret name, look at `s3-storage.yaml` for an example. Eg: `secretName: "my-azure-secret"`

manifests/metering-config/azure-blob-storage.yaml
```diff
@@ -0,0 +1,17 @@
...
+metadata:
+  name: "operator-metering"
+spec:
+  storage:
+    type: "hive"
+    hive:
+      type: "azure"
+      azure:
+        container: "bucket1"
+        storageAccountName: "meteringazure"
```
chancez: Remove, should be set in the secret

Documentation/configuring-storage.md
```diff
@@ -34,6 +34,25 @@ If you left `spec.storage.hive.s3.createBucket` set to true, or unset, then you
 Please note that this must be done before installation.
 Changing these settings after installation will result in broken and unexpected behavior.
 
+## Storing data in Azure
+
+You can also store your data in Azure blob storage, to do so, you must use an existing container.
```
timflannagan1: "and to do so"

Documentation/configuring-storage.md
```diff
@@ -34,6 +34,25 @@ If you left `spec.storage.hive.s3.createBucket` set to true, or unset, then you
...
+Edit the `spec.storage` section in the example [azure-blob-storage.yaml][azure-blob-storage-config]] configuration.
+Set the  `spec.storage.hive.azure.container` and `spec.storage.hive.azure.storageAccountName` value.
```
timflannagan1: there's an extra space after "the"

Documentation/configuring-storage.md
```diff
@@ -34,6 +34,25 @@ If you left `spec.storage.hive.s3.createBucket` set to true, or unset, then you
...
+`spec.storage.hive.azure.secretAccessKey` must also be set, however, the value can be set with an existing secret. To use an already existing secret, set createSecret to false and leave the two fields blank, with the secret following this format:
```
timflannagan1: set createSecret -> set `createSecret`

Documentation/configuring-storage.md
chancez: Let's instruct the users to only use an existing secret. The other stuff we don't want them to use directly, so we shouldn't mention setting `storageAccountName`, `secretAccessKey` or `createSecret`. Similar to S3.

Documentation/configuring-storage.md
```diff
@@ -34,6 +34,25 @@ If you left `spec.storage.hive.s3.createBucket` set to true, or unset, then you
...
+apiVersion: v1
+kind: Secret
+metadata:
+  name: your-azure-secret
+data:
+  azure-storage-account-name: "dGVzdAo="
+  azure-secret-access-key: "c2VjcmV0Cg=="
+```
+The `spec.storage.hive.azure.container` should be the container you wish to store metering data at, and the optional `spec.storage.hive.azure.roodDirectory` should be the folder you want your data in, inside the container.
+Note that `spec.storage.hive.azure.createSecret` cannot be true with `spec.storage.hive.azure.secretName` non-empty at the same time.
```
timflannagan1: ... cannot be true with a non-empty `spec.storage.hive.azure.secretName` at the same time.

images/metering-ansible-operator/roles/meteringconfig/tasks/configure_presto_openssl.yml
```diff
@@ -3,6 +3,7 @@
 #
 # Generate the Presto Server cert/key
 #
+
```
timflannagan1: remove this newline

Documentation/configuring-storage.md
chancez: We don't want users to set createSecret, so this is irrelevant

images/metering-ansible-operator/roles/meteringconfig/defaults/main.yml
```diff
@@ -314,7 +377,7 @@ meteringconfig_reporting_overrides:
           enabled: "{{ meteringconfig_reporting_enable_post_kube_1_14_datasources }}"
 
 # combine the _meteringconfig_tls_overrides dictionary last to enforce when spec.tls.enabled is specified and set to true
-meteringconfig_spec: "{{ meteringconfig_default_values | combine(meteringconfig_default_image_overrides, meteringconfig_storage_overrides, meteringconfig_reporting_overrides, meteringconfig_spec_overrides, _meteringconfig_root_ca_overrides, _meteringconfig_tls_overrides, recursive=True) }}"
+meteringconfig_spec: "{{ meteringconfig_default_values | combine(meteringconfig_default_image_overrides, meteringconfig_storage_overrides, _meteringconfig_tls_overrides, meteringconfig_reporting_overrides, meteringconfig_spec_overrides, _meteringconfig_root_ca_overrides, _meteringconfig_tls_overrides, recursive=True) }}"
```
chancez: _meteringconfig_tls_overrides is in here twice. The first one needs to be removed

images/metering-ansible-operator/roles/meteringconfig/defaults/main.yml
```diff
@@ -120,6 +161,28 @@ _storage_overrides:
...
+      spec:
+        config:
+          defaultFS: "wasbs://{{ _storage_spec | json_query('hive.azure.container') }}@{{ _azure_secret_account_name  }}.blob.core.windows.net{{ _storage_spec | json_query('hive.azure.rootDirectory')| default('/', True) | regex_replace('^\\/?', '/') }}"
+    hive:
+      spec:
+        config:
+          metastoreWarehouseDir: "wasbs://{{ _storage_spec | json_query('hive.azure.container') }}@{{ _azure_secret_account_name }}.blob.core.windows.net{{ _storage_spec | json_query('hive.azure.rootDirectory')| default('/', True) | regex_replace('^\\/?', '/') }}"
+
+_meteringconfig_azure_overrides: "{{ meteringconfig_storage_azure_create_secret | ternary({} , _azure_account_overides ) | default({}, true) }}"
+_storage_overrides: "{{ __storage_overrides | combine(_meteringconfig_azure_overrides, recursive = True) }}"
```
chancez: Update `__storage_overrides` to `_base_storage_overrides`

images/metering-ansible-operator/roles/meteringconfig/defaults/main.yml
```diff
@@ -120,6 +161,28 @@ _storage_overrides:
...
+    hadoop:
+      spec:
+        config:
+          defaultFS: "wasbs://{{ _storage_spec | json_query('hive.azure.container') }}@{{ _azure_secret_account_name  }}.blob.core.windows.net{{ _storage_spec | json_query('hive.azure.rootDirectory')| default('/', True) | regex_replace('^\\/?', '/') }}"
+    hive:
+      spec:
+        config:
+          metastoreWarehouseDir: "wasbs://{{ _storage_spec | json_query('hive.azure.container') }}@{{ _azure_secret_account_name }}.blob.core.windows.net{{ _storage_spec | json_query('hive.azure.rootDirectory')| default('/', True) | regex_replace('^\\/?', '/') }}"
+
+_meteringconfig_azure_overrides: "{{ meteringconfig_storage_azure_create_secret | ternary({} , _azure_account_overides ) | default({}, true) }}"
```
chancez: I don't feel like `meteringconfig_storage_azure_create_secret` is the correct condition here since it will be false when users provide an existing secret. I think you need to check if `storage.hive.azure` is set.

chancez: Ah I know. Check `meteringconfig_storage_hive_storage_type`, but thats' defined below, so you might need to re-organize a bit.

commanderchewbacca: > Ah I know. Check `meteringconfig_storage_hive_storage_type`, but that's defined below, so you might need to re-organize variable.

to my understanding I think its okay to generate a bogus value here because the hive type selector will select the correct value later on, it will only overwrite the 's3' part if its s3 for example. we don't want to select base on type because it will overwrite accountName to null if createSecret is true etc?

chancez: Let's add a comment saying:
```
# Use _azure_account_overides when meteringconfig_storage_azure_create_secret is false to correctly set the defaultFS and metastoreWarehouseDir with the storage account name from the user's secret.
```

PR #832, opened 2019-07-12
Rename the 'http' port name to 'api' in the reporting-operator service.

timflannagan1:
This can cause confusion when trying to access the reporting-operator API, i.e.:
```
curl "http://127.0.0.1:8001/api/v1/namespaces/$METERING_NAMESPACE/services/https:reporting-operator:http/proxy/api/v1/reports/get?name=namespace-cpu-utilization&namespace=$METERING_NAMESPACE&format=tab"
```

The `/services/https:reporting-operator:http` endpoint can be a bit misleading with the current port name being `http`, so this PR would change that endpoint to `/services/https:reporting-operator:api`.

For more information, check out: https://kubernetes.io/docs/tasks/administer-cluster/access-cluster-services/#manually-constructing-apiserver-proxy-urls

### DIFF
--- a/Documentation/reportqueries.md
+++ b/Documentation/reportqueries.md
@@ -144,14 +144,14 @@ This example, unlike the previous is designed to be used with Reports.
 
 You can modify the ReportQuery to display all columns or to hide or show columns as needed. The full endpoint displays all columns.
 To query report results using the reporting-operator API for full endpoint the api call is:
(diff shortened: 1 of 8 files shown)


timflannagan1:
Forgot to update the route port name to `api` (as the tests use that route by default), so the tests should be passing now.

PR #834, opened 2019-07-16
Update the Report CRD to use OpenAPI V3 schema validation.

timflannagan1:
This would add validation of the Report CRD fields using the OpenAPI V3 schema specification.

### DIFF
--- a/charts/metering-ansible-operator/templates/crds/report.crd.yaml
+++ b/charts/metering-ansible-operator/templates/crds/report.crd.yaml
@@ -35,3 +35,171 @@ spec:
   - name: Age
     type: date
     JSONPath: .metadata.creationTimestamp
+  validation:
+    openAPIV3Schema:
+      type: object
+      required:
+      - spec
+      properties:
+        # note: see https://github.com/operator-framework/operator-metering/blob/master/Documentation/reports.md for more information
+        # note: apiVersion, kind, metadata is implicitly validated by API server
+        spec:
+          type: object
+          required:
+          - query
+          properties:
+            query:
+              type: string
+              minLength: 1
+
+            reportingStart:
+              type: string
+              format: date-time
+
+            reportingEnd:
+              type: string
+              format: date-time
+
+            runImmediately:
+              type: boolean
+
+            inputs:
+              type: array
+              minItems: 1
+              items:
+                type: object
+                required:
+                - name
+                - value
+                properties:
+                  name:
+                    type: string
+                    minLength: 1
+                  value: {}
+
+            schedule:
+              type: object
+              required:
+              - period
+              properties:
+                period:
+                  type: string
+                  minLength: 1
+                  enum:
+                  - hourly
+                  - daily
+                  - weekly
+                  - monthly
+                  - cron
+
+                hourly:
+                  type: object
+                  properties:
+                    minute:
+                      type: integer
+                      minimum: 0
+                      maximum: 59
+                    hour:
+                      type: integer
+                      minimum: 0
+                      maximum: 23
+
+                daily:
+                  type: object
+                  properties:
+                    second:
+                      type: integer
+                      minimum: 0
+                      maximum: 60
+                    minute:
+                      type: integer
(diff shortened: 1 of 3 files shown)


charts/metering-ansible-operator/templates/crds/report.crd.yaml
```diff
@@ -35,3 +35,150 @@ spec:
...
+                type: object
+                required:
+                - name
+                - value
+                properties:
+                  name:
+                    type: string
+                    minLength: 1
+                  value:
+                    type: string
```
chancez: This should be a `oneOf` type I believe, the value can be any type type: eg array, object, string, double.

timflannagan1: hmm I wonder if I can just use `AnyValue: {}` instead of specifying each of the types explicitly using `oneOf`.

timflannagan1: I guess I could always just define the `value` property as `value: {}` as `AnyValue` is implicitly defined.

charts/metering-ansible-operator/templates/crds/report.crd.yaml
```diff
@@ -35,3 +35,153 @@ spec:
...
+                hourly:
+                  type: object
+                  properties:
+                    minute:
+                      type: integer
+                      minimum: 0
+                      maximum: 59
+                    hour:
+                      type: integer
+                      minLength: 0
```
chancez: SHould this be minimum?

charts/metering-ansible-operator/templates/crds/report.crd.yaml
```diff
@@ -35,3 +35,154 @@ spec:
...
     JSONPath: .metadata.creationTimestamp
+  validation:
+    openAPIV3Schema:
+      type: object
+      required:
+      - spec
+      properties:
+        # note: see https://github.com/operator-framework/operator-metering/blob/master/Documentation/reports.md for more information
+        # note: apiVersion, kind, metadata is implicitly validated by API server
+        # note: can't use $ref so we wouldn't be able to check reportingStart < reportingEnd
```
chancez: That actually isn't what `$ref` would allow you to do, and this kind of logic isn't possible that I'm aware of, but isn't worth calling out as a comment I don't think.

charts/metering-ansible-operator/templates/crds/report.crd.yaml
```diff
@@ -35,3 +35,153 @@ spec:
...
+      - spec
+      properties:
+        # note: see https://github.com/operator-framework/operator-metering/blob/master/Documentation/reports.md for more information
+        # note: apiVersion, kind, metadata is implicitly validated by API server
+        spec:
+          type: object
+          required:
+          - query
+          properties:
+            # there's no way to enforce what reportqueries get passed through the `name` field (besides the default ones using the `enum` array keyword)
```
chancez: This comment is a bit confusing to me

charts/metering-ansible-operator/templates/crds/report.crd.yaml
```diff
@@ -35,3 +35,153 @@ spec:
...
+                    hour:
+                      type: integer
+                      minimum: 0
+                      maximum: 23
+
+                monthly:
+                  type: object
+                  properties:
+                    dayOfMonth:
+                      type: string
```
chancez: This is an int between 1 and 31

charts/metering-ansible-operator/templates/crds/report.crd.yaml
```diff
@@ -35,3 +35,153 @@ spec:
...
+                    hour:
+                      type: integer
+                      minimum: 0
+                      maximum: 23
+
+                weekly:
+                  type: object
+                  properties:
+                    dayofWeek:
+                      type: string
```
chancez: This could have an enum validation for it. Checkout `convertDayOfWeek` in `pkg/operator/reports.go`

PR #835, opened 2019-07-18
Add validation for ReportQuery

EmilyM1:


charts/metering-ansible-operator/templates/crds/reportquery.crd.yaml
```diff
@@ -23,3 +23,103 @@ spec:
   - name: Age
     type: date
     JSONPath: .metadata.creationTimestamp
+
+  validation:
+  openAPIV3Schema:
+    required: ["spec"]
+    type: object
+    properties:
```
timflannagan1: I've been going with the `type: ...` -> `required` -> `properties` layout so far, and that style is what I've seen throughout most of the OLM stuff too.

charts/metering-ansible-operator/templates/crds/reportquery.crd.yaml
```diff
@@ -23,3 +23,103 @@ spec:
...
+      spec:
+        type: object
+        required: ["columns", "query"]
+        properties:
+          columns:
+            type: array
+            required: ["names"]
```
timflannagan1: where is this property?

charts/metering-ansible-operator/templates/crds/reportquery.crd.yaml
```diff
@@ -23,3 +23,103 @@ spec:
...
+                oneOf:
+                  - type: string
+                  - type: boolean
+                  - type: number
+                  - type: null
+                  - type: array
+          query:
+            type: string
+            minLength: 1
+            pattern: ".*SELECT.*"
```
timflannagan1: I'd put this section at the top of the spec `properties`

charts/metering-ansible-operator/templates/crds/reportquery.crd.yaml
```diff
@@ -23,3 +23,103 @@ spec:
...
+                    - MEMORY_BYTES_SECONDS
+                    - SECONDS
+                tableHidden:
+                  type: boolean
+          inputs:
+            type: array
+            required: ["name"]
+            items:
+              type: object
+            properties:
```
timflannagan1: you need to move the `properties` section to be inline with the `items` section as you're defining the properties of an array item, and not the contents of the `inputs` key.

timflannagan1: this section should mirror the layout of the `columns` section.

charts/metering-ansible-operator/templates/crds/reportquery.crd.yaml
```diff
@@ -23,3 +23,103 @@ spec:
...
+              name:
+                type: string
+                minLength: 1
+                type: boolean
```
timflannagan1: you have two `type` definitions (line 102, 104)

charts/metering-ansible-operator/templates/crds/reportquery.crd.yaml
```diff
@@ -23,3 +23,103 @@ spec:
...
+                enum:
+                  - STRING
+                  - INTEGER
+                  - TIME
+                  - REPORTDATASOURCE
+                  - REPORTQUERY
+                  - REPORT
+              default:
+                type: string
+                oneOf:
```
timflannagan1: you shouldn't need to define `type: string` (line 115), then specify a list of compatible types (which also includes `type: string`)

timflannagan1:
you also need to run `make metering-manifests` so the verify tests will pass.

charts/metering-ansible-operator/templates/crds/reportquery.crd.yaml
timflannagan1: Apparently, we can't do `oneOf` and include `type: ...`, see: https://kubernetes.io/blog/2019/06/20/crd-structural-schema/
```
(i) inside of value validations no additionalProperties, type, nullable, title, description
...
inside of oneOf it is not allowed to use type (rule 3-i).
```

charts/metering-ansible-operator/templates/crds/reportquery.crd.yaml
chancez: SELECT needs to be case-insensitive.

charts/metering-ansible-operator/templates/crds/reportquery.crd.yaml
timflannagan1: We would just be able to do something like this: 
```
default: {}
```
This would mean that `default` can take any value/type as the keyword `AnyValue` is implicit.

charts/metering-ansible-operator/templates/crds/reportquery.crd.yaml
```diff
@@ -23,3 +23,103 @@ spec:
...
+              type: object
+            properties:
+              name:
+                type: string
+                minLength: 1
+                type: boolean
+              type:
+                type: string
+                enum:
+                  - STRING
```
chancez: I think these should generally be lowercase/camel-case as our queries have. Currently our default ReportQueries wouldn't validate with these being all caps.

charts/metering-ansible-operator/templates/crds/reportquery.crd.yaml
```diff
@@ -23,3 +23,103 @@ spec:
...
+                    - CPU_CORE_SECONDS
+                    - CPU_CORES
+                    - MEMORY_BYTES
+                    - MEMORY_BYTES_SECONDS
+                    - SECONDS
+                tableHidden:
+                  type: boolean
+          inputs:
+            type: array
+            required: ["name"]
```
chancez: `type` is required I think

charts/metering-ansible-operator/templates/crds/reportquery.crd.yaml
```diff
@@ -23,3 +23,103 @@ spec:
...
+                    - ROW
+                    - IPADDRESS
+                    - UUID
+                    - HYPERLOGLOG
+                    - P4HYPERLOGLOG
+                    - QDIGEST
+                unit:
+                  type: string
+                  enum:
+                    - DATE
```
chancez: I believe these are all lowercase in our default ReportQueries, so the unit enum values need to be lowercase.

charts/metering-ansible-operator/templates/crds/reportquery.crd.yaml
chancez: I believe this line should be removed, and should be in `items`

charts/metering-ansible-operator/templates/crds/reportquery.crd.yaml
```diff
@@ -23,3 +23,103 @@ spec:
...
+      spec:
+        type: object
+        required: ["columns", "query"]
+        properties:
+          columns:
+            type: array
+            required: ["names"]
+            items:
+              type: object
+              required: ["type"]
```
chancez: Should be `name` and `type`

charts/metering-ansible-operator/templates/crds/reportquery.crd.yaml
```diff
@@ -23,3 +23,103 @@ spec:
   - name: Age
     type: date
     JSONPath: .metadata.creationTimestamp
+
+  validation:
+  openAPIV3Schema:
```
chancez: `openAPIV3Schema` should be indented under `validation`

charts/metering-ansible-operator/templates/crds/reportquery.crd.yaml
EmilyM1: One question about this: I have name required on line 38 and type required on 41, is this suggestion to combine them on line 38?

charts/metering-ansible-operator/templates/crds/reportquery.crd.yaml
```diff
@@ -23,3 +23,98 @@ spec:
...
     JSONPath: .metadata.creationTimestamp
+
+  validation:
+    openAPIV3Schema:
+      type: object
+      required: ["spec"]
+      properties:
+        spec:
+          type: object
+          required: ["columns", "query"]
```
timflannagan1: The current CRD's that use openAPI expand the `required` array to have a `property` per line, i.e.:
```
required:
- columns
- query
```

Minor nit, but I see there's no real convention on this (as k8s uses the `required: ["columns", ...]` format), and we already use this formatting in existing CRDs.

EmilyM1: Seems folks are using both ways, but I'll match the convention we have in there already. More cohesive that way.

charts/metering-ansible-operator/templates/crds/reportquery.crd.yaml
```diff
@@ -23,3 +23,104 @@ spec:
...
+                  type:
+                    type: string
+                    enum:
+                      - string
+                      - integer
+                      - time
+                      - ReportDataSource
+                      - ReportQuery
+                      - Report
+                  required:
```
chancez: This seems misplaced

charts/metering-ansible-operator/templates/crds/reportquery.crd.yaml
```diff
@@ -23,3 +23,104 @@ spec:
...
+                    type: boolean
```
chancez: This also looks misplaced

charts/metering-ansible-operator/templates/crds/reportquery.crd.yaml
EmilyM1: I believe this is correct for the enum, which line(s) appear off though?

charts/metering-ansible-operator/templates/crds/reportquery.crd.yaml
EmilyM1: 'required` isn't saying that a field is required, but rather the name of the field. So it's `inputs.required` rather than something that is required. Tricky.

charts/metering-ansible-operator/templates/crds/reportquery.crd.yaml
timflannagan1: this seems like the right spot to me? I thought the `ReportQuery.spec.inputs` field had a `required` property.

charts/metering-ansible-operator/templates/crds/reportquery.crd.yaml
```diff
@@ -23,3 +23,104 @@ spec:
...
+                required:
+                - name
+                - type
+                properties:
+                  name:
+                    type: string
+                    minLength: 1
+                  type:
+                    type: string
+                    enum:
```
chancez: None of our existing queries use uppercase, so we need to update those, or add lowercase support for these. Given how many queries we have, it may be easier to just relax the validation to support either all uppercase, or all lowercase than to update all our queries.

Now that we have the list of columns right here, it should be really easy to add the lowercase versions.

- copy the list, paste it below the current list (duplicating the items)
- select the newly pasted items, and lowercase the selected text
     - most editors make this easy: https://stackoverflow.com/questions/38790154/keyboard-shortcut-to-convert-selection-to-uppercase-or-lowercase-in-the-atom-e

charts/metering-ansible-operator/templates/crds/reportquery.crd.yaml
chancez: Oh, I see.

charts/metering-ansible-operator/templates/crds/reportquery.crd.yaml
chancez: Ah right

timflannagan1:
looks like the hosted zone might be fixed
/retest

timflannagan1:
Getting this error locally:
```
FAILED! => {"changed": false, "error": 422, "msg": "Failed to create object: {\"kind\":\"Status\",\"apiVersion\":\"v1\",\"metadata\":{},\"status\":\"Failure\",\"message\":\"ReportQuery.metering.openshift.io \\\"cluster-memory-capacity\\\" is invalid: []:
...
spec.columns.unit in body should be one of [date kubernetes_pod kubernetes_persistentvolumeclaim kubernetes_persistentvolume kubernetes_storageclass kubernetes_namespace bytes byte_seconds time cpu_core_seconds cpu_cores memory_bytes memory_bytes_seconds seconds
```

manifests/deploy/openshift/metering-ansible-operator/reportquery.crd.yaml
```diff
@@ -24,3 +24,131 @@ spec:
...
+                      - kubernetes_persistentvolume
+                      - kubernetes_storageclass
+                      - kubernetes_namespace
+                      - bytes
+                      - byte_seconds
+                      - time
+                      - cpu_core_seconds
+                      - cpu_cores
+                      - memory_bytes
+                      - memory_bytes_seconds
```
timflannagan1: this should be `memory_byte_seconds`

chancez:
Looks pretty good to me. Can you squash to 1 commit then I can approve the PR.

PR #838, opened 2019-07-19
Update the ReportDataSource CRD to use the OpenAPI V3 schema validation.

timflannagan1:
This would add schema validation to the `ReportDataSource` CRD using OpenAPI V3.

### DIFF
--- a/charts/metering-ansible-operator/templates/crds/reportdatasource.crd.yaml
+++ b/charts/metering-ansible-operator/templates/crds/reportdatasource.crd.yaml
@@ -39,3 +39,110 @@ spec:
   - name: Age
     type: date
     JSONPath: .metadata.creationTimestamp
+  validation:
+    openAPIV3Schema:
+      type: object
+      required:
+      - spec
+      properties:
+        spec:
+          type: object
+          properties:
+            prometheusMetricsImporter:
+              type: object
+              required:
+              - query
+              properties:
+                query:
+                  type: string
+                  minLength: 1
+                storage:
+                  type: object
+                  required:
+                  - storageLocationName
+                  properties:
+                    storageLocationName:
+                      type: string
+                      minLength: 1
+                prometheusConfig:
+                  type: object
+                  required:
+                  - url
+                  properties:
+                    url:
+                      type: string
+                      format: uri
+            reportQueryView:
+              type: object
+              required:
+              - queryName
+              properties:
+                queryName:
+                  type: string
+                  minLength: 1
+                inputs:
+                  type: array
+                  minItems: 1
+                  items:
+                    type: object
+                    required:
+                    - name
+                    - value
+                    properties:
+                      name:
+                        type: string
+                        minLength: 1
+                      value: {}
+                storage:
+                  type: object
+                  required:
+                  - storageLocationName
+                  properties:
+                    storageLocationName:
+                      type: string
+                      minLength: 1
+            awsBilling:
+              type: object
+              required:
+              - source
+              properties:
+                source:
+                  type: object
+                  required:
+                  - bucket
+                  - region
+                  properties:
+                    bucket:
+                      type: string
+                      minLength: 1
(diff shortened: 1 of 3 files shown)


charts/metering-ansible-operator/templates/crds/reportdatasource.crd.yaml
```diff
@@ -39,3 +39,118 @@ spec:
...
+                  - url
+                  properties:
+                    url:
+                      type: string
+                      format: uri
+            reportQueryView:
+              type: object
+              required:
+              - queryName
+              - inputs
```
chancez: Not required

charts/metering-ansible-operator/templates/crds/reportdatasource.crd.yaml
```diff
@@ -39,3 +39,118 @@ spec:
...
+      - spec
+      properties:
+        spec:
+          type: object
+          properties:
+            prometheusMetricsImporter:
+              type: object
+              required:
+              - query
+              # note: we could specify maxProperties: 2 or use `oneOf` to only allow two sub-schemas
```
chancez: Probably fine to leave this off for now.

charts/metering-ansible-operator/templates/crds/reportdatasource.crd.yaml
```diff
@@ -39,3 +39,118 @@ spec:
...
+                    - type
+                    properties:
+                      name:
+                        type: string
+                        minLength: 1
+                      type:
+                        type: string
+                        minLength: 1
+                        enum:
+                        - varchar
```
chancez: I don’t think this Is correct. I believe we use string? Iirc it’s based on the go types.

charts/metering-ansible-operator/templates/crds/reportdatasource.crd.yaml
```diff
@@ -39,3 +39,118 @@ spec:
...
+                  - bucket
+                  - prefix
+                  - region
+                  properties:
+                    bucket:
+                      type: string
+                      minLength: 1
+                    prefix:
+                      type: string
+                      minLength: 1
```
chancez: If prefix is unset does it validate still with minLength 1?

timflannagan1: it would - I'll have to remove that

charts/metering-ansible-operator/templates/crds/reportdatasource.crd.yaml
```diff
@@ -39,3 +39,113 @@ spec:
...
+                  items:
+                    type: object
+                    required:
+                    - name
+                    - type
+                    properties:
+                      name:
+                        type: string
+                        minLength: 1
+                      type:
```
chancez: Looking again, I see how this happened cause our documentation is confusing, but there is no `type` field for this, we need `values`, like reports instead.

charts/metering-ansible-operator/templates/crds/reportdatasource.crd.yaml
```diff
@@ -39,3 +39,113 @@ spec:
...
+                        type: string
+                        minLength: 1
+                        enum:
+                        - string
+                        - time
+                        - int
+                        - ReportDataSource
+                        - ReportQuery
+                        - Report
+                      default: {}
```
chancez: There is no `default` field here

timflannagan1:
Tests were failing due to a redundant `anyOf` sub-schema (missing a sub-schema definition for `prestoTable`), and not updating the `prestoTable.tableRef` property to be of `type: object` with a `name` field.

timflannagan1:
e2e tests are finally passing, and this should be ready to reviewed

timflannagan1:
Whoops

timflannagan1:
Looks like e2e tests are passing, so this should be ready for another review.

PR #842, opened 2019-07-19
Documentation: Add troubleshooting for the metering-ansible-operator.

timflannagan1:
This would add ways to access the Metering Ansible operator, depending on the way you installed Metering resources, and checking the status of the `MeteringConfig` CR.

### DIFF
--- a/Documentation/dev/debugging.md
+++ b/Documentation/dev/debugging.md
@@ -2,6 +2,7 @@
 
 Debugging metering can be fairly difficult if you do not know how to directly interact with the components the operator is speaking to.
 Below, we detail how you can connect and query Presto and Hive, as well as view the dashboards of the Presto and HDFS components.
+For debugging issues surrounding the metering-ansible-operator, see the [ansible-operator](#metering-ansible-operator) section.
 
 All of the follow commands assume you've set the `METERING_NAMESPACE` environment variable to the namespace your Metering installation is located in:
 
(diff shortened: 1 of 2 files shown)


Documentation/dev/debugging.md
```diff
@@ -169,3 +170,34 @@ kubectl -n $METERING_NAMESPACE port-forward hdfs-datanode-0 9864
...
+
+## Metering Ansible Operator
+Metering uses the ansible operator to watch and reconcile resources in a cluster environment. When debugging a failed Metering install, it can be helpful to view the Ansible logs or status of your `MeteringConfig` custom resource.
+
+##### Accessing Ansible Logs
+There are a couple of ways of accessing the Ansible logs depending on how you installed the Metering resources.
+
+In a typical install, the Metering operator is deployed as a pod. In this case, we can simply check the logs of the `ansible` container within this pod:
+```
+kubectl -n $METERING_NAMESPACE logs $(oc get pods -l app=metering-operator -o name | cut -c 5-) -c ansible
```
chancez: You can use `cut -d/ -f2` to use `/` as the delimiter, and select the 2nd field, which is the pod name.

Documentation/dev/debugging.md
```diff
@@ -169,3 +170,34 @@ kubectl -n $METERING_NAMESPACE port-forward hdfs-datanode-0 9864
...
+```
+docker exec -it metering-operator bash -c 'tail -n +1 -f /tmp/ansible-operator/runner/metering.openshift.io/v1/MeteringConfig/*/*/artifacts/*/stdout'
+```
+
+When tracking down a failed task, you may encounter this output:
+```
+changed: [localhost] => (item=None) => {"censored": "the output has been hidden due to the fact that 'no_log: true' was specified for this result", "changed": true}
+```
+
+This is because we specify the Ansible property, `no_log: true`, on output-extensive tasks (running helm template, creation of resources, etc.). If you fail on one of these tasks, you can set `no_log: false` on the respective task and re-run for more information on why your installation failed.
```
chancez: We actually only have a few tasks like this, and we have an option to enable logging the helm template task output now, directly from the MeteringConfig CR: `spec.logHelmTemplate=true`

timflannagan1: yeah, pretty much forgot about that field - that's a better suggestion.

chancez: It's new.

PR #847, opened 2019-07-23
charts/openshift-metering: Set bind host for hive server and metastore to 127.0.0.1 with TLS/auth enabled

chancez:
Ensures that the pod is unaccessible on it's unauthenticated ports and
all communication must be done through ghostunnel which requires TLS and
auth.


chancez:
/hold
I want to wait on this since I'd like to verify manually, but the public haven't been updated with the latest hive builds yet.

PR #848, opened 2019-07-23
Update the storage portion of the MeteringConfig CRD to use OpenAPI V3 schema validation.

timflannagan1:
There's a couple of things that still need to be specified, but the general structure matches the storage manifests and documentation.

### DIFF
--- a/charts/metering-ansible-operator/templates/crds/meteringconfig.crd.yaml
+++ b/charts/metering-ansible-operator/templates/crds/meteringconfig.crd.yaml
@@ -1483,111 +1483,111 @@ spec:
                 enableHDFS:
                   type: boolean
 
-            # storage:
-            #   type: object
-            #   required:
-            #   - hive
-            #   properties:
-            #     type:
-            #       type: string
-            #       enum:
-            #       - hive
-            #     hive:
-            #       type: object
-            #       required:
-            #       - type
-            #       properties:
-            #         type:
-            #           type: string
-            #           enum:
-            #           - hdfs
-            #           - sharedPVC
-            #           - s3
-            #           - azure
-            #         hdfs:
-            #           type: object
-            #           required:
-            #           - namenode
-            #           properties:
-            #             namenode:
-            #               type: string
-            #               format: uri
-            #               minLength: 1
-            #         s3:
-            #           type: object
-            #           required:
-            #           - bucket
-            #           - region
-            #           - secretName
-            #           properties:
-            #             bucket:
-            #               type: string
-            #               minLength: 1
-            #             region:
-            #               type: string
-            #               minLength: 1
-            #             secretName:
-            #               type: string
-            #               minLength: 1
-            #             createBucket:
-            #               type: boolean
-            #         azure:
-            #           type: object
-            #           required:
-            #           - container
-            #           - secretName
-            #           properties:
-            #             container:
-            #               type: string
-            #             secretName:
-            #               type: string
-            #             rootDirectory:
-            #               type: string
-            #         sharedPVC:
-            #           type: object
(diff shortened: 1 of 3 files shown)


charts/metering-ansible-operator/templates/crds/meteringconfig.crd.yaml
```diff
@@ -20,3 +20,163 @@ spec:
...
     storage: false
+  validation:
+    openAPIV3Schema:
+      required:
+      - spec
+      properties:
+        spec:
+          type: object
+          required:
+          - storage
```
timflannagan1: This should be the only thing that `spec` requires

charts/metering-ansible-operator/templates/crds/meteringconfig.crd.yaml
```diff
@@ -20,3 +20,163 @@ spec:
   - name: v1alpha1
     served: true
     storage: false
+  validation:
+    openAPIV3Schema:
```
timflannagan1: Due to the `status` subresource, I can't define `openAPIV3Schema` as `type: object` like the other CRD updates.

charts/metering-ansible-operator/templates/crds/meteringconfig.crd.yaml
```diff
@@ -20,3 +20,163 @@ spec:
...
+                  type: object
+                  required:
+                  - type
+                  properties:
+                    type:
+                      type: string
+                      enum:
+                      - hdfs
+                      - sharedPVC
+                      - s3
```
timflannagan1: This works fine, not sure what's up with the syntax highlighting.

charts/metering-ansible-operator/templates/crds/meteringconfig.crd.yaml
```diff
@@ -20,3 +20,163 @@ spec:
...
+                            replicas:
+                              type: integer
+                              minimum: 0
+                            storage:
+                              type: object
+                              properties:
+                                class:
+                                  type: string
+                                size:
+                                  type: string
```
timflannagan1: I could potentially add a pattern for the `size` field that checks whether or not the value is properly formatted.

charts/metering-ansible-operator/templates/crds/meteringconfig.crd.yaml
timflannagan1: I don't believe that this violates any structuring rules, as the root (`.spec`) isn't missing `type: object`, which would violate the second rule.

charts/metering-ansible-operator/templates/crds/meteringconfig.crd.yaml
```diff
@@ -20,3 +20,164 @@ spec:
   - name: v1alpha1
     served: true
     storage: false
+  preserveUnknownFields: true
```
timflannagan1: This will be needed in `apiextensions.k8s.io/v1` when pruning becomes the default behavior. We need to set this to `true` while the `MeteringConfig` CRD is still being mapped out.

timflannagan1:
I'm not sure what happened w/ that last CI-run, this is passing locally.
/retest

charts/metering-ansible-operator/templates/crds/meteringconfig.crd.yaml
timflannagan1: This was only a problem in a 3.x cluster and has been resolved by upgrading to a 4.x cluster.

timflannagan1:
This is also passing tests (now that the #851 branch has been rebasing on-top of this branch) locally, so removing hold.
/hold cancel

PR #852, opened 2019-07-26
Add 'logHelmTemplate' to MeteringConfig values.

timflannagan1:
I was getting the error:
```
ValidationError(MeteringConfig.spec): unknown field "logHelmTemplate" in io.openshift.metering.v1.MeteringConfig.spec
```
and noticed that there was no `spec.logHelmTemplate` in the `charts/openshift-metering/values.yaml`.

### DIFF
--- a/charts/openshift-metering/values.yaml
+++ b/charts/openshift-metering/values.yaml
@@ -4,6 +4,8 @@ global:
 unsupportedFeatures:
   enableHDFS: false
 
+logHelmTemplate: true
+
 storage:
   type: "hive"
   hive: {}


timflannagan1:
I'm not sure if this change is entirely needed, but we specify `spec.logHelmTemplate` in the metering-ci, and reference it in the Ansible role, so it would be nice for consistency.

chancez:
There's a few things not in values.yaml because they're not used by helm charts, only by ansible. This includes the storage section, and this field primarily. I'm not 100% sure we really want to add it to values.yaml.

timflannagan1:
Yeah that makes sense - I'll close this.

PR #858, opened 2019-07-31
Add TLS sub-schema validation to the MeteringConfig CRD.

timflannagan1:


charts/metering-ansible-operator/templates/crds/meteringconfig.crd.yaml
```diff
@@ -159,6 +159,26 @@ spec:
...
                 secretName:
                   type: string
+              oneOf:
+              - required:
+                - enabled
+                properties:
+                  certificate:
+                    maxLength: 0
+                  key:
+                    maxLength: 0
```
timflannagan1: this kind of pattern (`maxLength: 0`) is necessary as `oneOf` only matches one valid sub-schema. In the case where I specify something like this (and don't specify the `maxLength: 0`):
```
  tls:
    enabled: false
    key: stub
```
This would be considered a valid `MeteringConfig` configuration, despite it being an invalid configuration of the TLS sub-schema. Alternatively, if I configured a `MeteringConfig` like this:
```
  tls:
    enabled: false
    key: stub
    certificate: ""
```
This configuration would still be considered valid, despite the `minLength: 1` property specified under the `spec.tls.certificate` property.

charts/metering-ansible-operator/templates/crds/meteringconfig.crd.yaml
```diff
@@ -159,6 +159,28 @@ spec:
...
+              oneOf:
+              - required:
+                - enabled
+                allOf:
+                - not:
+                    required:
+                    - certificate
+                - not:
+                    required:
+                    - key
```
chancez: I believe you can put this into the other `not` with `certificate`

timflannagan1: For some reason, that doesn't enforce the behavior we want, i.e.:
```
              oneOf:
              - required:
                - enabled
                allOf:
                - not:
                    required:
                    - certificate
                    - key
```
In the `MeteringConfig` CR:
```
  tls:
    enabled: false
    certificate: stub
    # key: stub
```
And applying that file:
```
[tflannag@localhost operator-metering]$ k create -f invalid_config.yaml 
meteringconfig.metering.openshift.io/operator-metering created
[tflannag@localhost operator-metering]$ 
```

charts/metering-ansible-operator/templates/crds/meteringconfig.crd.yaml
```diff
@@ -392,23 +414,84 @@ spec:
...
+                                - secretName
+                                properties:
+                                  createSecret:
+                                    enum:
+                                    - false
+                                  secretName:
+                                    minLength: 1
+                              - required:
+                                - caCertificate
+                                - certificate
```
chancez: certificate is under `hive.auth` not `hive.tls`

charts/metering-ansible-operator/templates/crds/meteringconfig.crd.yaml
```diff
@@ -392,23 +414,84 @@ spec:
...
+                                - key
```
chancez: key is under `hive.auth` not `hive.tls`

timflannagan1:
:(
/retest

PR #861, opened 2019-07-31
Update the storage sub-schema validation to restrict mutually exclusive fields from being set.

timflannagan1:
This allows us to control what fields are set, and error when a user provides multiple `spec.storage.hive` items, i.e.:
```
kind: MeteringConfig
metadata:
  name: "operator-metering"
spec:
  unsupportedFeatures:
    enableHDFS: true

  storage:
    type: "hive"
    hive:
      type: "azure"
      hdfs:
        namenode: "hdfs-namenode-0.hdfs-namenode:9820"
      s3:
        bucket: stub
        region: stubstub
        secretName: stubstubstub
```
Before this change, this configuration would pass validation, but now this would produce the error message:
```
"spec.storage.hive" must validate one and only one schema (oneOf). Found none valid
"spec.storage.hive" must not validate the schema (not)
"spec.storage.hive" must validate all the schemas (allOf)
```

### DIFF
--- a/charts/metering-ansible-operator/templates/crds/meteringconfig.crd.yaml
+++ b/charts/metering-ansible-operator/templates/crds/meteringconfig.crd.yaml
@@ -129,24 +129,64 @@ spec:
                       type:
                         enum:
                         - hdfs
+                    allOf:
+                    - not:
+                        required:
+                        - s3
+                    - not:
+                        required:
+                        - azure
+                    - not:
+                        required:
+                        - sharedPVC
                   - required:
                     - s3
                     properties:
                       type:
                         enum:
                         - s3
+                    allOf:
+                    - not:
+                        required:
+                        - hdfs
+                    - not:
+                        required:
+                        - azure
+                    - not:
+                        required:
+                        - sharedPVC
                   - required:
                     - azure
                     properties:
                       type:
                         enum:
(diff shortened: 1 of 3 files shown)


timflannagan1:
This still needs to be updated to restrict duplicate `spec.storage.hive.type: ...` fields from being set, i.e.:
```
kind: MeteringConfig
metadata:
  name: "operator-metering"
spec:
  unsupportedFeatures:
    enableHDFS: true

  storage:
    type: "hive"
    hive:
      type: "azure"
      type: "hdfs"
      hdfs:
        namenode: "hdfs-namenode-0.hdfs-namenode:9820"
```
The configuration above would pass validation, but I haven't found a way to enforce this behavior yet.

chancez:
@timflannagan1 The above is more of a yaml parsing thing. Specifying `type` twice, the yaml parser simply overwrites the values of the first with `"hdfs"`, so the validation only ever sees `type: "hdfs"`

timflannagan1:
@chancez okay that makes sense, and this would imply that the inverse (`type: hdfs`, `type: azure`) would be invalid, which would be the intended behavior.

chancez:
Yeah, that would basically be setting: type: azure, the hdfs: {} section  which would be an invalid combo

PR #863, opened 2019-08-01
Updates docs: oc create project-> new-project

bentito:
`oc new-project metering` works.


pruan-rht:
probably should use `oc create ns $METERING_NAMESPACE` because oc new-project has some restrictions such as not allowing project name to begin with `openshift-`

chancez:
Ah, yes that's true.

PR #865, opened 2019-08-01
Add properties and update validation to spec.hive.storage in the MeteringConfig CRD.

timflannagan1:
This PR adds the `spec.storage.hive.sharedPVC.mountpath` property, which was undefined previously. 

I also needed to update the `spec.storage.hive.azure` properties to include variables used in the ansible-operator (for developmental purpose), which weren't specified in the documentation.

### DIFF
--- a/charts/metering-ansible-operator/templates/crds/meteringconfig.crd.yaml
+++ b/charts/metering-ansible-operator/templates/crds/meteringconfig.crd.yaml
@@ -94,23 +94,40 @@ spec:
                       - container
                       - secretName
                       properties:
+                        createSecret:
+                          type: boolean
                         container:
                           type: string
+                          minLength: 1
                         secretName:
                           type: string
+                          minLength: 1
+                        storageAccountName:
+                          type: string
+                          minLength: 1
+                        secretAccessKey:
+                          type: string
+                          minLength: 1
                         rootDirectory:
                           type: string
+                          minLength: 1
                     sharedPVC:
                       type: object
                       properties:
-                        claimName:
-                          type: string
-                        storageClass:
-                          type: string
-                        size:
-                          type: string
                         createPVC:
                           type: boolean
+                        claimName:
+                          type: string
+                          minLength: 1
+                        storageClass:
+                          type: string
+                          minLength: 1
+                        size:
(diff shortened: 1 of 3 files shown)


charts/metering-ansible-operator/templates/crds/meteringconfig.crd.yaml
```diff
@@ -94,23 +94,40 @@ spec:
...
+                          minLength: 1
                         storageClass:
                           type: string
+                          minLength: 1
                         size:
                           type: string
-                        createPVC:
-                          type: boolean
+                          minLength: 1
+                        mountPath:
```
timflannagan1: this was missing from the initial storage PR.

charts/metering-ansible-operator/templates/crds/meteringconfig.crd.yaml
```diff
@@ -94,23 +94,40 @@ spec:
...
+                          minLength: 1
+                        storageAccountName:
+                          type: string
+                          minLength: 1
+                        secretAccessKey:
+                          type: string
+                          minLength: 1
                         rootDirectory:
                           type: string
+                          minLength: 1
```
timflannagan1: specifying `minLength: 1` on a non-required property, just ensures that the `spec.storage.hive.azure.rootDirectory` key is non-empty, when it's specified in a `MeteringConfig` CR.

charts/metering-ansible-operator/templates/crds/meteringconfig.crd.yaml
```diff
@@ -94,23 +94,40 @@ spec:
...
+                          type: boolean
                         claimName:
                           type: string
+                          minLength: 1
                         storageClass:
                           type: string
+                          minLength: 1
                         size:
                           type: string
-                        createPVC:
```
timflannagan1: just moving this property to match the convention of other keys throughout the CRD.

charts/metering-ansible-operator/templates/crds/meteringconfig.crd.yaml
chancez: Yep, awesome

PR #867, opened 2019-08-01
Documentation: Use namespace instead of projects for OLM install

chancez:
Projects have restrictions on the names, so create a namespace instead,
which also creates a matching project.

### DIFF
--- a/Documentation/olm-install.md
+++ b/Documentation/olm-install.md
@@ -11,17 +11,17 @@ This procedure covers:
 
 Create a dedicated OpenShift project for Metering, and then install the Metering Operator:
 
-1. Create a new project/namespace called *metering* using the OpenShift web console (navigate to **Project > Create Project**) or the `oc` command:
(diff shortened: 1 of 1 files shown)


Documentation/olm-install.md
```diff
@@ -11,17 +11,17 @@ This procedure covers:
...
-1. Create a new project/namespace called *metering* using the OpenShift web console (navigate to **Project > Create Project**) or the `oc` command:
+1. Create a new project/namespace called *openshift-metering* using the OpenShift web console (navigate to **Administration > Namespaces > Create Namespace**) or the `oc` command:
 
 ```
-oc new-project metering
+oc create namespace openshift-metering
 ```
 
-2. From the web console, click **Catalog > OperatorHub**, and search for *metering* to find the Metering Operator.
+2. From the web console, click **Catalog > OperatorHub**, and search for *openshift-metering* to find the Metering Operator.
```
pruan-rht: search should  just use `metering`,  using `openshift-metering` would result in **No Results Match the Filter Criteria**, this is probably a global search and replace error. :)

chancez: Oh lol yah

PR #871, opened 2019-08-02
adds doc note about hdfs-storage

bentito:
Default hack/install.sh fails and currently doc path to understanding how to fix is:
Follow link to Documentation/common-configuration.md > Documentation/configuring-storage.md > Documentation/configuring-hdfs.md and then sort of infer that the path partial is what's needed for that env variable.

This change adds a note with the solution line in the document.

### DIFF
--- a/images/metering-ansible-operator/roles/meteringconfig/tasks/configure_storage.yml
+++ b/images/metering-ansible-operator/roles/meteringconfig/tasks/configure_storage.yml
@@ -46,6 +46,11 @@
         namespace: "{{ meta.namespace }}"
       register: azure_secret_buf
 
+    - name: Verify azure storage credentials secret exists
+      assert:
+        that:
+          - azure_secret_buf is defined and azure_secret_buf.resources | length > 0
+          - azure_secret_buf.resources[0].data['azure-storage-account-name'] != ""
     - name: Extract Azure storage account name
       set_fact:
         _azure_secret_account_name: "{{ azure_secret_buf.resources[0].data['azure-storage-account-name'] | b64decode | default(omit)  }}"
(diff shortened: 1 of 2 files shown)


Documentation/manual-install.md
```diff
@@ -38,6 +38,8 @@ $ export METERING_NAMESPACE=metering-$USER
 $ ./hack/openshift-install.sh
 ```
 
+[Note: the default has no storage setup so you may want to choose HDFS, do so with: ```export METERING_CR_FILE=manifests/metering-config/hdfs-storage.yaml```, you *must* choose some storage option for a successful install.]
```
timflannagan1: This renders weird - might want to just remove the starting and ending brackets, and replace with `**Note**`

timflannagan1: the default `MeteringConfig` CR has no storage ...

bentito: 👍 changed in fe2da36 and 7a63280

Documentation/manual-install.md
```diff
@@ -38,6 +38,8 @@ $ export METERING_NAMESPACE=metering-$USER
 $ ./hack/openshift-install.sh
 ```
 
+**Note** The default `MeteringConfig` CR has no storage setup so you may want to choose HDFS, do so with: ```export METERING_CR_FILE=manifests/metering-config/hdfs-storage.yaml```, you *must* choose some storage option for a successful install.
```
chancez: Two things:
- We do not want to recommend HDFS to users.
- This is covered under `## Customize installation`. It might make sense to move that section up now that the minimal configuration requires storage, and update to to mention the configuring storage documentation.

PR #872, opened 2019-08-02
validate fsGroup as integer

mmariani:
With the new validation schema, fsGroup in metering-custom.yaml cannot be an integer (the schema requires a string) nor a string (fails while parsing \ufffd as an integer, if I remember correctly), nor a map (like the k8s docs suggest).

After reading the sources, I am still not sure how it's supposed to work, or if this is due to a bug or work in progress. But reverting the schema to integer fixes the issue for me.

### DIFF
--- a/charts/metering-ansible-operator/templates/crds/meteringconfig.crd.yaml
+++ b/charts/metering-ansible-operator/templates/crds/meteringconfig.crd.yaml
@@ -1348,7 +1348,7 @@ spec:
                       type: object
                       properties:
                         fsGroup:
-                          type: string
+                          type: integer
                         runAsGroup:
                           type: integer
                           format: int64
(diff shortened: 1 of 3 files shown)


timflannagan1:
@mmariani yeah, under further investigation, this should be `type: integer` and was an oversite on my part when I was adding all the `MeteringConfig` schema validation.

charts/metering-ansible-operator/templates/crds/meteringconfig.crd.yaml
```diff
@@ -1348,7 +1348,7 @@ spec:
                       type: object
                       properties:
                         fsGroup:
-                          type: string
+                          type: integer
```
timflannagan1: while you're at it, can you also add the `format: int64` under the `type: integer` for the `fsGroup` properties?

chancez: @timflannagan1 Let's merge this to unbreak things, and can you add the format as a follow up PR?

PR #874, opened 2019-08-02
specifies hdfs opt in manual install

bentito:
Adds specific configuration line for using hdfs storage.

### DIFF
--- a/images/metering-ansible-operator/roles/meteringconfig/tasks/configure_storage.yml
+++ b/images/metering-ansible-operator/roles/meteringconfig/tasks/configure_storage.yml
@@ -46,6 +46,11 @@
         namespace: "{{ meta.namespace }}"
       register: azure_secret_buf
 
+    - name: Verify azure storage credentials secret exists
+      assert:
+        that:
+          - azure_secret_buf is defined and azure_secret_buf.resources | length > 0
+          - azure_secret_buf.resources[0].data['azure-storage-account-name'] != ""
     - name: Extract Azure storage account name
       set_fact:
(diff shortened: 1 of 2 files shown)


Documentation/manual-install.md
```diff
@@ -78,7 +78,10 @@ image tag of each component.
 $ export METERING_NAMESPACE=metering-$USER
 $ export METERING_CR_FILE=metering-custom.yaml
 ```
-
+**Note:** You must configure some storage, for instance hdfs as an option with no external requirements, with:
```
chancez: Can we just link to `configuring-storage.md` instead?

bentito: right but then someone (me) could fall into that trap again, because `configuring-storage` links to `configuring-hdfs` which isn't explicit about the thing that actually got it working for me and clued me in about how the env variables were being set to reference the YAML: `export METERING_CR_FILE=manifests/metering-config/hdfs-storage.yaml

So perhaps change to linking `configuring-storage.md` and add that bit to `configuring-hdfs.md`?

chancez: That's why I suggested (in the previous PR) moving this entire customizing section above the current install instructions at the top. It covers setting `METERING_CR_FILE`. Then we can mention you need to configure storage, and link to configuring storage. 

We can't mention `METERING_CR_FILE` in storage docs because that's specific to the manual install. We should just be more explicit in the manual install docs that you must configure storage in your $METERING_CR_FILE before install.

bentito: done in 94c5f5e

PR #878, opened 2019-08-03
Adds jq -r for correct metering host value

bentito:
Without the `-r` flag to `jq`, command as written in the docs to get the `METERING_ROUTE_HOSTNAME` was incorrect -- was surrounded by double quotes.


timflannagan1:
We also mentioned this in [configuring reporting-operator](https://github.com/operator-framework/operator-metering/blob/master/Documentation/configuring-reporting-operator.md#openshift-authentication)

PR #880, opened 2019-08-05
charts,manifests: Update validation to the reporting-operator authProxy section.

timflannagan1:


charts/metering-ansible-operator/templates/crds/meteringconfig.crd.yaml
```diff
@@ -367,55 +367,147 @@ spec:
...
                               type: boolean
                             data:
                               type: string
-                            enabled:
-                              type: boolean
+                              minLength: 1
                             secretName:
                               type: string
+                              minLength: 1
+                          oneOf:
```
timflannagan1: This would allow you to 
```
reporting-operator:
  spec:
    authProxy:
      authenticatedEmails:
        enabled: false
```
or:
```
reporting-operator:
  spec:
    authProxy:
      authenticatedEmails:
        enabled: true
        createSecret: false
        secretName: existing-secret-name
```
or: 
```
reporting-operator:
  spec:
    authProxy:
      authenticatedEmails:
        enabled: true
        createSecret: true
        data: non-empty-data
        # secretName is optional
        secretName: new-secret-name
```

charts/metering-ansible-operator/templates/crds/meteringconfig.crd.yaml
```diff
@@ -367,55 +367,146 @@ spec:
...
                           properties:
                             createSecret:
                               type: boolean
                             secretName:
                               type: string
+                              minLength: 1
                             seed:
                               type: string
+                              minLength: 32
+                          oneOf:
```
timflannagan1: This would enforce the following configurations (but no way to enforce that `tls.enabled: false` is set:
```
reporting-operator:
  spec:
    authProxy:
      cookie:
        createSecret: true
        seed: non-empty-seed
        # secretName is optional
```
or:
```
reporting-operator:
  spec:
    authProxy:
      cookie:
        createSecret: false
        secretName: existing-secret-name
```

timflannagan1: In the case where `spec.tls.enabled` is omitted, or uses the default value, `true`, then these values would be overridden by the ansible-operator.

charts/metering-ansible-operator/templates/crds/meteringconfig.crd.yaml
timflannagan1: Note: `authProxy.authenticatedEmails` is not part of the `_tls_overrides` variables.

PR #886, opened 2019-08-06
Add OpenAPI V3 schema validation to the PrestoTable CRD.

timflannagan1:
This would add schema validation to the `PrestoTable` CRD.

To-Do:
- [x] Add a `description` field to each property
- [x] Test different valid configurations
- [x] Ensure that the `spec.anyOf` correctly identifies valid configurations.

This is the output of `oc explain prestotables` after adding the top-level description fields:
```
[tflannag@localhost operator-metering]$ oc explain prestotables
KIND:     PrestoTable
VERSION:  metering.openshift.io/v1

DESCRIPTION:
     PrestoTable is a custom resource that represents a database table
     accessible from within Presto. When a PrestoTable resource is created, the
     reporting-operator creates a table within Presto according to the
     configuration provided, or exposes an existing table. More info:
     https://prestosql.io/docs/current/index.html

FIELDS:
   apiVersion	<string>
     APIVersion defines the versioned schema of this representation of an
     object. Servers should convert recognized schemas to the latest internal
     value, and may reject unrecognized values. More info:
     https://git.k8s.io/community/contributors/devel/api-conventions.md#resources

   kind	<string>
     Kind is a string value representing the REST resource this object
     represents. Servers may infer this from the endpoint the client submits
     requests to. Cannot be updated. In CamelCase. More info:
     https://git.k8s.io/community/contributors/devel/api-conventions.md#types-kinds

   metadata	<Object>
     Standard object's metadata. More info:
     https://git.k8s.io/community/contributors/devel/api-conventions.md#metadata

   spec	<Object> -required-
     PrestoTableSpec is the desired specification of a PrestoTable custom
     resource. Required fields: unmanaged, catalog, schema, tableName, and
     columns. More info:
     https://github.com/operator-framework/operator-metering/blob/master/Documentation/prestotables.md

[tflannag@localhost operator-metering]$ 
```

### DIFF
--- a/Documentation/prestotables.md
+++ b/Documentation/prestotables.md
@@ -2,25 +2,27 @@
 
 A `PrestoTable` is a custom resource that represents a database table accessible from within Presto.
 
-When created, a PrestoTable resource causes the reporting-operator to create a table within Presto according to the configuration provided or can expose an existing table that already exists to the reporting-operator.
+When a `PrestoTable` resource is created, the reporting-operator creates a table within Presto according to the configuration provided, or can expose an existing table that already exists to the reporting-operator.
 
 Please read the [Presto concepts][presto-concepts] documentation to gain an understanding of any Presto specific terminology used below.
 
 ## Fields
-
-- `unmanaged`: If true this indicates the PrestoTable resource is referencing an existing table, and should not attempt to create or manage the table within Presto.
-- `catalog`: The catalog within Presto the table to be created in, or the catalog the table should exist within if unmanaged. In many cases, this will be `hive`.
-- `schema`: The schema within the Presto catalog for the table to created in, or the schema the table should exist in if unmanaged. If the catalog is `hive` then there will always be at least the `default` schema.
-- `tableName`: The name of the table to create in Presto, or the name of an existing table if unmanaged.
-- `columns`: A list of columns that match the schema of the of the PrestoTable.
-- `properties`: Optional: A map containing string keys and values, where each key value pair is a table property for configuring the table. The available properties depends on the `catalog` used, see the Presto documentation for information on available properties.
-- `comment`: Optional: If specified, sets a comment on the table. Comments are just arbitrary strings that have no meaning to Presto, but can be used to store arbitrary information about a table.
-- `view`: Optional: If true, then `query` must also be set. When `view` is true it causes reporting-operator to create a view within Presto using `query` as the SELECT statement for creating the view. `columns` must still be set in order for other Resources to determine the correct schema for the table, despite the schema being determined from the query used.
-- `createTableAs`: Optional: If true, then `query` must also be set. When `createTableAs` is true it causes reporting-operator to create a table within Presto using `query` as the SELECT statement for creating the table, using the schema of the query and the results as the content of the table. `columns` must still be set in order for other Resources to determine the correct schema for the table, despite the schema being determined from the query used.
-- `query`: Optional: If non-empty, you must set `view` or `createTableAs` to true. Must be a SELECT query to be used for creating the table or view.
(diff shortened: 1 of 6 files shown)


charts/metering-ansible-operator/templates/crds/prestotable.crd.yaml
```diff
@@ -21,3 +21,105 @@ spec:
   - name: Table Name
     type: string
     JSONPath: .status.tableName
+  validation:
+    openAPIV3Schema:
+      description: 'PrestoTable is a custom resource that represents a database table accessible from within Presto.
```
chancez: Use `: |` and remove the quotes, and indent this at the next level on a newline for a multiline string. https://yaml-multiline.info/

chancez: Do this for all descriptions.

charts/metering-ansible-operator/templates/crds/prestotable.crd.yaml
```diff
@@ -21,3 +21,132 @@ spec:
...
+              minLength: 1
+          oneOf:
+          - required:
+            - createTableAs
+            - query
+            properties:
+              createTableAs:
+                enum:
+                - true
+              unmanaged:
```
EmilyM1: I see a few cases of enums with one value. Is enum just for one item?

timflannagan1: It's just restricting the possible values for the `createTableAs` property in this sub-schema. If I only want `createTableAs: true` to be the only valid value, then I would only specify a single value for `enum`.

See: https://swagger.io/docs/specification/data-models/enums/

Documentation/prestotables.md
```diff
@@ -2,25 +2,62 @@
...
+##### Optional Fields:
+- `query`: The SELECT [query](https://prestosql.io/docs/current/overview/concepts.html#query) used for creating the table or view. If `query` is non-empty, you must set either `view` or `createTableAs` to true. Continue onto the next section for [examples](#example-prestotables) that use the `query` field.
+- `view`: Controls whether the reporting-operator needs to create a view within Presto. If true, the reporting-operator uses the `query` field as the SELECT statement used to create the table, using both the schemas of the query and results for the content of the table.
+- `createTableAs`: Controls whether the reporting-operator needs to create a table within Presto using the `query` fields as the SELECT statement for creating the table.
+- `properties`: A map containing string keys and values, where each key-value pair is a table property for configuring the table. The available properties depend on the `catalog` specified. See the Presto documentation for information on available properties.
```
timflannagan1: @chancez  "See the Presto documentation for information on available properties."
I wasn't able to find this when I ran a quick google search.

chancez: It depends on the catalog, so it's per connector. We could potentially just say something to that end. "See the Presto connector documentation to find available properties for your connector's catalog/."

timflannagan1: :+1:

Documentation/prestotables.md
```diff
@@ -2,25 +2,62 @@
...
+      "timestamp",
+      dt,
+      sum(node_capacity_cpu_cores) as cpu_cores,
+      sum(node_capacity_cpu_core_seconds) as cpu_core_seconds,
+      count(*) AS node_count
+    FROM hive.metering.example_node_cpu_capacity_view
+    GROUP BY "timestamp", dt
+```
 
+##### Creating a table in Presto using a `SELECT` query:
```
chancez: Can this go above the other example?

chancez:
I like your view example 👍 
/lgtm
/approve

PR #887, opened 2019-08-08
Add OpenAPIV3 schema validation to the HiveTable CRD.

timflannagan1:
## Overview
This would add schema validation to the `HiveTable` CRD. I verified that you need at the minimum, the `tableName`, `columns`, and `databaseName` for a `HiveTable` resource to turn into a valid table in Presto.

## To-Do:
- [x] Add `oneOf` to top-level `spec.properties` to enforce certain configurations.
- [x] Update `hivetables.md` to clarify what fields are required, and optional.
- [x] Update `hivetables.md` and add more examples.

### DIFF
--- a/charts/metering-ansible-operator/templates/crds/hive.crd.yaml
+++ b/charts/metering-ansible-operator/templates/crds/hive.crd.yaml
@@ -21,3 +21,253 @@ spec:
   - name: Table Name
     type: string
     JSONPath: .status.tableName
+  validation:
+    openAPIV3Schema:
+      description: |
+        HiveTable is a custom resource that represents a database table within Hive.
+        When created, a HiveTable resource causes the reporting-operator to create a
+        table witin Hive according to the configuration provided.
+      required:
+      - spec
+      properties:
+        spec:
+          type: object
+          description: |
+            HiveTableSpec is the desired specification of a HiveTable custom resource.
+            Required fields: database, tableName, and columns.
+            More info: https://github.com/operator-framework/operator-metering/blob/master/Documentation/hivetables.md
+          required:
+          - databaseName
+          - tableName
+          - columns
+          properties:
+            databaseName:
+              type: string
+              description: |
+                DatabaseName is the name of the Hive database to use.
+                Generally, this field should be set to "default", or the value of the "databaseName" in a Hive StorageLocation.
+                More info: https://github.com/operator-framework/operator-metering/blob/master/Documentation/storagelocations.md
+              minLength: 1
+            tableName:
+              type: string
+              description: |
+                TableName is the desired name of the table to be created in Hive.
+              minLength: 1
+            columns:
+              type: array
+              description: |
+                A list of columns that match the schema of the HiveTable.
+                For each list item, you must specify a `name` field, which is the name of an individual column for the Hive table,
+                and a `type` field, which corresponds to a valid type in Hive.
+                Note: the only complex types supported are map's of primitive types.
+                More info: https://cwiki.apache.org/confluence/display/Hive/LanguageManual+Types
+              items:
+                type: object
+                required:
+                - name
+                - type
+                properties:
+                  name:
+                    type: string
+                    description: |
+                      The name of the column.
+                    minLength: 1
+                  type:
+                    type: string
+                    description: |
+                      The column data type.
+                    minLength: 1
+            partitionedBy:
+              type: array
+              description: |
+                A list of columns that are used as partition columns.
+                Columns in "partitionedBy" and "columns" must not overlap.
+                For each list item, you must specify both a `name` and `type` field.
+                Note: this is an optional field.
+              minItems: 1
+              items:
+                type: object
+                required:
+                - name
+                - type
+                properties:
+                  name:
+                    type: string
+                    description: |
+                      The name of the column.
+                    minLength: 1
+                  type:
+                    type: string
+                    description: |
+                      The column data type.
+                    minLength: 1
+            clusteredBy:
+              type: array
+              description: |
+                A list of columns from "columns" to use for bucketed tables.
+                This field must be set if "numBuckets" is specified.
+                Note: this is an optional field.
+                More info: https://cwiki.apache.org/confluence/display/Hive/LanguageManual+DDL+BucketedTables
+              minItems: 1
+              items:
+                type: string
+            sortedBy:
+              type: array
+              description: |
+                A list of column names from "columns" to use for bucketed tables.
+                This field must be set if "clusteredBy" and "numBuckets" are specified.
+                Note: this is an optional field.
+                More info: https://cwiki.apache.org/confluence/display/Hive/LanguageManual+DDL+BucketedTables
+              required:
+              - name
+              items:
+                type: object
+                properties:
+                  name:
+                    type: string
+                    description: |
+                      The name of the column from "columns".
+                    minLength: 1
+                  descending:
+                    type: boolean
+                    description: |
+                      Descending controls whether the column is descending or ascending.
+                      If "descending" is true, then the column is descending, else ascending.
+                      If this field is unspecified, then it defaults to Hive's default behavior.
+                      Note: this is an optional field.
+            numBuckets:
+              type: integer
+              description: |
+                The number of buckets to create for a bucketed table.
+                If this field is set, then "clusteredBy" also needs to be set.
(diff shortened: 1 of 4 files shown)


timflannagan1:
For the time being, I added `nullable: true` to the `properties` array so that tests can pass w/ just the structural schema defined.

charts/metering-ansible-operator/templates/crds/hive.crd.yaml
```diff
@@ -185,6 +185,7 @@ spec:
               description: |
                 A list of partitions that this Hive table should contain.
                 Note: this is an optional field.
+              nullable: true
```
timflannagan1: For the time being, I added `nullable: true` to the `properties` array so that tests can pass w/ just the structural schema defined.

This wouldn't violate any structural schema rules.

charts/metering-ansible-operator/templates/crds/hive.crd.yaml
```diff
@@ -21,3 +21,216 @@ spec:
...
+                enum:
+                - true
+          - allOf:
+            - not:
+                required:
+                - numBuckets
+            - not:
+                required:
+                - clusteredBy
+            properties:
```
timflannagan1: Note: it's redundant to specify something like:
```yaml
properties:
  partitions:
    type: null
```
as we already attributed the `nullable: true` to the `partitions` array property.

timflannagan1: See: https://github.com/OAI/OpenAPI-Specification/issues/1900

charts/metering-ansible-operator/templates/crds/hive.crd.yaml
```diff
@@ -21,3 +21,216 @@ spec:
...
+                enum:
+                - true
+          - required:
+            - managePartitions
+            - partitions
+            properties:
+              managePartitions:
+                enum:
+                - true
+          - allOf:
```
chancez: This case doesn't make much sense to me, managePartitions has nothing to do with `numBuckets` or `clusteredBy`. 

I think if you're trying to handle numBuckets + clusteredBy being dependent on each other, you could have something like (updating the top case in this oneOf and removing this allOf):
```
- oneOf:
   - required:
      - numBuckets
      - clusteredBy
   - not:
        required:
        - numBuckets
        - clusteredBy
```

timflannagan1: I needed a case that allows the minimal configuration (`columns`, `tableName`, and `databaseName`) as a valid configuration, but also allows other fields like `external`, `rowFormat`, `fileFormat`, etc.

I wasn't sure what the best approach was to do something like this, while still listing fields that are dependent on each other.

timflannagan1: Also, any `HiveTable` resource that's implicitly created will have a `managePartitions: false` and `partitions: null` included in the spec (like in the case where a `PrometheusMetricsImporter` ReportDatasource.)

chancez: @timflannagan1 then why not add a oneOf case that has `required` for columns, tableName, and databaseName then?

timflannagan1: I was thinking about that initially, but it's redundant to specify both those fields as required in the property definitions, and then include them as required fields in the `oneOf` clause.

chancez: Can we have an empty object or something in the oneOf? or empty required?

timflannagan1: That could work, assuming that an invalid configuration would produce two valid alternatives, and the `oneOf` clause would error, as it only allows a single valid configuration.

chancez: We can use `anyOf` then

chancez: Or a combination of nesting an anyOf within the oneOf

timflannagan1: We would need to use `oneOf` as the top-level clause. I can experiment w/ different combinations of nested logical junctor's.

charts/metering-ansible-operator/templates/crds/hive.crd.yaml
```diff
@@ -21,3 +21,225 @@ spec:
...
+                A list of columns from "columns" to use for bucketed tables.
+                This field must be set if "numBuckets" is specified.
+                Note: this is an optional field.
+                More info: https://cwiki.apache.org/confluence/display/Hive/LanguageManual+DDL+BucketedTables
+              items:
+                type: object
+                required:
+                - name
+                - type
+                properties:
```
timflannagan1: @chancez there wasn't much info on this field in the documentation, besides that you need to list the columns to use for bucketed fields. Is this just an array of column names, or does it emulate the `columns` property (where it's an array and each array element is an object.)

timflannagan1: It looks like it's the fomer: `ClusteredBy   []string     `json:"clusteredBy,omitempty"``

chancez: https://github.com/operator-framework/operator-metering/blob/master/pkg/apis/metering/v1/hive_table.go#L39

charts/metering-ansible-operator/templates/crds/hive.crd.yaml
```diff
@@ -209,3 +209,41 @@ spec:
...
+            - managePartitions
+            - partitions
+            properties:
+              managePartitions:
+                enum:
+                - true
+              partitions:
+                type: array
+          # note: allow the minimal configuration, and enforce the schemas defined above
+          - properties:
```
timflannagan1: @chancez I wasn't able to think of anything that still enforced that the dependent properties were set together w/o listing the negation of those properties. While this `oneOf` case isn't very intuitive, it does enforce the sub-schemas listed above, while allowing the minimal configurations still.

Another thing to note is that [this would violate a structural schema](https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/#status-subresource)(as it includes `maxItems` outside of the root), but every CRD that's been updated recently also does this.

timflannagan1:
The `oneOf` clause isn't handling all invalid configurations.
/hold

charts/metering-ansible-operator/templates/crds/hive.crd.yaml
chancez: Why use maxItems when we can just do `not: required: - clusteredBy`?

charts/metering-ansible-operator/templates/crds/hive.crd.yaml
```diff
@@ -21,3 +21,253 @@ spec:
...
+                enum:
+                - true
+              partitions:
+                type: array
+              external:
+                enum:
+                - true
+          # note: allow the minimal configuration, and enforce the schemas defined above
+          - allOf:
+            - not:
```
chancez: Maybe im confused but doesn't this prevent setting these fields at all?

timflannagan1: Yes, this would just allow two configurations, the one where only `tableName`, `databaseName`, and `columns` are provided (the minimal configuration) and also the one that the `PrometheusMetricsImporter` provides

chancez: Oh, I'd rather we go to the oneOf validating the main requirements than preventing valid configs from being created. That wasn't super clear that supporting the "main" meant "not supporting advanced". I'd rather simply not validate more advanced options than not allow them at all.

timflannagan1: Unless you mean the nested `allOf` clause preventing the setting of these properties within the context of the overall `anyOf`, which isn't the case. 

This just acts as the default case for the configurations described above.

chancez: Oh i totally missed this was nested! Shoot.

timflannagan1:
the prow jobs have been frozen in pending for 15 hours
/retest

PR #888, opened 2019-08-08
charts,manifests: Remove the required properties in the Report CRD.

timflannagan1:
This was invaliding reports that didn't provide a schedule like hourly or daily.


chancez:
tested in our cluster, and since this is a decent regression im going to merge it now to avoid breakage for existing reports.
/lgtm
/approve

PR #889, opened 2019-08-08
fix for azure in meteringconfig.crd 

commanderchewbacca:
added different configs for azure in the meteringconfig.crd to allow creating secrets for testing


charts/metering-ansible-operator/templates/crds/meteringconfig.crd.yaml
```diff
@@ -91,8 +91,24 @@ spec:
                     azure:
                       type: object
                       required:
-                      - container
-                      - secretName
+                        - container
```
chancez: Can you match existing indentation and align this with `required`?

PR #890, opened 2019-08-08
Update OCP metering-ansible-operator Root CA CSR creation

chancez:
- Use python2-cryptography because openssl_csr requires a newer version of pyOpenSSL than available in the OCP rpm repos.
- Change the generated metering Root CA common name to a hostname to fix issues caused by using python cryptography package instead of pyOpenSSL.


chancez:
I might have found a way to avoid changing the root CA common name.

PR #898, opened 2019-08-14
Dockerfile.src: Run yum clean and remove yum cache before running yum

chancez:
Without this, we sometimes get errors updating metadata such as:

http://mirror.math.princeton.edu/pub/epel/7/x86_64/repodata/59cd2d904711571ac63cfec0fec0641233dc027173c6667914bbcc2e10ea11dd-primary.sqlite.bz2: [Errno 14] HTTP Error 404 - Not Found


chancez:
CI doesn't test changes from the PR HEAD for the src image so we have to bypass CI to fix this.

PR #900, opened 2019-08-14
Add the capability to disable OCP-only features.

timflannagan1:
This would allow us to support upstream kubernetes' users and not just Openshift users.

What this PR changes:
1. Adds a set of override variables to the metering-ansible-operator that allows a user to toggle between OCP and non-OCP features by setting the `DISABLE_OCP_FEATURES="true"` environment variable, or `spec.disableOCPFeatures.enabled: true` in their `MeteringConfig` CRD. These override variables contain things that are invalid configurations for non-OKD distributions, most notably `fsGroup`, which needs to be set to a zero value, and Openshift-specific features like a route, or service-serving certificates. The default behavior is still geared towards Openshift, so this value is set to false and needs to be explicitly set in order to override the OCP-only features.
2. Alters the `hack/*` scripts that set up the installation, and create the various metering manifests. For starters, I added another set of manifests to `./hack/generate-metering-manifests.sh`, which creates another set of upstream manifests located in `./manifests/deploy/upstream/*`. I also updated `./hack/lib/init.sh`, adding another set of `UPSTREAM_*` variables that are used throughout the installation scripts. The `OLM_MANIFESTS_DIR`, and `INSTALLATION_MANIFESTS_DIR` are now set to the default manifests (openshift), and all openshift related manifests were given the prefix of `OCP_*`. I added the following upstream installation files: `./hack/upstream-install.sh` and `./hack/upstream-uninstall.sh`, which override the `INSTALLATION_MANIFESTS_DIR` to point to the upstream manifests directory location, and then call their respective `./hack/install.sh` and `./hack/uninstall.sh` scripts. On the OLM front, I added `./hack/upstream-olm-install.sh`, and `./hack/upstream-olm-uninstall.sh`, and like the upstream install/uninstall scripts added, just override the `OLM_MANIFESTS_DIR` to point to the upstream manifests directory, and then call their respective `./hack/olm-install.sh`, and `./hack/olm-uninstall.sh`. 
3. Created and populated a set of upstream manifests located in `./manifests/deploy/upstream/*`.
4. Added a `charts/metering-ansible-operator/upstream-values.yaml` that controls the upstream values for the metering-ansible-operator.
5. Updated the `charts/openshift-metering/values.yaml` and the respective CRD manifests to include a top-level `spec.disableOCPFeatures` key, which controls the toggle between OCP and non-OCP features.
6. Remove a default value for the `spec.reporting-operator.spec.config.prometheus.url` to make sure that this value is set for non-OCP users, and is given a default value for OCP users in the role execution in the metering-ansible-operator.


charts/metering-ansible-operator/templates/_helpers.tpl
```diff
@@ -40,6 +40,8 @@ template:
       env:
       - name: OPERATOR_NAME
         value: "metering-ansible-operator"
+      - name: DISABLE_OCP_FEATURES
+        value: "$(DISABLE_OCP_FEATURES)"
```
chancez: This value doesn't look right. You want `"{{ .Values.operator.disableOCPFeatures }}"` probably.

images/metering-ansible-operator/roles/meteringconfig/tasks/reconcile.yml
```diff
@@ -14,8 +14,16 @@
   vars:
     hiveStorageType: "{{ meteringconfig_spec_overrides | json_query('storage.hive.type') }}"
 
+- name: Validate fields when OCP features are disabled
+  assert:
+    that:
+      - meteringconfig_spec | json_query('reporting-operator.spec.config.prometheus.url') != ""
```
timflannagan1: @chancez I put this check in place to ensure that the URL is set, and wasn't sure if we wanted to use the default name (`url: https://prometheus-k8s.openshift-monitoring.svc:9091/`). I can reset the name of this field in the override variables to enforce this.

images/metering-ansible-operator/roles/meteringconfig/tasks/reconcile.yml
```diff
@@ -14,8 +14,16 @@
...
+- name: Validate fields when OCP features are disabled
+  assert:
+    that:
+      - meteringconfig_spec['reporting-operator'] | json_query('spec.config.prometheus.url') != ""
+    msg: "Invalid configuration for non-OCP distributions. You must set the reporting-operator.spec.config.prometheus.url."
+  when: meteringconfig_ocp_disabled
+
 - name: Configure TLS
   include_tasks: configure_tls.yml
+  when: not meteringconfig_ocp_disabled
```
chancez: This needs to be removed otherwise TLS being enabled, then disabled won't work correctly.

images/metering-ansible-operator/roles/meteringconfig/tasks/reconcile.yml
```diff
@@ -14,8 +14,16 @@
   vars:
     hiveStorageType: "{{ meteringconfig_spec_overrides | json_query('storage.hive.type') }}"
 
+- name: Validate fields when OCP features are disabled
```
chancez: Maybe we should just be checking this in a validate_reporting_operator section and remove the `when`, since if a user overrides the value to empty, in any situation (OCP or not) it's invalid.

images/metering-ansible-operator/roles/meteringconfig/tasks/reconcile.yml
chancez: See my other comment. I think validating it makes sense. I also think _not_ having a defalt for non-OCP is probably best.

Perhaps we should unset the prometheus.url values.yaml default, and use an override to set the default for OCP, and then it's unset for non-OCP (which will fail validation if they don't set the prometheus URL).

images/metering-ansible-operator/roles/meteringconfig/tasks/reconcile.yml
timflannagan1: Yeah, makes sense to remove the `when` conditional, but I'm not sure this belongs in the `validate_reporting_operator_tls.yml` task file.

timflannagan1: I could either create a task file for validating this one task, or place it elsewhere. I think it makes sense to validate that this value is set for non-okd distributions early in the role execution.

hack/upstream-install.sh
```diff
@@ -0,0 +1,15 @@
+#! /bin/bash
+
+ROOT_DIR=$(dirname "${BASH_SOURCE}")/..
+source "${ROOT_DIR}/hack/common.sh"
+source "${ROOT_DIR}/hack/lib/customize-manifests.sh"
+
+set +u
+
+if [[ -z "$DISABLE_OCP_FEATURES" || "$DISABLE_OCP_FEATURES" = "false" ]]; then
```
timflannagan1: This can most likely be reduced into ```INSTALLER_MANIFESTS_DIR="$UPSTREAM_INSTALLER_MANIFESTS_DIR"``` throughout the upstream install/uninstall scripts.

hack/upstream-install.sh
```diff
@@ -0,0 +1,15 @@
+#! /bin/bash
+
+ROOT_DIR=$(dirname "${BASH_SOURCE}")/..
+source "${ROOT_DIR}/hack/common.sh"
+source "${ROOT_DIR}/hack/lib/customize-manifests.sh"
+
+set +u
```
timflannagan1: I can remove this if I don't need to check the value of the `DISABLE_OCP_FEATURES` environment variable.

images/metering-ansible-operator/roles/meteringconfig/defaults/main.yml
```diff
@@ -284,6 +284,60 @@ meteringconfig_storage_s3Compatible_create_secret: "{{ _storage_spec | json_quer
 meteringconfig_storage_s3Compatible_credentials_secret_name: "{{ _storage_spec | json_query('hive.s3Compatible.secretName') }}"
 meteringconfig_storage_s3Compatible_endpoint: "{{ _storage_spec | json_query('hive.s3Compatible.endpoint') }}"
 
+#
+# Override the default values to handle Openshift, and non-Openshift configurations
+#
+meteringconfig_ocp_disabled: "{{ lookup('env', 'DISABLE_OCP_FEATURES') | bool or meteringconfig_default_values | combine(meteringconfig_spec_overrides, recursive=True) | json_query('disableOCPFeatures.enabled') }}"
+_meteringconfig_ocp_disabled_overrides: "{{ meteringconfig_ocp_disabled | ternary(_ocp_disabled_overrides, _ocp_enabled_overrides) }}"
```
timflannagan1: If `meteringconfig_ocp_disabled` is false, then instead of assigning these overrides w/ an empty dictionary, it assigns them the `_ocp_enabled_overrides`, which only handles setting the prometheus url.

images/metering-ansible-operator/roles/meteringconfig/defaults/main.yml
```diff
@@ -475,7 +529,7 @@ meteringconfig_reporting_overrides:
           enabled: "{{ meteringconfig_reporting_enable_post_kube_1_14_datasources }}"
 
 # combine the _meteringconfig_tls_overrides dictionary last to enforce when spec.tls.enabled is specified and set to true
-meteringconfig_spec: "{{ meteringconfig_default_values | combine(meteringconfig_default_image_overrides, meteringconfig_storage_overrides, meteringconfig_reporting_overrides, meteringconfig_spec_overrides, _meteringconfig_root_ca_overrides, _meteringconfig_tls_overrides, recursive=True) }}"
+meteringconfig_spec: "{{ meteringconfig_default_values | combine(meteringconfig_default_image_overrides, meteringconfig_storage_overrides, meteringconfig_reporting_overrides, _meteringconfig_ocp_disabled_overrides, meteringconfig_spec_overrides, _meteringconfig_root_ca_overrides, _meteringconfig_tls_overrides, recursive=True) }}"
```
timflannagan1: I put the ```_meteringconfig_ocp_disabled_overrides``` dictionary before the ```meteringconfig_spec_overrides``` to handle overriding the default Prometheus URL set in early role execution in a `MeteringConfig` configuration.

charts/openshift-metering/templates/reporting-operator/reporting-operator-config.yaml
```diff
@@ -34,7 +34,7 @@ data:
 {{- end }}
 {{- end }}
 
-  prometheus-url: {{ required "a valid reporting-operator.spec.config.prometheusURL must be set" $operatorValues.spec.config.prometheus.url | quote}}
```
timflannagan1: This was messing up the `make verify` tests, and we already do this validation in the ansible role, so I removed it.

timflannagan1:
@chancez do we need to disable monitoring by default? On a fresh minikube instance, I get the error:
```
fatal: [localhost]: FAILED! => {"changed": false, "msg": "Failed to find exact match for monitoring.coreos.com/v1.ServiceMonitor by [kind, name, singularName, shortNames]"}
```

I previously disabled this in my `METERING_CR_FILE` configuration.

hack/lib/init.sh
```diff
@@ -20,13 +20,22 @@ source "${ROOT_DIR}/hack/lib/util.sh"
 : "${DELETE_PVCS:=true}"
 
 : "${DEPLOY_PLATFORM:=openshift}"
```
chancez: Should probably rename this to OCP_DEPLOY_PLATFORM and then just set DEPLOY_PLATFORM to it `: ${DEPLOY_PLATFORM:=$OCP_DEPLOY_PLATFORM}`

hack/upstream-install.sh
chancez: We shouldn't ever need to check `DISABLE_OCP_FEATURES` in these scripts unless it's for running local. It shouldb ehandled by the manifests themselves.

hack/upstream-install.sh
chancez: Yes there shouldn't be any use of `DISABLE_OCP_FEATURES` in here. And yes, we'll want to set `export INSTALLER_MANIFESTS_DIR="$UPSTREAM_INSTALLER_MANIFESTS_DIR"`

hack/upstream-install.sh
```diff
@@ -0,0 +1,15 @@
...
+
+if [[ -z "$DISABLE_OCP_FEATURES" || "$DISABLE_OCP_FEATURES" = "false" ]]; then
+    INSTALLER_MANIFESTS_DIR="$OCP_INSTALLER_MANIFESTS_DIR"
+else
+    INSTALLER_MANIFESTS_DIR="$UPSTREAM_INSTALLER_MANIFESTS_DIR"
+fi
+
+source "${ROOT_DIR}/hack/install.sh"
```
chancez: No `source`, we're executing the script. Also pass `"$@"` to the install.sh as args

hack/upstream-olm-install.sh
```diff
@@ -0,0 +1,14 @@
+#! /bin/bash
+
+ROOT_DIR=$(dirname "${BASH_SOURCE}")/..
+source "${ROOT_DIR}/hack/common.sh"
+
+set +u
+
+if [[ -z "$DISABLE_OCP_FEATURES" || "$DISABLE_OCP_FEATURES" = "false" ]]; then
```
chancez: Same as `upstream-install.sh`, none of this should use `$DISABLE_OCP_FEATURES`

hack/upstream-olm-uninstall.sh
```diff
@@ -0,0 +1,14 @@
+#! /bin/bash
+
+ROOT_DIR=$(dirname "${BASH_SOURCE}")/..
+source "${ROOT_DIR}/hack/common.sh"
+
+set +u
+
+if [[ -z "$DISABLE_OCP_FEATURES" || "$DISABLE_OCP_FEATURES" = "false" ]]; then
```
chancez: Same as others

hack/upstream-uninstall.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+ROOT_DIR=$(dirname "${BASH_SOURCE}")/..
+source "${ROOT_DIR}/hack/common.sh"
+source "${ROOT_DIR}/hack/lib/customize-manifests.sh"
+
+set +u
+
+if [[ -z "$DISABLE_OCP_FEATURES" || "$DISABLE_OCP_FEATURES" = "false" ]]; then
```
chancez: Same as others

images/metering-ansible-operator/roles/meteringconfig/defaults/main.yml
chancez: Awesome.

images/metering-ansible-operator/roles/meteringconfig/defaults/main.yml
```diff
@@ -284,6 +284,60 @@ meteringconfig_storage_s3Compatible_create_secret: "{{ _storage_spec | json_quer
 meteringconfig_storage_s3Compatible_credentials_secret_name: "{{ _storage_spec | json_query('hive.s3Compatible.secretName') }}"
 meteringconfig_storage_s3Compatible_endpoint: "{{ _storage_spec | json_query('hive.s3Compatible.endpoint') }}"
 
+#
+# Override the default values to handle Openshift, and non-Openshift configurations
+#
+meteringconfig_ocp_disabled: "{{ lookup('env', 'DISABLE_OCP_FEATURES') | bool or meteringconfig_default_values | combine(meteringconfig_spec_overrides, recursive=True) | json_query('disableOCPFeatures.enabled') }}"
+_meteringconfig_ocp_disabled_overrides: "{{ meteringconfig_ocp_disabled | ternary(_ocp_disabled_overrides, _ocp_enabled_overrides) }}"
+
+_meteringconfig_ocp_prometheus_default_url: "https://prometheus-k8s.openshift-monitoring.svc:9091/"
```
chancez: We only seem to use this once, maybe we should just set this directly in `_ocp_enabled_overrides`

images/metering-ansible-operator/roles/meteringconfig/defaults/main.yml
chancez: Does TLS overrides coming later have any potential issues?

images/metering-ansible-operator/roles/meteringconfig/defaults/main.yml
```diff
@@ -461,7 +515,7 @@ _tls_overrides:
 
 
 # check if the user's spec contains the top-level tls.enabled field. if true, use the value stored in that field, else default to the value stored in values.yaml
-meteringconfig_tls_enabled: "{{ meteringconfig_default_values | combine(meteringconfig_spec_overrides, recursive=True) | json_query('tls.enabled') }}"
+meteringconfig_tls_enabled: "{{ meteringconfig_default_values | combine(meteringconfig_spec_overrides, recursive=True) | json_query('tls.enabled') and not meteringconfig_ocp_disabled }}"
```
chancez: Hmm, I dislike this. I feel like we may want to reorganize the `meteringconfig_spec` overrides ordering so this isn't necessary and so that disable OCP takes precedence over it. Thinking...

images/metering-ansible-operator/roles/meteringconfig/defaults/main.yml
chancez: Oh I see above, I'm not sure that I'm a fan of this approach since it's a bit harder to fully understand, but let me think on it.

images/metering-ansible-operator/roles/meteringconfig/defaults/main.yml
timflannagan1: I can do some more manual testing to cover this case, and in the case you want to go away from this implementation, I would need to update the `_ocp_disabled_overrides` to include all the sections that `_tls_overrides` handles too.

chancez: I don't understand that, why can't we just have 
```
meteringconfig_tls_enabled: "{{ meteringconfig_default_values | combine(_meteringconfig_ocp_disabled_overrides, meteringconfig_spec_overrides, recursive=True) | json_query('tls.enabled') }}"
```

`_meteringconfig_ocp_disabled_overrides` will have `tls.enabled` false, thus `meteringconfig_tls_enabled` will be false, thus, `_tls_overrides` will be `{}`.

timflannagan1: Yeah that would definitely do the trick - I was under the impression you didn't want to link the `meteringconfig_tls_enabled` and the value of the `meteringconfig_ocp_disabled`, and instead, only modify the order that their respective dictionaries appear in `meteringconfig_spec`.

chancez: That's correct, I wanted to decouple them, but I think we have no choice, but this is a little more general than the previous approach. They're intrinsically related, so we must know the values of the other. But at least this makes the way we do it consistent with how we check values elsewhere.

charts/metering-ansible-operator/upstream-values.yaml
```diff
@@ -0,0 +1,515 @@
...
+        ]
+
+  subscriptionName: metering
+  subscriptionChannel: stable
+  subscriptionCatalogSource: redhat-operators
+  subscriptionCatalogSourceNamespace: openshift-marketplace
+
+  # I assume this needs to be changed
+  operatorGroupName: metering-operators
+  operatorGroupTargetNamespaces: ["openshift-metering"]
```
chancez: We should probably remove  `openshift-` prefix

charts/metering-ansible-operator/upstream-values.yaml
```diff
@@ -0,0 +1,515 @@
...
+              "location": "s3a://my-bucket/apache_logs",
+              "rowFormat": "SERDE 'org.apache.hadoop.hive.serde2.RegexSerDe'\nWITH SERDEPROPERTIES (\n  \"input.regex\" = \"([^ ]*) ([^ ]*) ([^ ]*) (-|\\\\[[^\\\\]]*\\\\]) ([^ \\\"]*|\\\"[^\\\"]*\\\") (-|[0-9]*) (-|[0-9]*)(?: ([^ \\\"]*|\\\"[^\\\"]*\\\") ([^ \\\"]*|\\\"[^\\\"]*\\\"))?\"\n)\n",
+              "tableName": "apache_log"
+            }
+          }
+        ]
+
+  subscriptionName: metering
+  subscriptionChannel: stable
+  subscriptionCatalogSource: redhat-operators
```
chancez: should be `upstream-community-operators` I think

charts/metering-ansible-operator/upstream-values.yaml
```diff
@@ -0,0 +1,515 @@
...
+  subscriptionCatalogSourceNamespace: openshift-marketplace
```
chancez: Needs to be modified, and figure out what the correct value is. Not sure if/how this is installed when using OLM on upstream.

charts/metering-ansible-operator/upstream-values.yaml
```diff
@@ -0,0 +1,515 @@
...
+      - roles
+      verbs: ["*"]
+    - apiGroups:
+      - route.openshift.io
+      resources:
+      - routes
+      verbs: ["*"]
+
+olm:
+  packageName: metering-ocp
```
chancez: rename

timflannagan1: `metering-ocp` -> `metering-upstream`

Makefile
```diff
@@ -167,17 +167,20 @@ vet:
...
-verify: verify-codegen verify-olm-manifests verify-helm-templates fmt vet
+verify: verify-codegen verify-ocp-olm-manifests verify-upstream-olm-manifests verify-helm-templates fmt vet
 	@echo Checking for unstaged changes
 	# validates no unstaged changes exist in $(VERIFY_FILE_PATHS)
 	git diff --stat HEAD --ignore-submodules --exit-code -- $(VERIFY_FILE_PATHS)
 
 verify-helm-templates:
 	helm template ./charts/openshift-metering > /dev/null
 
-verify-olm-manifests: metering-manifests
```
chancez: I feel like we should just have a single target which verifies both. Any reason you did them separately?

charts/metering-ansible-operator/upstream-values.yaml
```diff
@@ -0,0 +1,64 @@
...
+    tag: release-4.2
+    pullPolicy: Always
+
+olm:
+  packageName: metering-upstream
+  channels:
+  - name: stable
+    currentCSV: metering-operator.v4.2.0
+
+  imageTags:
```
chancez: Need to add ghostunnel

charts/metering-ci/templates/metering.yaml
```diff
@@ -56,6 +56,7 @@ spec:
           secretAccessKey: "{{ .Values.awsSecretAccessKey }}"
 
         prometheus:
+          url: "https://prometheus-k8s.openshift-monitoring.svc:9091/"
```
chancez: This shouldn't be necessary.

charts/openshift-metering/values.yaml
```diff
@@ -4,6 +4,9 @@ global:
 unsupportedFeatures:
   enableHDFS: false
 
+disableOCPFeatures:
```
chancez: I just realized this has some double negatives going on. Let's not use a nested field here, let's just have `disableOCPFeatures: false`

hack/lib/init.sh
```diff
@@ -19,14 +19,24 @@ source "${ROOT_DIR}/hack/lib/util.sh"
...
+: "${OCP_DEPLOY_PLATFORM:=openshift}"
+: "${UPSTREAM_DEPLOY_PLATFORM:=upstream}"
 METERING_NAMESPACE=$(sanetize_namespace "${METERING_NAMESPACE:-metering}")
 
+: "${DEPLOY_PLATFORM:=$OCP_DEPLOY_PLATFORM}"
 : "${DEPLOY_MANIFESTS_DIR:=$MANIFESTS_DIR/deploy}"
 : "${RBAC_MANIFESTS_DIR:=$MANIFESTS_DIR/rbac}"
-: "${INSTALLER_MANIFESTS_DIR:=$DEPLOY_MANIFESTS_DIR/$DEPLOY_PLATFORM/metering-ansible-operator}"
+
+: "${OCP_INSTALLER_MANIFESTS_DIR:=$DEPLOY_MANIFESTS_DIR/$DEPLOY_PLATFORM/metering-ansible-operator}"
```
chancez: This should use `$OCP_DEPLOY_PLATFORM`  instead of `$DEPLOY_PLATFORM` I think?

Makefile
timflannagan1: nope, was just doing this on the fly and didn't consider that - I'll update that

charts/metering-ansible-operator/upstream-values.yaml
timflannagan1: good catch

images/metering-ansible-operator/roles/meteringconfig/defaults/main.yml
```diff
@@ -291,6 +291,61 @@ meteringconfig_storage_s3Compatible_create_secret: "{{ _storage_spec | json_quer
 meteringconfig_storage_s3Compatible_credentials_secret_name: "{{ _storage_spec | json_query('hive.s3Compatible.secretName') }}"
 meteringconfig_storage_s3Compatible_endpoint: "{{ _storage_spec | json_query('hive.s3Compatible.endpoint') }}"
 
+#
+# Override the default values to handle Openshift, and non-Openshift configurations
+#
+meteringconfig_ocp_disabled: "{{ lookup('env', 'DISABLE_OCP_FEATURES') | bool or meteringconfig_default_values | combine(meteringconfig_spec_overrides, recursive=True) | json_query('disableOCPFeatures.enabled') }}"
```
chancez: `json_query('disableOCPFeatures.enabled')` should be `json_query('disableOCPFeatures')`

charts/metering-ansible-operator/templates/crds/meteringconfig.crd.yaml
```diff
@@ -39,6 +39,12 @@ spec:
                 enableHDFS:
                   type: boolean
 
+            disableOCPFeatures:
+              type: object
```
chancez: type boolean

charts/metering-ansible-operator/templates/crds/meteringconfig.crd.yaml
```diff
@@ -39,6 +39,12 @@ spec:
...
+              properties:
+                enabled:
```
chancez: No enabled field anymore

hack/generate-metering-manifests.sh
```diff
@@ -8,5 +8,10 @@ TMPDIR="$(mktemp -d)"
...
 "$ROOT_DIR/hack/create-metering-manifests.sh" \
-    "$INSTALLER_MANIFESTS_DIR" \
-    "$OLM_MANIFESTS_DIR"
+    "$OCP_INSTALLER_MANIFESTS_DIR" \
+    "$OCP_OLM_MANIFESTS_DIR" \
+
+"$ROOT_DIR/hack/create-metering-manifests.sh" \
+    "$UPSTREAM_INSTALLER_MANIFESTS_DIR" \
+    "$UPSTREAM_OLM_MANIFESTS_DIR" \
+    "$ROOT_DIR/charts/metering-ansible-operator/upstream-values.yaml" \
```
chancez: extra `\` at end

hack/generate-metering-manifests.sh
```diff
@@ -8,5 +8,10 @@ TMPDIR="$(mktemp -d)"
 trap "rm -rf $TMPDIR" EXIT
 
 "$ROOT_DIR/hack/create-metering-manifests.sh" \
-    "$INSTALLER_MANIFESTS_DIR" \
-    "$OLM_MANIFESTS_DIR"
+    "$OCP_INSTALLER_MANIFESTS_DIR" \
+    "$OCP_OLM_MANIFESTS_DIR" \
```
chancez: extra `\` at end

PR #903, opened 2019-08-16
Add OpenAPIV3 schema validation to the StorageLocation CRD.

timflannagan1:


Documentation/storagelocations.md
```diff
@@ -1,23 +1,22 @@
...
+Normally, users shouldn't need to configure a `StorageLocation` resource unless they want to store data in multiple locations, like multiple S3 buckets or both S3 and HDFS, or if they wish to access a database in Hive/Presto that was not created by metering.
 Instead, users should use the [configuring storage](configuring-storage.md) documentation to manage configuration of all components in the metering stack.
 
 The Operator Metering default installation provides a few ways of configuring the [Default StorageLocation](#default-storagelocation), and normally it shouldn't be necessary to create these directly.
 Refer to the [Metering Configuration doc](metering-config.md#storing-data-in-s3) for details on using the `Metering` resource to set your default StorageLocation.
 
 ## Fields
-
-- `hive`: If this section is present, then the `StorageLocation` will be configured to store data in Presto by creating the table using Hive server.
+- `hive`: If this section is present, then the `StorageLocation` will be configured to store data in Presto by creating the table using Hive server. Note: only `databaseName` and `unmanagedDatabase` are required fields.
```
chancez: Why remove the information on what's required not required in the fields?

timflannagan1: I was trying to conform with the format used in (prestotables.md)[https://github.com/operator-framework/operator-metering/blob/master/Documentation/prestotables.md#fields] and (hivetables.md)[https://github.com/operator-framework/operator-metering/blob/master/Documentation/hivetables.md#fields], but it's a bit awkward to replicate that pattern when the optional fields are nested under the `hive` property, i.e.:
```
## Fields
##### Required Fields:
- `hive`: If this section is present, then the `StorageLocation` will be configured to store data in Presto by creating the table using Hive server. Note: only `databaseName` and `unmanagedDatabase` are required fields.
  - `databaseName`: The name of the database within hive.
  - `unmanagedDatabase`: If true, then this StorageLocation will not be actively managed, and the databaseName is expected to already exist in Hive. If false, this will cause the reporting-operator to create the Database in Hive.
##### Optional Fields:
  - `location`: The filesystem URL for Presto and Hive to use for the database. This can be an `hdfs://` or `s3a://` filesystem URL.
  - `defaultTableProperties`: Contains configuration options for creating tables using Hive.
    - `fileFormat`: The file format used for storing files in the filesystem. See the [Hive Documentation on File Storage Format for a list of options and more details][hiveFileFormat].
    - `rowFormat`: Controls the [Hive row format][hiveRowFormat]. This controls how Hive serializes and deserializes rows. See the [Hive Documentation on Row Formats & SerDe for more details][hiveRowFormat].
```

I'll just revert the changes made to this section back to the original layout.

chancez: I think this leads to the point where, at some level, we'll want each field to be it's own heading so we can link to individual fields in the docs. That may be a bit much to do now though, so I think going back to how it was may be best until we revisit the doc structure for fields.

PR #905, opened 2019-08-19
hack: Update the target directory location for the minikube directory.

timflannagan1:
I was still getting an error when running `make run-metering-operator-local` saying that the `$HOME/.minikube/*` certificates were missing, but `minikube ssh -- ls $HOME/.minikube` verified they were there. 

After adding a target directory location, I was able to run the metering operator locally.


chancez:
Haha whoops!

PR #908, opened 2019-08-21
Remove *_DEPLOY_REPO and *_DEPLOY_TAG variables

chancez:
This removes METERING_OPERATOR_DEPLOY_REPO, METERING_OPERATOR_DEPLOY_TAG,
REPORTING_OPERATOR_DEPLOY_REPO, and REPORTING_OPERATOR_DEPLOY_TAG variables
from scripts to simplify our logic to override the metering and reporting
operator images at install time.

The `*_DEPLOY_*` variables previously were used by e2e/integration to
regenerate the metering-manifests with different images for metering-operator,
and to override the MeteringConfig's image for reporting-operator during CI.

For a while now, we've supported overriding the metering-operators image
information directly in install.sh, meaning we had two ways to override the
image being used.

This PR removes the old method in favor of leveraging the logic in install.sh,
and unifies all the variables for overriding the image to use to use the same
ones variables been using for install.sh, namely:
`METERING_OPERATOR_IMAGE_REPO` and `METERING_OPERATOR_IMAGE_TAG`.
For reporting-operator we've simply switched the names of the variables to
match the same convention.


hack/create-metering-manifests.sh
```diff
@@ -39,43 +40,43 @@ trap 'rm -rf $TMPDIR' EXIT SIGINT
 CHART="$ROOT_DIR/charts/metering-ansible-operator"
 
 helm template "$CHART" \
-    "${VALUES_ARGS[@]}" \
+    ${VALUES_ARGS[@]+"${VALUES_ARGS[@]}"} \
```
timflannagan1: I'm not sure I understand this, I took a look at [indexing array by the @ character](https://www.gnu.org/savannah-checkouts/gnu/bash/manual/bash.html#Shell-Parameter-Expansion), so wouldn't `VALUES_ARGS` always be empty?

timflannagan1: I should've viewed the entire file, I didn't see the while loop on line 28 - nvm

chancez: Also read the commit message.

PR #915, opened 2019-08-28
hack,charts,manifests: Skip generating the ART package for non-ocp manifests.

timflannagan1:
This is a follow-up to the PR #900 to get it ready for the upstream-community operators.


chancez:
I'm not a fan of how 1d30a4ff2fa0f3caf2e2138e1fec4100519af22f is done, so can you remove that from this PR and tackle it in a different PR? It can't be hardcoded because the CSV name isn't going to be the same for each platform.

Ideally you could source it from the file (using `faq`), or by querying the subscription before deleting it, to get the `currentCSV` field value (using `faq`). Either way, let's do that in it's own PR.

timflannagan1:
@chancez good call - updated.

PR #916, opened 2019-08-28
Documentation: Improve documentation on create secrets for storage credentials

chancez:


chancez:
I know, but they're all different, and I feel like users will miss it if we don't put it in each section, if we only put it at the top.

Issue #922, opened 2019-09-06
cannot set allowIncompleteChunks to false


gangseok514:
Hello, I am looking into operater-metering sourcecode recently due to some issues.

I set pollInterval: 10m, chunksize: 10m, stepsize: 5m

In the importer.go and query.go sourcecode,
lastTimestamp is set current time because of allowIncompleteChunks,
so the next collected metric is unintentional.

example:
```
time="2019-09-05T10:19:31Z" level=info msg="stored a total of 700 metrics for data between 2019-09-05 10:17:00 +0000 UTC and 2019-09-05 10:19:00 +0000 UTC into datasource_gpu_request_slots" ReportDataSource=gpu-request-slots app=metering chunkSize=10m0s component=PrometheusImporter endTime="2019-09-05 10:19:00.044390211 +0000 UTC" logID=5cT8O1v0rR queryName=gpu-request-slots reportDataSource=gpu-request-slots startTime="2019-09-05 10:17:00 +0000 UTC" stepSize=5m0s tableName=datasource_gpu_request_slots

I0905 19:31:58.603887   20495 kubectl.go:56] Execution time : 463.397807ms
time="2019-09-05T10:29:37Z" level=info msg="stored a total of 1431 metrics for data between 2019-09-05 10:24:00 +0000 UTC and 2019-09-05 10:29:00 +0000 UTC into datasource_gpu_request_slots" ReportDataSource=gpu-request-slots app=metering chunkSize=10m0s component=PrometheusImporter endTime="2019-09-05 10:29:33.589955865 +0000 UTC" logID=5cT8O1v0rR queryName=gpu-request-slots reportDataSource=gpu-request-slots startTime="2019-09-05 10:24:00 +0000 UTC" stepSize=5m0s tableName=datasource_gpu_request_slots
```

In fact, I would like to collect metric at 10:17 , 10:22(+5m), 10:27(+5m) by stepsize.

but lastTimestamp is set 10:19 because of allowIncompleteChunks,
So the unintentional metrics between 10:24(10:19+5m) and 10:29 are collected.

```golang
func getTimeRangesChunked(beginTime, endTime time.Time, chunkSize, stepSize time.Duration, maxTimeRanges int64, allowIncompleteChunks bool) []prom.Range {
	chunkStart := truncateToSecond(beginTime)
	chunkEnd := truncateToSecond(chunkStart.Add(chunkSize))

	// don't set a limit if negative or zero
[...]

PR #925, opened 2019-09-09
Always import full chunks when querying metrics

chancez:
This PR changes metrics importing to always import metrics of the configured import chunk size.
This ensures consistency in metrics being imported.


chancez:
Fixes #922

PR #936, opened 2019-09-16
Documentation/dev: Add developer documentation for using Metering with NooBaa

chancez:
This is for developers/QE to begin testing NooBaa. This will be updated
more as NooBaa becomes GA.

### DIFF
--- a/Documentation/dev/noobaa.md
+++ b/Documentation/dev/noobaa.md
@@ -1,0 +1,133 @@
+# Installing NooBaa
+
+Follow the instructions on https://github.com/noobaa/noobaa-operator/ for installing NooBaa.
+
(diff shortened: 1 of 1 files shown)


Documentation/dev/noobaa.md
```diff
@@ -0,0 +1,133 @@
...
+```
+kubectl create ns $METERING_NAMESPACE
+```
+
+### Create noobaa credentials secret
+
+Run the following command to print our your credentials:
+
+```
+noobaa status 2>&1 | grep AWS_
```
pruan-rht: @chancez alternatively, we can get a more parsible output with `oc get secret noobaa-admin -n noobaa -o yaml`

PR #938, opened 2019-09-17
Fix package file name on art.yaml

thiagoalessio:
OCP build failing:
```
ose-metering-ansible-operator: /.../manifests/deploy/openshift/olm/bundle/package.yaml
      does not exist as defined in art.yaml
```


chancez:
That file is auto-generated, I opened another PR that updates the config + the generated files #939

PR #934, opened 2019-09-16
Add a cobra cli that manages the deployment process of metering.

timflannagan1:
### Overview
This PR adds a cobra cli that adds commands/flags which are used to customize the deployment process of metering.

### Building the CLI:
```bash
$ make bin/deploy-metering
```
This creates the `deploy-metering` binary in `./bin/`

### Design
Most of the logic for setting the `Deployer` internal states have moved away from all environment variables, to a mixture of cli flags and environment variables. In the current iteration of this CLI, the value of the cli flags take precedence of the value of the environment variables:
```bash
$ export METERING_NAMESPACE="metering"
$ ./bin/deploy-metering install --namespace test
INFO[09-17-2019 14:54:42] Setting the log level to info                 app=deploy
INFO[09-17-2019 14:54:42] Metering Deploy Namespace: test               app=deploy
...
```

##### Vanilla Openshift Install:
```bash
$ ./bin/deploy-metering install
```
##### Vanilla Upstream install:
```bash
$ ./bin/deploy-metering install --platform upstream
```
##### Vanilla Openshift Uninstall:
```bash
$ ./bin/deploy-metering uninstall
```
##### Delete the `$METERING_NAMESPACE` (`--namespace`) namespace:
```bash
$ ./bin/deploy-metering uninstall --delete-namespace
```

PR #940, opened 2019-09-17
cmd,pkg/deploy: Refactor the Deployer structure and methods.

timflannagan1:


cmd/deploy/main.go
```diff
@@ -15,6 +15,7 @@ import (
 
 func main() {
 	logger := setupLogger("info")
+	config := deploy.Config{}
```
chancez: var config deploy.Config

timflannagan1: I'll update in #934

PR #942, opened 2019-09-19
pkg/apis: Update fields and json encodings in the MeteringConfig structure.

timflannagan1:


timflannagan1:
@chancez yes, it's used in the `spec.storage.hive`.

chancez:
I see.

PR #944, opened 2019-09-19
hack,Makefile: Update the deploy scripts to use the deploy-metering binary.

timflannagan1:
This is based on top of #940, which adds a CLI for deploying metering resources.

In this PR, all the `./hack/*install.sh` and `./hack/*uninstall.sh` scripts are replaced (except for olm) in the deploy scripts and now use the `deploy-metering` binary to manage the metering resources. Also, the `./hack/run-metering-operator-local.sh` now calls the `./hack/lib/util.sh install_metering()` helper function to install metering resources. 

At this moment, only Openshift/OCP and OLM installs are tested in e2e/integration suite. In the future, we can think about testing upstream installs (e.g. `install_metering upstream`) and alter the `install_metering` helper function to call `./bin/deploy-metering install --platform upstream` (or `export DEPLOY_PLATFORM="upstream"` and forgo the platform flag.)

### DIFF
--- a/hack/deploy.sh
+++ b/hack/deploy.sh
@@ -8,7 +8,8 @@ source "${ROOT_DIR}/hack/common.sh"
 
 : "${UNINSTALL_METERING_BEFORE_INSTALL:=true}"
 : "${INSTALL_METERING:=true}"
-: "${INSTALL_METHOD:=openshift-direct}"
+: "${INSTALL_METHOD:=openshift}"
+: "${UNINSTALL_METHOD:=openshift}"
 : "${METERING_CREATE_PULL_SECRET:=false}"
 : "${METERING_PULL_SECRET_NAME:=metering-pull-secret}"
 : "${DEPLOY_METERING_OPERATOR_LOCAL:=false}"
@@ -20,7 +21,8 @@ fi
 
 if [ "$UNINSTALL_METERING_BEFORE_INSTALL" == "true" ]; then
     echo "Uninstalling metering"
-    kubectl delete ns "$METERING_NAMESPACE" || true
+    export METERING_DELETE_NAMESPACE=true
+    uninstall_metering "${UNINSTALL_METHOD}"
 else
     echo "Skipping uninstall"
 fi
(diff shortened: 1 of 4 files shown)


timflannagan1:
Oh yeah forgot to build the binary.

hack/deploy.sh
```diff
@@ -8,7 +8,8 @@ source "${ROOT_DIR}/hack/common.sh"
 
 : "${UNINSTALL_METERING_BEFORE_INSTALL:=true}"
 : "${INSTALL_METERING:=true}"
-: "${INSTALL_METHOD:=openshift-direct}"
+: "${INSTALL_METHOD:=generic}"
+: "${UNINSTALL_METHOD:=generic}"
 : "${METERING_CREATE_PULL_SECRET:=false}"
```
EmilyM1: What about naming it "openshift" instead of "generic"?

hack/run-metering-operator-local.sh
```diff
@@ -6,7 +6,7 @@ source "${ROOT_DIR}/hack/common.sh"
 
 : "${METERING_OPERATOR_IMAGE:=${METERING_OPERATOR_IMAGE_REPO}:${METERING_OPERATOR_IMAGE_TAG}}"
 : "${LOCAL_METERING_OPERATOR_RUN_INSTALL:=true}"
-: "${METERING_INSTALL_SCRIPT:=./hack/openshift-install.sh}"
+: "${INSTALL_METHOD:=generic}"
```
chancez: Should default to openshift no?

chancez:
/hold

So far this seems fine, but we need to be very careful to not break e2e/integration. I'd like to hold on this until 4.2 freeze is done, then we can merge without affecting 4.2 builds/CI.

hack/run-metering-operator-local.sh
timflannagan1: Fixed.

timflannagan1:
1.16 rebase is done
/hold cancel

PR #951, opened 2019-09-25
*: Update to go 1.12

chancez:
Updates Dockerfiles to all use Go 1.12 and dev-guide to require Go 1.12


chancez:
/hold
I want to make sure everyone on the team is updated to Go 1.12 before this merges.

chancez:
Our ci-operator/prow CI doesn't use the Dockerfile.src from the PR, but rather the merge HEAD (master) so this still fails, but should pass once merged.

PR #954, opened 2019-09-26
Set XDG_CACHE_HOME to /tmp

chancez:
Running builds as in CI is failing with failed to initialize build cache at /go/.cache: open /go/.cache/log.txt: permission denied


chancez:
CI doesn't use Dockerfile.src from PRs, so I'm merging to see if that fixes it.

PR #956, opened 2019-09-26
cmd,pkg/deploy: Support passing an object to the MeteringConfig resource.

timflannagan1:
This would refactor the `deploy.Config` structure to take a `MeteringConfig` object, instead of a file that contains a `MeteringConfig` yaml file. This makes for a cleaner structure and makes testing different configurations easier.

### DIFF
--- a/cmd/deploy-metering/main.go
+++ b/cmd/deploy-metering/main.go
@@ -18,9 +18,10 @@ import (
 )
 
 var (
-	cfg        deploy.Config
-	deployType string
-	logLevel   string
+	cfg            deploy.Config
+	deployType     string
+	logLevel       string
+	meteringCRFile string
 
 	rootCmd = &cobra.Command{
 		Use:   "deploy-metering",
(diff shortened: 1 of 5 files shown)


pkg/operator/deploy/deploy.go
```diff
@@ -52,18 +53,18 @@ type CRD struct {
...
-	MeteringCR               string
 	SkipMeteringDeployment   bool
 	DeleteCRDs               bool
 	DeleteCRB                bool
 	DeleteNamespace          bool
 	DeletePVCs               bool
 	DeleteAll                bool
+	Namespace                string
+	Platform                 string
+	DeployManifestsDirectory string
```
chancez: I feel like this should be removed from the `Config` struct and all files should be moved to being read outside of the Deployer code

pkg/operator/deploy/deploy.go
```diff
@@ -52,18 +53,18 @@ type CRD struct {
...
 	Repo                     string
 	Tag                      string
+	MeteringCR               meteringv1.MeteringConfig
```
chancez: `MeteringConfig`

pkg/operator/deploy/deploy.go
```diff
@@ -161,6 +162,11 @@ func NewDeployer(
 	return deploy, nil
 }
 
+// SetConfig is a setter function that updates the unexported config field
```
chancez: We don't need a Setter, we should just take this in the constructor.

pkg/operator/deploy/deploy.go
timflannagan1: I agree, like managing a `MeteringConfig` structure directly, we could also handle building up the YAML manifests outside of the deployer methods and instead manage the Metering CR structures. I'd prefer to do this in a follow-up PR as that's a bigger change.

chancez: Okay. That's fine.

cmd/deploy-metering/main.go
```diff
@@ -18,9 +18,10 @@ import (
 )
 
 var (
-	cfg        deploy.Config
-	deployType string
-	logLevel   string
+	cfg            deploy.Config
+	deployType     string
+	logLevel       string
+	meteringCRFile string
```
chancez: In the follow-up we might want another struct for containing all the paths to the files so we can have function which opens all files + this and returns all the objects in the struct.

PR #958, opened 2019-09-27
bug 1757547: charts/openshift-metering: Update cpu/memory usage promqueries to handle many to many matching error

chancez:
These queries were returning many to many matching errors on the
group_left that I didn't end up diagnosing, but I discovered we can
remove it since the container level metrics got the node name field that
we were joining from kube_pod_info. This simplifies them overall.
Additionally, I bumped up the window size for when we take the rate of
the CPU.

### DIFF
--- a/charts/openshift-metering/values.yaml
+++ b/charts/openshift-metering/values.yaml
@@ -215,12 +215,12 @@ openshift-reporting:
             spec:
               prometheusMetricsImporter:
                 query: |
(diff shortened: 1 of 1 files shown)


chancez:
/hold 
I want to file a bug on this and possibly backport to 4.2

chancez:
This PR and others are blocked until #960 merges

chancez:
The cherry-pick needs https://github.com/operator-framework/operator-metering/pull/962 since that has changes this PR requires (which came from #960)

PR #959, opened 2019-09-27
charts,images: Disable post 1.14 kube Prom datasources

chancez:
It doesn't look like kube 1.16 in OCP 4.3 has the new consistent labels
on container metrics which is breaking our metrics imports for pod usage
metrics.


chancez:
/hold
This may need 4.2 backporting.

chancez:
Well, on 4.2 both sets of metrics work.

PR #960, opened 2019-09-27
bug 1756548: charts/openshift-metering: Fix kube 1.14 metrics queries

chancez:
We need to container, not container_name as the metric name we're
filtering on in container usage metrics. This was missed when changing
from pod_name to pod. In kube 1.16 (OCP 4.3) the older metric labels 
are removed which is what's breaking our CI currently.


chancez:
/hold
I think this would be a good idea to backport to 4.2 even though it may not be required as it will ensure the field name change works in 4.2 as well as 4.3 so we can ensure the migration will work in 4.3

PR #961, opened 2019-09-30
cmd,pkg/deploy: Add function to the deploy package that reads the YAML manifests and builds up objects.

timflannagan1:


pkg/operator/deploy/helpers.go
```diff
@@ -45,47 +46,171 @@ func decodeYAMLManifestToObject(path string, resource interface{}) error {
...
+		"clusterRole": Resource{
+			Path:     meteringClusterRoleFile,
+			Resource: &resources.ClusterRole,
+		},
+	}
+
+	for name, resource := range meteringResourceMap {
+		path := filepath.Join(ansibleOperatorManifestDir, resource.Path)
+
+		if resource.IsMeteringConfig {
```
chancez: Maybe change IsMeteringConfig to `AbsolutePath`, then it's not as much of a one-off.

pkg/operator/deploy/helpers.go
```diff
@@ -45,47 +46,171 @@ func decodeYAMLManifestToObject(path string, resource interface{}) error {
...
+// Resource contains information about an individual resource that metering manages
+type Resource struct {
+	IsMeteringConfig bool
+	Path             string
+	Resource         interface{}
+}
+
+// InitObjectFromManifest is the driver function for building up a path to the metering-ansible-operator directory
+// from the base @manifestDir directory and initializing the metering resource/CRD objects from the YAML manifests.
+func InitObjectFromManifest(manifestDir, meteringConfigCRFile string, cfg *Config, resources *MeteringResources) ([]CRD, error) {
```
chancez: I find the way we're handling different objects so inconsistently a bit odd. We have both `cfg` and `resources`, when we can probably have one. We return CRDs, when it could probably be part of MeteringResources.

How about:

- We return `resources *MeteringResources` instead of taking it as an argument, and instead of returning CRDs.
- CRDs is a field on `*MeteringResources`
- MeteringConfig is a field on `*MeteringResources`

timflannagan1: I tried going the route of eliminating the need for the `deployer.crds` field, but wasn't able to find an elegant solution that didn't involve using pointers and type assertion.

Previously I tried doing something similar to this:
```go
	meteringResourceMap := map[string]Resource{
		"hivetables.metering.io": Resource{
			Path:             hiveTablePath,
                        IsCRD:          true,
			Resource:     new(apiextv1beta.CustomResourceDefinition),
		},
        ...
```

If you have any idea on how to consolidate these resources into one grouping, then I don't think it would be that hard to adapt the metering CRD install/uninstall deploy methods to check for this boolean field.

timflannagan1: Unless I misunderstood, and what you meant was to adapt the `MeteringResources` structure to include managing the slice of CRDs (and `MeteringConfig`), which shouldn't be that difficult to do either.

chancez: I'm not suggesting changing the logic, just moving it. So the latter comment.

pkg/operator/deploy/deploy.go
```diff
@@ -52,18 +54,28 @@ type CRD struct {
...
+	Namespace              string
+	Platform               string
+	Repo                   string
+	Tag                    string
+}
+
+// MeteringResources contains all the objects that metering manages
+type MeteringResources struct {
+	CRDs               []CRD
+	Deployment         appsv1.Deployment
```
chancez: Any reason we're not using pointers for all of these when we need to be passing them as references to all the functions that use them?

PR #965, opened 2019-10-01
pkg/operator,cmd/deploy-metering: Move reading of MeteringConfig into main package

chancez:
Also read InitObjectFromManifest to ReadMeteringAnsibleOperatorManifests.
Moving the reading of MeteringConfig into main means we can re-use
ReadMeteringAnsibleOperatorManifests even if we're creating a
MeteringConfig directly, instead of reading it from a file.

### DIFF
--- a/pkg/operator/deploy/helpers.go
+++ b/pkg/operator/deploy/helpers.go
@@ -133,22 +133,12 @@ func getMeteringAnsiblePath(manifestDir, platform string) (string, error) {
 	return ansibleOperatorManifestDir, nil
 }
 
-// Resource contains information about an individual resource that metering manages
-type Resource struct {
-	UseAbsolutePath  bool
-	SkipYAMLDecoding bool
-	Path             string
-	Resource         interface{}
-}
+func ReadMeteringAnsibleOperatorManifests(manifestDir, platform string) (*OperatorResources, error) {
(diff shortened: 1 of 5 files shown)


pkg/operator/deploy/helpers.go
```diff
@@ -133,28 +133,17 @@ func getMeteringAnsiblePath(manifestDir, platform string) (string, error) {
...
+	ansibleOperatorManifestDir, err := getMeteringAnsiblePath(manifestDir, platform)
 	if err != nil {
-		return fmt.Errorf("Failed to get the path to the metering-ansible-operator directory: %v", err)
+		return nil, fmt.Errorf("Failed to get the path to the metering-ansible-operator directory: %v", err)
 	}
 
 	pathToCRDMap := map[string]string{
 		"hiveTable":        hivetableFile,
 		"prestoTable":      prestotableFile,
-		"meteringConfig":   meteringconfigFile,
```
timflannagan1: this is the metering config CRD, not the CR file

timflannagan1: In general, these names might need to be updated to be more verbose.

chancez: hah oops

PR #971, opened 2019-10-07
WIP: Refactor the e2e testing suite to use the deploy metering package.

timflannagan1:
### Overview:
This would refactor the e2e package to handle both deploying metering and run the reporting tests. This PR adds a deploy framework that's in charge of ensuring all of the metering resources are ready to proceed to run the reporting tests and teardown after the testing phase has completed.

### TODO:
- [x] Update `hack/` scripts to handle `make e2e-local`. These values can be hardcoded for now as there's only one `MeteringConfig` that's being tested
- [ ] Log test output (junit reports)
- [x] Pass an overall testing directory to the deploy framework
- [x] Log resource and containers logs to a tmpdir/specified path + random prefix
- [x] Update `testInstall` to respect the defer closure call

### Follow-up PRs:
- Pass clientsets to `reportingframework.New` instead of that method creating the clientsets explicitly


test/deployframework/framework.go
```diff
@@ -0,0 +1,275 @@
...
+	}
+
+	df.Logger.Debugf("Deployer obj: %+v", df.Deployer)
+
+	err = df.Deployer.Install()
+	if err != nil {
+		return nil, fmt.Errorf("Failed to install metering: %v", err)
+	}
+
+	err = df.addE2ENamespaceLabel(namespace)
```
timflannagan1: Depending on how we want to deal with deleting testing namespaces before running `go test ...`, this is probably not needed.

```bash
# ./hack/e2e.sh
echo "Deleting testing namespaces
kubectl delete ns -l name=e2e-testing
```

chancez: Yeah, not sure which is best yet.

test/e2e/main_test.go
```diff
@@ -3,61 +3,149 @@ package e2e
...
+				Platform:        defaultPlatform,
+				DeleteNamespace: true,
+				MeteringConfig: &meteringv1.MeteringConfig{
+					Spec: testhelpers.NewMeteringConfigSpec(),
+				},
+			},
+		},
+	}
+
+	for _, testCase := range testInstallConfigs {
```
chancez: after this line:
```
t  := t
testCase := testCase
```
https://github.com/golang/go/wiki/CommonMistakes#using-goroutines-on-loop-iterator-variables

test/e2e/report_test.go
```diff
@@ -7,88 +7,90 @@ import (
...
-					if dataSource.Spec.PrometheusMetricsImporter == nil {
-						return true, nil
-					}
-					if dataSource.Status.PrometheusMetricsImportStatus != nil && dataSource.Status.PrometheusMetricsImportStatus.EarliestImportedMetricTime != nil {
-						// keep the EarliestImportedMetricTime that is the
-						// least far back, so that we ensure the reportStart is
-						// a time that all datasources have metrics for.
-						if reportStart.IsZero() || dataSource.Status.PrometheusMetricsImportStatus.EarliestImportedMetricTime.Time.After(reportStart) {
-							reportStart = dataSource.Status.PrometheusMetricsImportStatus.EarliestImportedMetricTime.Time
+	t.Run("group", func(t *testing.T) {
```
chancez: Remove this extra sub-test closure

timflannagan1:
Moving this to a feature branch instead of master (#974)

timflannagan1:
no tests in feature branch so re-opening

test/e2e/report_test.go
timflannagan1: Without this, the defer closure in `testInstall` isn't respected.

test/e2e/main_test.go
```diff
@@ -3,61 +3,166 @@ package e2e
...
+	cleanupScriptPath := flag.String("cleanup-script-path", "../../hack/run-test-cleanup.sh", "The absolute/relative path to the testing cleanup hack script.")
+	testOutputPath := flag.String("test-output-path", "", "The absolute/relative path that you want to store test logs within.")
+	logLevel := flag.String("log-level", logrus.DebugLevel.String(), "The log level")
+	flag.Parse()
+
+	logger := testhelpers.SetupLogger(*logLevel)
+
+	if testOutputDirectory == "" {
+		if *testOutputPath == "" {
+			logger.Fatalf("You must specify the $TEST_OUTPUT_PATH or --test-output-path.")
```
timflannagan1: @chancez how do you want to handle specified directories (through `$TEST_OUTPUT_PATH/--test-output-path)? I think it makes more sense to use that path instead of making a directory with that path + a prefix (e.g. `ioutil.TempDir`).

We also need to either support making a temp directory when the flag/ENV var is unspecified or make it a required value, which doesn't seem ideal.

timflannagan1: We could do something like this, but seems awkward:
```go
	var loggingPath string
	var dirCreated bool

	switch {
	case testOutputDirectory != "":
		loggingPath = testOutputDirectory
	case *testOutputPathFlag != "":
		loggingPath = *testOutputPathFlag
	default:
		loggingPath, err = ioutil.TempDir(testOutputDirectory, *nsPrefix)
		if err != nil {
			logger.Fatalf("Failed to create the test output directory '%s': %v", testOutputDirectory, err)
		}
		dirCreated = true
	}

	if !dirCreated {
		err := os.MkdirAll(testOutputDirectory, 0777)
		if err != nil {
			logger.Fatalf("Failed to create the test output directory '%s': %v", testOutputDirectory, err)
		}
	}
```

PR #973, opened 2019-10-08
test: Remove redundant code block from the reportingframework constructor.

timflannagan1:
This is already done in the testing init function for both integration and e2e:
```go
func init() {
	reportTestOutputDirectory = os.Getenv("TEST_RESULT_REPORT_OUTPUT_DIRECTORY")
	if reportTestOutputDirectory == "" {
		log.Fatalf("$TEST_RESULT_REPORT_OUTPUT_DIRECTORY must be set")
	}

	err := os.MkdirAll(reportTestOutputDirectory, 0777)
	if err != nil {
		log.Fatalf("error making directory %s, err: %s", reportTestOutputDirectory, err)
	}
}
```

### DIFF
--- a/test/e2e/main_test.go
+++ b/test/e2e/main_test.go
@@ -46,7 +46,7 @@ func TestMain(m *testing.M) {
 	flag.Parse()
 
 	var err error
-	if testReportingFramework, err = reportingframework.New(*ns, *kubeconfig, *httpsAPI, *useKubeProxyForReportingAPI, *useRouteForReportingAPI, *routeBearerToken, *reportingAPIURL); err != nil {
+	if testReportingFramework, err = reportingframework.New(*ns, *kubeconfig, *httpsAPI, *useKubeProxyForReportingAPI, *useRouteForReportingAPI, *routeBearerToken, *reportingAPIURL, reportTestOutputDirectory); err != nil {
 		logrus.Fatalf("failed to setup framework: %v\n", err)
 	}
 
--- a/test/integration/main_test.go
+++ b/test/integration/main_test.go
@@ -251,7 +251,7 @@ func TestMain(m *testing.M) {
 	flag.Parse()
 
 	var err error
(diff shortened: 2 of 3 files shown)


chancez:
Maybe instead of doing it in both e2e main and integration main we should do it once in the framework?

chancez:
Or is the idea that we would have this come from main when we consolidate?

timflannagan1:
> Maybe instead of doing it in both e2e main and integration main we should do it once in the framework?

I need a way to be able to pass a directory hierarchy to the reportingframework that the deployframework has already build up:
```go
func testInstall(...) {
	// create a directory for this test case, e.g. /tmp/metering-tflannag-e2e/hdfsInstall
	path := filepath.Join(df.LoggingPath, testName)
	
	cfg, err := df.Setup(deployerConfig, path, targetPods)
	require.NoError(t, err, "Initializing the deploy framework should produce no error")

	...

	testReportingFramework, err := reportingframework.New(
		cfg.Namespace,
		cfg.KubeConfigPath,
		cfg.HTTPSAPI,
		cfg.UseKubeProxyForReportingAPI,
		cfg.UseRouteForReportingAPI,
		cfg.RouteBearerToken,
		cfg.ReportingAPIURL,
		cfg.ReportOutputDir,
	)
...
```
If I add a field to the `cfg` returned from the deploy framework constructor that initializes (but doesn't create) the reports directory (`filepath.Join(path, "reports")`) then I can pass that to the reporting framework constructor.

timflannagan1:
@chancez oh didn't see your second comment before replying - yes.

timflannagan1:
@chancez I'm not sure if main or the reporting framework should handle creating this directory as the deploy framework needs to teardown a failed install, so it's a preference of an empty `/reports` directory or a non-existent one.

chancez:
We could probably make the reporting framework error if it doesn't exist, and create it from the main/deployerFramework.

PR #974, opened 2019-10-09
test,hack: Refactor the e2e testing suite to use the deploy metering package.

timflannagan1:
### Overview:
This would refactor the e2e package to handle both deploying metering and run the reporting tests. This PR adds a deploy framework that's in charge of ensuring all of the metering resources are ready to proceed to run the reporting tests and teardown after the testing phase has completed.

### TODO:
- [ ] Update hack/ scripts to handle make e2e-local
- [ ] Log test output (junit reports)
- [ ] Avoid harcoding the reporting framework flags
- [x] Pass an overall testing directory to the deploy framework
- [ ] Log resource and containers logs to a tmpdir/specified path + random prefix

### Follow-up PRs:
- Pass clientsets to reportingframework.New instead of that method creating the clientsets explicitly


timflannagan1:
No CI tests on feature branch, so closing.

PR #976, opened 2019-10-11
bugfix: add missing comment clause to create table sql

beautytiger:


pkg/presto/util.go
```diff
@@ -149,7 +149,7 @@ func generateCreateTableSQL(catalog, schema, tableName string, columns []Column,
 )
 %s
 %s`
-	return fmt.Sprintf(sqlStr, ifNotExists, table, columnsStr, propsStr)
+	return fmt.Sprintf(sqlStr, ifNotExists, table, columnsStr, propsStr, comment)
```
chancez: comment comes before `propsStr` and we need to `prefix `COMMENT` to the message if set.

Eg:
```
commentStr := ""
if comment != "" {
  commentStr = "COMMENT " + comment
}

...
return fmt.Sprintf(sqlStr, ifNotExists, table, columnsStr, commentStr, propsStr)
```

beautytiger: Got it! Thank you very much.

beautytiger: Got it! Thank you very much.

pkg/presto/util.go
```diff
@@ -136,6 +136,11 @@ func generateCreateTableSQL(catalog, schema, tableName string, columns []Column,
 
 	columnsStr := generateColumnDefinitionListSQL(columns)
 
+	commentStr := ""
+	if len(comment) != 0 {
```
chancez: This can be simplified to `if comment != "" {`

PR #978, opened 2019-10-11
pkg/apis: Update the Hive object to use a pointer to the HiveSpec object in the MeteringConfig API.

timflannagan1:


timflannagan1:
We could also update the `MaxQueryRangeDuration` field to be of type `meta.Duration` too:

```go
ChunkSize:                 &meta.Duration{Duration: 5 * time.Minute},
PollInterval:              &meta.Duration{Duration: 30 * time.Second},
StepSize:                  &meta.Duration{Duration: 1 * time.Minute},
MaxImportBackfillDuration: &meta.Duration{Duration: 15 * time.Minute},
MaxQueryRangeDuration:     "5m",
```

chancez:
@timflannagan1 yeah let's do any updates all at once if we can. Also you could just include these changes in your other PR as a separate commit. Especially since we'll be using the feature branch and merges to master won't be making their way into it.

PR #979, opened 2019-10-11
test,hack: Refactor the e2e package to use the deploy metering package.

timflannagan1:
### Overview:
This would refactor the e2e package (`/test/e2e/`) to be responsible for both deploying metering and running the reporting tests against that deployment. This PR adds a `deployframework` package that's in charge of ensuring all of the metering resources are installed and ready to proceed to run the reporting tests, then outputs the resource and container logs to a directory before uninstalling the metering deployment after the testing phase has completed.

### TODO:
- [x] Support overriding the metering and reporting operator images.
- [x] Log test output for go junit reports.
- [x] Pass a base directory for test output to the e2e package.
- [x] Support creating a specified dir and a temp dir when unspecified.
- [x] Update `testInstall` to respect the defer closure call.
- [x] Update the teardown method to ensure `df.Deployer.Uninstall()` is ran.

### Follow-up PRs:
- Pass clientsets to `reportingframework.New` instead of that method creating the clientsets explicitly

### DIFF
--- a/hack/e2e.sh
+++ b/hack/e2e.sh
@@ -1,20 +1,32 @@
 #!/bin/bash
-set -e
 
 : "${KUBECONFIG:?}"
 
-ROOT_DIR=$(dirname "${BASH_SOURCE}")/..
+ROOT_DIR=$(dirname "${BASH_SOURCE[0]}")/..
 source "${ROOT_DIR}/hack/common.sh"
 source "${ROOT_DIR}/hack/lib/tests.sh"
 
+function remove_namespaces() {
+    echo "Removing namespaces with the 'name=metering-testing-ns' label"
+    kubectl delete ns -l "name=metering-testing-ns" || true
+}
+
+trap remove_namespaces SIGINT
+
 export METERING_NAMESPACE="${METERING_E2E_NAMESPACE:=${METERING_NAMESPACE}-e2e}"
+export METERING_SHORT_TESTS=${METERING_SHORT_TESTS:=false}
+export METERING_DEPLOY_MANIFESTS_PATH=${METERING_DEPLOY_MANIFESTS_PATH:=${ROOT_DIR}/manifests/deploy}
+export METERING_CLEANUP_SCRIPT_PATH=${METERING_CLEANUP_SCRIPT_PATH:=${ROOT_DIR}/hack/run-test-cleanup.sh}
+export TEST_OUTPUT_PATH=${TEST_OUTPUT_PATH:="$(mktemp -d)/${METERING_E2E_NAMESPACE}"}
+export TEST_LOG_LEVEL=${TEST_LOG_LEVEL:="debug"}
 
-export DEPLOY_SCRIPT="${DEPLOY_SCRIPT:-deploy-e2e.sh}"
-export TEST_SCRIPT="$ROOT_DIR/hack/run-e2e-tests.sh"
+TEST_OUTPUT_DIR="${TEST_OUTPUT_PATH}/tests"
+TEST_LOG_FILE="${TEST_LOG_FILE:-e2e-tests.log}"
+TEST_LOG_FILE_PATH="${TEST_LOG_FILE_PATH:-$TEST_OUTPUT_DIR/$TEST_LOG_FILE}"
+TEST_JUNIT_REPORT_FILE="${TEST_JUNIT_REPORT_FILE:-junit-e2e-tests.xml}"
+TEST_JUNIT_REPORT_FILE_PATH="${TEST_JUNIT_REPORT_FILE_PATH:-$TEST_OUTPUT_DIR/$TEST_JUNIT_REPORT_FILE}"
 
-export TEST_LOG_FILE="${TEST_LOG_FILE:-e2e-tests.log}"
-export DEPLOY_LOG_FILE="${DEPLOY_LOG_FILE:-e2e-deploy.log}"
-export TEST_TAP_FILE="${TEST_TAP_FILE:-e2e-tests.tap}"
+mkdir -p "$TEST_OUTPUT_DIR"
 
 echo "\$KUBECONFIG=$KUBECONFIG"
 echo "\$METERING_NAMESPACE=$METERING_NAMESPACE"
@@ -22,31 +34,28 @@ echo "\$METERING_OPERATOR_IMAGE_REPO=$METERING_OPERATOR_IMAGE_REPO"
 echo "\$REPORTING_OPERATOR_IMAGE_REPO=$REPORTING_OPERATOR_IMAGE_REPO"
 echo "\$METERING_OPERATOR_IMAGE_TAG=$METERING_OPERATOR_IMAGE_TAG"
 echo "\$REPORTING_OPERATOR_IMAGE_TAG=$REPORTING_OPERATOR_IMAGE_TAG"
+echo "\$TEST_OUTPUT_PATH=$TEST_OUTPUT_PATH"
 
-export DISABLE_PROMETHEUS_METRICS_IMPORTER=false
-if [ "$DEPLOY_REPORTING_OPERATOR_LOCAL" == "true" ]; then
-    export REPORTING_OPERATOR_PROMETHEUS_DATASOURCE_MAX_IMPORT_BACKFILL_DURATION='15m'
-    export REPORTING_OPERATOR_PROMETHEUS_METRICS_IMPORTER_INTERVAL="30s"
-    export REPORTING_OPERATOR_PROMETHEUS_METRICS_IMPORTER_CHUNK_SIZE="5m"
-    export REPORTING_OPERATOR_PROMETHEUS_METRICS_IMPORTER_INTERVAL="5m"
-    export REPORTING_OPERATOR_PROMETHEUS_METRICS_IMPORTER_STEP_SIZE="60s"
+set +e
+set +o pipefail
+set -x
+echo "Running E2E Tests"
 
-    export REPORTING_OPERATOR_API_LISTEN="127.0.0.1:8100"
-    export REPORTING_OPERATOR_METRICS_LISTEN="127.0.0.1:8101"
-    export REPORTING_OPERATOR_PPROF_LISTEN="127.0.0.1:8102"
+go test \
+    -test.short="${METERING_SHORT_TESTS}" \
+    -test.v \
+    -parallel 10 \
+    -timeout 30m \
+    "./test/e2e" \
+    -kubeconfig="${KUBECONFIG}" \
+    -namespace-prefix="${METERING_NAMESPACE}" \
+    -deploy-manifests-dir="${METERING_DEPLOY_MANIFESTS_PATH}" \
+    -cleanup-script-path="${METERING_CLEANUP_SCRIPT_PATH}" \
+    -test-output-path="${TEST_OUTPUT_PATH}" \
+    -log-level="${TEST_LOG_LEVEL}" \
+    |& tee "$TEST_LOG_FILE_PATH"
 
-    export METERING_PRESTO_PORT_FORWARD_PORT="8103"
-    export METERING_HIVE_PORT_FORWARD_PORT="8104"
-    export METERING_PROMETHEUS_PORT_FORWARD_PORT="8105"
-
-    export METERING_HTTPS_API="false"
-    export METERING_REPORTING_API_URL="http://$REPORTING_OPERATOR_API_LISTEN"
-    export METERING_USE_KUBE_PROXY_FOR_REPORTING_API="false"
-    export METERING_USE_ROUTE_FOR_REPORTING_API="false"
+# if go-junit-report is installed, create a junit report also
+if command -v go-junit-report >/dev/null 2>&1; then
+    go-junit-report < "$TEST_LOG_FILE_PATH" > "${TEST_JUNIT_REPORT_FILE_PATH}"
 fi
-
-if [ "$DEPLOY_METERING_OPERATOR_LOCAL" == "true" ]; then
-    export METERING_OPERATOR_CONTAINER_NAME="metering-operator-e2e"
-fi
-
-"$ROOT_DIR/hack/e2e-test-runner.sh"
--- a/test/deployframework/framework.go
+++ b/test/deployframework/framework.go
@@ -1,0 +1,87 @@
+package deployframework
+
+import (
+	"fmt"
+
+	"github.com/sirupsen/logrus"
+	apiextclientv1beta1 "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset/typed/apiextensions/v1beta1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/clientcmd"
+
+	meteringv1 "github.com/operator-framework/operator-metering/pkg/generated/clientset/versioned/typed/metering/v1"
+)
+
+const (
+	reportResultsDir                    = "report_results"
+	logDir                              = "logs"
+	meteringconfigDir                   = "meteringconfigs"
+	reportsDir                          = "reports"
+	datasourcesDir                      = "reportdatasources"
+	reportqueriesDir                    = "reportqueries"
+	hivetablesDir                       = "hivetables"
+	prestotablesDir                     = "prestotables"
+	storagelocationsDir                 = "storagelocations"
+	testNamespaceLabel                  = "metering-testing-ns"
+	meteringconfigMetadataName          = "operator-metering"
+	reportingOperatorServiceAccountName = "reporting-operator"
+	defaultPlatform                     = "openshift"
+	defaultDeleteNamespace              = true
+)
+
+// DeployFramework contains all the information necessary to deploy
+// different metering instances and run tests against them
+type DeployFramework struct {
+	NamespacePrefix   string
+	ManifestsDir      string
+	LoggingPath       string
+	CleanupScriptPath string
+	KubeConfigPath    string
+	Logger            logrus.FieldLogger
+	Client            kubernetes.Interface
+	APIExtClient      apiextclientv1beta1.CustomResourceDefinitionsGetter
+	MeteringClient    meteringv1.MeteringV1Interface
+}
+
+// New is the constructor function that creates and returns a new DeployFramework object
+func New(
+	logger logrus.FieldLogger,
+	nsPrefix,
+	manifestDir,
+	kubeconfig,
+	cleanupScriptPath,
+	loggingPath string,
+) (*DeployFramework, error) {
+	config, err := clientcmd.BuildConfigFromFlags("", kubeconfig)
+	if err != nil {
+		return nil, fmt.Errorf("failed to build a kube config from %s: %v", kubeconfig, err)
+	}
+
+	client, err := kubernetes.NewForConfig(config)
+	if err != nil {
+		return nil, fmt.Errorf("failed to initialize the k8s clientset: %v", err)
+	}
+
+	apiextClient, err := apiextclientv1beta1.NewForConfig(config)
+	if err != nil {
+		return nil, fmt.Errorf("failed to initialize the apiextensions clientset: %v", err)
+	}
+
+	meteringClient, err := meteringv1.NewForConfig(config)
+	if err != nil {
+		return nil, fmt.Errorf("failed to initialize the metering clientset: %v", err)
+	}
+
+	deployFramework := &DeployFramework{
+		NamespacePrefix:   nsPrefix,
+		ManifestsDir:      manifestDir,
+		CleanupScriptPath: cleanupScriptPath,
+		KubeConfigPath:    kubeconfig,
+		LoggingPath:       loggingPath,
+		Logger:            logger,
+		Client:            client,
+		APIExtClient:      apiextClient,
+		MeteringClient:    meteringClient,
+	}
+
+	return deployFramework, nil
+}
--- a/test/deployframework/helpers.go
+++ b/test/deployframework/helpers.go
@@ -1,0 +1,69 @@
+package deployframework
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+
+	"github.com/sirupsen/logrus"
+	v1 "k8s.io/api/core/v1"
+)
+
+func checkPodStatus(pod v1.Pod) (bool, int) {
+	if pod.Status.Phase != v1.PodRunning {
+		return false, 0
+	}
+
+	var unreadyContainers int
+
+	for _, status := range pod.Status.ContainerStatuses {
+		if !status.Ready {
+			unreadyContainers++
+		}
+	}
+
+	return unreadyContainers == 0, len(pod.Status.ContainerStatuses) - unreadyContainers
+}
+
+func createResourceDirs(namespace, path string) ([]string, error) {
+	envVarArr := []string{
+		"METERING_TEST_NAMESPACE=" + namespace,
+	}
+
+	testDirsMap := map[string]string{
+		logDir:              "LOG_DIR",
+		reportsDir:          "REPORTS_DIR",
+		meteringconfigDir:   "METERINGCONFIGS_DIR",
+		datasourcesDir:      "DATASOURCES_DIR",
+		reportqueriesDir:    "REPORTQUERIES_DIR",
+		hivetablesDir:       "HIVETABLES_DIR",
+		prestotablesDir:     "PRESTOTABLES_DIR",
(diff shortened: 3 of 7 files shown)


timflannagan1:
I think I covered the main functionality of the existing e2e process. The only thing that I think needs major work is the `testhelpers.NewMeteringConfig(...)` function as most of those values are hardcoded (storage, resource limits, images).

Not sure what the best approach is for that function, as we'll have a better idea of what it's going to look like once we add more `MeteringConfig` test cases.

chancez:
> Not sure what the best approach is for that function, as we'll have a better idea of what it's going to look like once we add more MeteringConfig test cases.


probably not using a function but just defining struct literals is best honestly.

timflannagan1:
@chancez I agree, I'll update to use a literal - I was mainly worried about the length of the test table and the general readability of the `TestMultipleInstalls` function.

chancez:
@timflannagan1 We can also define the variables separately from the test tables and then just use them as variables.

test/deployframework/framework.go
```diff
@@ -0,0 +1,335 @@
...
+		return nil, fmt.Errorf("failed to add the testing label to the %s namespace", namespace)
+	}
+
+	_, err = df.WaitForMeteringPods(targetPods, cfg.Namespace)
+	if err != nil {
+		return nil, fmt.Errorf("error waiting for metering pods to become ready: %v", err)
+	}
+
+	routeBearerToken, err := df.GetRouteBearerToken(cfg.Namespace)
+	if err != nil {
```
EmilyM1: Where is the deployframework getting the BearerToken from?

test/deployframework/framework.go
```diff
@@ -0,0 +1,335 @@
...
+
+	return reportingFrameworkConfig, nil
+}
+
+// Teardown is a method that creates the resource and container logging
+// directories, then populates those directories by executing the
+// df.CleanupScriptPath bash script, while streaming the script output
+// to stdout. Once the cleanup script has finished execution, we can
+// uninstall the metering stack and return an error if there is any.
+func (df *DeployFramework) Teardown(pathToTestCaseOutputDir string) error {
```
EmilyM1: Great comments 👍

test/deployframework/framework.go
```diff
@@ -0,0 +1,335 @@
...
+
+	df.Logger.Infof("Installing metering took %v", time.Since(start))
+
+	return true, nil
+}
+
+// GetRouteBearerToken queries the @namespace for the reporting-operator service account and attempts
+// to find the secret that contains the reporting-operator token. If that secret exists, return the
+// string representation of the token (key) byte slice (value), or return an error.
+func (df *DeployFramework) GetRouteBearerToken(namespace string) (string, error) {
```
EmilyM1: Minor nit: This is referenced in line 159, so maybe throwing this function more towards the beginning of the script would keep it clearer

hack/e2e.sh
```diff
@@ -1,52 +1,49 @@
...
-fi
+set +e
+set +o pipefail
+echo "Running E2E Tests"
+"${ROOT_DIR}/hack/$TEST_HACK_SCRIPT" 2>&1 | tee "$TEST_LOG_FILE_PATH"
 
-if [ "$DEPLOY_METERING_OPERATOR_LOCAL" == "true" ]; then
-    export METERING_OPERATOR_CONTAINER_NAME="metering-operator-e2e"
+# attempt to create a go junit report if the package is already installed
+if [[ -f "$GOPATH"/bin/go-junit-report ]]; then
```
chancez: Can you use the logic from `e2e-test-runner.sh` instead of this new logic for detecting go-junit-report? Previously we just looked for `go-junit-report` was in your PATH using `command -v go-junit-report`

hack/run-go-test.sh
```diff
@@ -0,0 +1,17 @@
+#! /bin/bash
```
chancez: `run-go-test.sh` is a bit too vague for what this file does. Can you pick a new name that's more representative like `run-e2e.sh` or something. Also it seems like this script could just be inlined into `e2e.sh`, not really sure we need it to be a separate script.

timflannagan1: The only reason that this needs to be its own standalone script is to capture the go test flags used.

I agree the name was poorly chosen and unfortunately `run-e2e-tests.sh` would be the more descriptive name. I'll update the name, and maybe we can rename this after integration is refactored too.

test/deployframework/framework.go
```diff
@@ -0,0 +1,335 @@
...
+	}
+
+	return deployFramework, nil
+}
+
+// Setup handles the process of deploying metering, and waiting for all necessary resources
+// to become ready in order to proceed with running the reporting tests.
+func (df *DeployFramework) Setup(cfg deploy.Config, namespace, testOutputPath string, targetPods int) (*ReportingFrameworkConfig, error) {
+	var err error
+
```
chancez: There's a lot of mutation of `df` happening during `Setup` that I think could be avoided by doing more of the setup during `New`.  Anything assigning to `df` or `cfg` in here I would consider something which we want to avoid in general if we can. 

Partially initialized objects is one issue, as it can be harder to test and understand in general. Additionally, this makes it possible to use incorrectly, as you could call teardown before setup for example, and teardown might be expecting some of these values to have been set during `Setup`. This might point to the idea that `Setup` should return something we can use for Teardown, but I'm not 100% convinced we need that.

Thing I think we can do:
- Can we try to initialize the `DeployFramework` as much as possible before `Setup` or `Teardown` is called?
- Avoid assigning to `deploy.Config` during setup. Instead I think it would be best if `deploy.Config` was automatically created within `Setup`, not passed as an argument. 90% of the values for `deploy.Config` do not need to be specified on a per-test case basis, only MeteringConfig does, and not even that, we only actually need MeteringConfig spec (this way we're not dealing with the namespace being set/unset, we just create the MeteringConfig with it set). So perhaps you should be creating a MeteringConfig given a MeteringConfigSpec, and then creating a DeployConfig using the MeteringConfig object you create.

test/deployframework/framework.go
```diff
@@ -0,0 +1,335 @@
...
+	df.Config.Namespace = namespace
+	cfg.Namespace = namespace
+	cfg.MeteringConfig.ObjectMeta = meta.ObjectMeta{
+		Name:      meteringconfigMetadataName,
+		Namespace: namespace,
+	}
+
+	df.Logger.Debugf("Deployer config: %+v", cfg)
+
+	df.Deployer, err = deploy.NewDeployer(cfg, df.Logger, df.Client, df.APIExtClient, df.MeteringClient)
```
chancez: The only issue I see with my comment from above is how we keep track of the `Deployer` object and use it between Install and Uninstall. This is the reason I think we may need another type of object which is returned by `Setup()` which you could call Teardown on, this way `Deployer` isn't a field on `DeployFramework` since that's more of an intermediate object/state.

For example:
```
df := deployFramework.New(...)
deployCtrl, err := df.Setup()
if err != nil { ... }
defer deployCtrl.Teardown()
reportingFramework := deployCtrl.ReportingFramework
runReportTests(reportingFramework)
```

test/deployframework/framework.go
```diff
@@ -0,0 +1,335 @@
...
+
+	routeBearerToken, err := df.GetRouteBearerToken(cfg.Namespace)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get the route bearer token: %v", err)
+	}
+
+	reportResultsPath := filepath.Join(testOutputPath, reportResultsDir)
+	err = os.Mkdir(reportResultsPath, 0777)
+	if err != nil {
+		df.Logger.Fatalf("Failed to make the directory %s: %v", reportResultsPath, err)
```
chancez: Never log fatal outside of `main()`. Return errors.

test/deployframework/framework.go
```diff
@@ -0,0 +1,335 @@
...
+// directories, then populates those directories by executing the
+// df.CleanupScriptPath bash script, while streaming the script output
+// to stdout. Once the cleanup script has finished execution, we can
+// uninstall the metering stack and return an error if there is any.
+func (df *DeployFramework) Teardown(pathToTestCaseOutputDir string) error {
+	logger := df.Logger.WithFields(logrus.Fields{"component": "cleanup"})
+
+	envVarArr, err := df.createResourceDirs(pathToTestCaseOutputDir)
+	if err != nil {
+		logger.Warnf("failed to create the resource output directories: %v", err)
```
chancez: Probably return the error

test/deployframework/framework.go
```diff
@@ -0,0 +1,335 @@
...
+	}
+
+	cleanupCmd := exec.Command(df.CleanupScriptPath)
+	cleanupStdout, err := cleanupCmd.StdoutPipe()
+	if err != nil {
+		logger.Warnf("Failed to create pipe from command output to Stdout: %v", err)
```
chancez: Probably return the error

test/deployframework/framework.go
```diff
@@ -0,0 +1,335 @@
...
+	}
+
+	scanner := bufio.NewScanner(cleanupStdout)
+	go func() {
+		for scanner.Scan() {
+			line := scanner.Text()
+			logger.Infof(line)
+		}
+		if err := scanner.Err(); err != nil {
+			logger.Errorf("error reading output from command: %v", err)
```
chancez: Probably return the error

test/deployframework/framework.go
```diff
@@ -0,0 +1,335 @@
...
+		}
+		return
+	}()
+
+	cleanupCmd.Env = append(os.Environ(), envVarArr...)
+	err = cleanupCmd.Run()
+	if err != nil {
+		logger.Warnf("error running the cleanup script: %v", err)
```
chancez: Probably return the error

test/deployframework/framework.go
chancez: This isn't really that good of advice since usually functions can live in separate files from where they're used, so where it is in a file shouldn't really matter.

test/deployframework/framework.go
```diff
@@ -0,0 +1,335 @@
...
+
+	if len(sa.Secrets) == 0 {
+		return "", fmt.Errorf("failed to return a list of secrets in the reporting-operator service account")
+	}
+
+	var secretName string
+
+	for _, secret := range sa.Secrets {
+		if strings.Contains(secret.Name, "token") {
+			secretName = secret.Name
```
chancez: `break` from the loop when you find the item you want.

test/deployframework/helpers.go
```diff
@@ -0,0 +1,104 @@
...
+	for _, status := range pod.Status.ContainerStatuses {
+		if !status.Ready {
+			unreadyContainers++
+		}
+	}
+
+	return unreadyContainers == 0, len(pod.Status.ContainerStatuses) - unreadyContainers
+}
+
+func (df *DeployFramework) addTestingNamespaceLabel(namespace, labelVal string) error {
```
chancez: Ideally we could just set this when we create the namespace we're using for the test.

test/e2e/report_test.go
```diff
@@ -7,88 +7,93 @@ import (
...
-					if dataSource.Status.PrometheusMetricsImportStatus != nil && dataSource.Status.PrometheusMetricsImportStatus.EarliestImportedMetricTime != nil {
-						// keep the EarliestImportedMetricTime that is the
-						// least far back, so that we ensure the reportStart is
-						// a time that all datasources have metrics for.
-						if reportStart.IsZero() || dataSource.Status.PrometheusMetricsImportStatus.EarliestImportedMetricTime.Time.After(reportStart) {
-							reportStart = dataSource.Status.PrometheusMetricsImportStatus.EarliestImportedMetricTime.Time
+	// note: we need to run the testCase table cases as a group in order to avoid
+	// preemptively running the defer closure in main_test.go early and uninstalling
+	// the metering resources during the middle of the test execution
+	t.Run("group", func(t *testing.T) {
```
chancez: This still needs to be removed

hack/run-go-test.sh
chancez: Not sure I understand the "to capture the flags used" part. Can you elaborate?

test/deployframework/framework.go
timflannagan1: If I return an error here, the `df.Deployer.Uninstall()` is not run.

test/deployframework/framework.go
timflannagan1: whoops, good call

test/deployframework/framework.go
chancez: Okay, does the cleanup script regularly error then or what?

test/deployframework/framework.go
chancez: Note, if we remove `deploy.Config` as an argument, we will also want to take the reporting/metering operator image repo/tag values as arguments so that we can override those still, and we'll use those to create the `deploy.Config` within Setup().

test/deployframework/framework.go
timflannagan1: So I can either update this method to be purely focused on logging, and move the `df.Deployer.Uninstall()` to the defer closure, e.g.:
```go
	defer func() {
		err := df.Teardown(testOutputDir)

		err := df.Deployer.Uninstall()
		assert.NoError(t, err, ...)
	}()
```

hack/run-go-test.sh
timflannagan1: @chancez if inlined the multi-line go test command into `./hack/e2e.sh`, then the `tee` command doesn't log any of the multi-line go test command (so none of the values passed to the go test flags are logged), which seems like poor behavior.

test/deployframework/framework.go
timflannagan1: @chancez no the cleanup script doesn't error often but was something that I encountered while developing this and placed more value on the resources being torn down properly.

hack/run-go-test.sh
timflannagan1: More concretely:
```bash
└─ $ ▶ bat /tmp/tmp.YDEJch1xed/metering-tflannag-e2e/tests/e2e-tests.log 
───────┬──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
       │ File: /tmp/tmp.YDEJch1xed/metering-tflannag-e2e/tests/e2e-tests.log
───────┼──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
   1   │ time="10-14-2019 18:04:42" level=info msg="Setting the log level to debug" app=deploy
   2   │ === RUN   TestMultipleInstalls
   3   │ === RUN   TestMultipleInstalls/HDFSInstall
   4   │ time="10-14-2019 18:04:42" level=debug msg="Deployer config: {SkipMeteringDeployment:false DeleteCRDs:false DeleteCRB:false DeleteNamespace:true DeletePVCs:false DeleteAll:false Namespace:metering-tflannag-e2e-7 Platform:openshift Repo: Tag: OperatorResources:0xc000137900 MeteringConfig:0xc00000d2c0}" app=de
       │ ploy
   5   │ time="10-14-2019 18:04:42" level=info msg="Metering Deploy Namespace: metering-tflannag-e2e-7" app=deploy
   6   │ time="10-14-2019 18:04:42" level=info msg="Metering Deploy Platform: openshift" app=deploy
   7   │ time="10-14-2019 18:04:42" level=info msg="Labeling the metering-tflannag-e2e-7 namespace with 'openshift.io/cluster-monitoring=true'" app=deploy
   8   │ time="10-14-2019 18:04:42" level=info msg="Created the metering-tflannag-e2e-7 namespace" app=deploy
   9   │ time="10-14-2019 18:04:42" level=info msg="Updated the hivetables.metering.openshift.io CRD" app=deploy
  10   │ time="10-14-2019 18:04:42" level=info msg="Updated the prestotables.metering.openshift.io CRD" app=deploy
  11   │ time="10-14-2019 18:04:42" level=info msg="Updated the storagelocations.metering.openshift.io CRD" app=deploy
  12   │ time="10-14-2019 18:04:42" level=info msg="Updated the reports.metering.openshift.io CRD" app=deploy
  13   │ time="10-14-2019 18:04:42" level=info msg="Updated the reportqueries.metering.openshift.io CRD" app=deploy
  14   │ time="10-14-2019 18:04:42" level=info msg="Updated the reportdatasources.metering.openshift.io CRD" app=deploy
  15   │ time="10-14-2019 18:04:43" level=info msg="Updated the meteringconfigs.metering.openshift.io CRD" app=deploy
  16   │ time="10-14-2019 18:04:43" level=info msg="Created the metering deployment" app=deploy
  17   │ time="10-14-2019 18:04:43" level=info msg="Created the metering serviceaccount" app=deploy
  18   │ time="10-14-2019 18:04:43" level=info msg="Created the metering role" app=deploy
  19   │ time="10-14-2019 18:04:43" level=info msg="Created the metering role binding" app=deploy
  20   │ time="10-14-2019 18:04:43" level=info msg="Created the metering cluster role" app=deploy
  21   │ time="10-14-2019 18:04:43" level=info msg="Created the metering cluster role binding" app=deploy
  22   │ time="10-14-2019 18:04:43" level=info msg="Created the MeteringConfig resource" app=deploy
  23   │ time="10-14-2019 18:04:43" level=info msg="Updated the 'name=metering-testing-ns' label to the metering-tflannag-e2e-7 namespace" app=deploy
  24   │ time="10-14-2019 18:04:43" level=info msg="Waiting for all metering pods to be ready" app=deploy
  25   │ time="10-14-2019 18:04:53" level=info msg="Poll Summary" app=deploy
  26   │ time="10-14-2019 18:04:53" level=info msg="Current ratio of ready to target pods: 1/7" app=deploy
  27   │ time="10-14-2019 18:05:03" level=info msg="Poll Summary" app=deploy
  28   │ time="10-14-2019 18:05:03" level=info msg="Current ratio of ready to target pods: 1/7" app=deploy
```

timflannagan1: In comparison to what's current in this PR:
```bash
└─ $ ▶ bat /tmp/tmp.jP958XLTky/metering-tflannag-e2e/tests/e2e-tests.log 
───────┬──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
       │ File: /tmp/tmp.jP958XLTky/metering-tflannag-e2e/tests/e2e-tests.log
───────┼──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
   1   │ + go test -test.short=false -test.v -parallel 10 -timeout 30m /home/tflannag/go/src/github.com/operator-framework/operator-metering/test/e2e -kubeconfig=/home/tflannag/openshift-dev/auth/kubeconfig -namespace-prefix=metering-tflannag-e2e -deploy-manifests-dir=/home/tflannag/go/src/github.com/operator-framewo
       │ rk/operator-metering/manifests/deploy -cleanup-script-path=/home/tflannag/go/src/github.com/operator-framework/operator-metering/hack/run-test-cleanup.sh -test-output-path=/tmp/tmp.jP958XLTky/metering-tflannag-e2e -log-level=debug
   2   │ time="10-14-2019 18:06:53" level=info msg="Setting the log level to debug" app=deploy
   3   │ === RUN   TestMultipleInstalls
   4   │ === RUN   TestMultipleInstalls/HDFSInstall
   5   │ time="10-14-2019 18:06:53" level=debug msg="Deployer config: {SkipMeteringDeployment:false DeleteCRDs:false DeleteCRB:false DeleteNamespace:true DeletePVCs:false DeleteAll:false Namespace:metering-tflannag-e2e-38 Platform:openshift Repo: Tag: OperatorResources:0xc0002e19a0 MeteringConfig:0xc00000cf00}" app=d
       │ eploy
```

test/deployframework/framework.go
chancez: I'm just wondering, what if we consider the script to be very important to run too. I would assume we want to know that. Perhaps we should accumulate errors in a slice, then run uninstall, and if there's any errors in the slice, return the list of errors joined together.

hack/run-go-test.sh
chancez: That can be accomplished by putting `set -x` before running Go test.

timflannagan1: I originally had `set -x` set before running the `go test ...` portion, and still wasn't able to get that to work. 

Maybe I need to do something different with the file descriptors, but I've tried a couple of different combinations and haven't been able to find any references that achieve this behavior.

test/deployframework/helpers.go
timflannagan1: The namespace is created using the deploy package so that would involve altering that package to support passing a label to the `deployer.installNamespace()` method, but that method is also private (so `deployer.Install()`).

hack/run-go-test.sh
chancez: `set -x` is what does this, and it doesn't need a separate file for it to work, it's just printing the shell expressions as they're executed. It doesn't have anything to do with files or file descriptors.

test/deployframework/helpers.go
chancez: I see. I think it would be ideal if we could set extra labels by passing a map to the deployer.New constructor or deployer config.

hack/run-go-test.sh
timflannagan1: I was under the impresison that `set -x` writes to stderr (which is a file descriptor).

chancez: Ah, that might be true, though I don't necessarily think that's the issue here. I do see in `e2e.sh` you put `set +x` though.

test/e2e/main_test.go
```diff
@@ -3,57 +3,245 @@ package e2e
...
+				ReportingOperator: &metering.ReportingOperator{
+					Spec: &metering.ReportingOperatorSpec{
+						Resources: &v1.ResourceRequirements{
+							Requests: v1.ResourceList{
+								v1.ResourceCPU:    resource.MustParse("1"),
+								v1.ResourceMemory: resource.MustParse("250Mi"),
+							},
+						},
+						Image: &metering.ImageConfig{
+							Repository: reportingOperatorImageRepo,
```
chancez: As a follow up, we need to make sure these are always correctly set. This might be a case where we have to force it set Setup() or something like you had done for namespace cause it'll probably be too much to extract out all these sub-configs.

test/e2e/main_test.go
```diff
@@ -3,57 +3,245 @@ package e2e
...
+		t := t
+		testCase := testCase
+
+		t.Run(testCase.Name, func(t *testing.T) {
+			testInstall(t, testCase.MeteringOperatorImageRepo, testCase.MeteringOperatorImageTag, testCase.Name, testCase.TargetPods, testCase.MeteringConfigSpec)
+		})
+	}
+}
+
+func testInstall(
```
chancez: We'll want to figure out a better name for this, but I also think it'll change a bit once it needs to run other sub-tests besides `testReportingProducesData`

test/e2e/main_test.go
```diff
@@ -3,57 +3,245 @@ package e2e
...
+		*testOutputPathFlag,
+	); err != nil {
+		logger.Fatalf("Failed to create a new deploy framework: %v", err)
 	}
 
 	os.Exit(m.Run())
 }
 
-func TestReportingProducesData(t *testing.T) {
+func TestMultipleInstalls(t *testing.T) {
```
chancez: For now, let's just call this `TestInstallMeteringAndReportingProducesData`

test/e2e/main_test.go
```diff
@@ -3,57 +3,245 @@ package e2e
...
+	}()
+
+	// note: in order to respect the defer closure and cleanup responsibly, we need
+	// to avoid returning errors or ending the function early as testInstall could
+	// be run in a goroutine - so check if the cfg object is nil to avoid a panic
+	if rf != nil {
+		// note: we need to run the testReportingProducesData as a group in order
+		// to avoid preemptively running the defer closure early and uninstalling
+		// the metering resources during the middle of the test execution
+		t.Run("group", func(t *testing.T) {
```
chancez: change `group` to `testReportingProducesData` would be best i think.

test/e2e/main_test.go
timflannagan1: Proposed fix: https://github.com/operator-framework/operator-metering/pull/994

PR #981, opened 2019-10-14
test/reportingframework: Stat the report output directory.

timflannagan1:
This would add a validation check that the path to the intended report output directory that's passed to the reportingframework constructor is a real directory. 

This is nice to have as metering needs to be deployed before running any tests, so waiting for that stack to become ready, and failing the reporting tests due to an empty/invalid `reportOutputDir` is a waste of time/resources.

### DIFF
--- a/test/reportingframework/framework.go
+++ b/test/reportingframework/framework.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"net/http"
 	"net/url"
+	"os"
 	"sync"
 	"time"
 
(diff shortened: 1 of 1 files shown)


timflannagan1:
Could probably update this to check if the path is `os.IsDir` and differentiate between a stat error and `os.IsNotExist(err)`.

chancez:
@timflannagan1 Can you describe the purpose of this change in your commit message and PR descriptions?

timflannagan1:
@chancez updated both

PR #984, opened 2019-10-15
pkg/deploy: Support passing extra namespace labels for use in testing.

timflannagan1:
This would add the functionality to pass extra namespace labels, which can be used in the e2e/integration refactoring for easier cleanup, e.g.
```go
cfg.ExtraNamespaceLabels = map[string]string{"name": "metering-testing-ns"}
```
This is useful in the case where there's existing namespaces with the label defined above that you want to cleanup, e.g.:
```bash
kubectl delete namespace -l name=metering-testing-ns
```

### DIFF
--- a/pkg/operator/deploy/deploy.go
+++ b/pkg/operator/deploy/deploy.go
@@ -64,6 +64,7 @@ type Config struct {
 	Platform               string
 	Repo                   string
 	Tag                    string
+	ExtraNamespaceLabels   map[string]string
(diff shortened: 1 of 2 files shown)


timflannagan1:
It's important to note that this would only add extra namespaces to namespaces that don't already exist (like in the current testing structure) and does not update pre-existing namespaces, as I wasn't sure if this situation would need to be supported.

PR #985, opened 2019-10-16
images: Update the HDFS pod security to run as root user.

timflannagan1:
When installing the upstream metering package, the `hdfs-datanode-*` pod is unable to report a ready status. 

Updating the `spec.hadoop.spec.hdfs.securityContext` field in the `_ocp_disabled_overrides` dictionary needs to be updated in order for the pod to be run as a root user.


chancez:
I'm not sure this is necessary. I believe this only matters for hostPath volumes, in which case the user can override this themselves.

PR #987, opened 2019-10-16
hack: Change junit report file name for e2e/integration tests

chancez:


timflannagan1:
just in case this doesn't get merged before checking if it worked
/hold

PR #989, opened 2019-10-17
charts/openshift-metering: Support annotations on service accounts

### LINKED ISSUE #947: Support IAM Roles for Service Accounts
Amazon recently released support for assigning IAM roles to Service Accounts (https://aws.amazon.com/blogs/opensource/introducing-fine-grained-iam-roles-service-accounts/). 

In order to take advantage of this, the AWS SDK needs to be upgraded (The Go SDK needs to be a minimum of version 1.23.13), and we need a way to indicate the IAM role to be used, so it can be added as an annotation to the service account(s) used by the operator.

Assigning IAM roles through this method is easier and more secure than keeping long-running IAM users around, which have fixed credentials and [ideally] need them rotated periodically.

chancez:
This re-uses spec.annotations for each component that are also set on
the pod template annotations to allow additionally setting annotations
on the pod service accounts.

Also adds app labels to each service account.

Partial fix for #947

### DIFF
--- a/charts/openshift-metering/templates/hadoop/hdfs-serviceaccount.yaml
+++ b/charts/openshift-metering/templates/hadoop/hdfs-serviceaccount.yaml
@@ -3,4 +3,8 @@ kind: ServiceAccount
 metadata:
   name: hdfs
   labels:
-      app: hdfs
+    app: hdfs
+{{- if .Values.hadoop.spec.hdfs.namenode.annotations }}
+  annotations:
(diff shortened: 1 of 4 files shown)


chancez:
/hold
@SapientGuardian does this look like what you need?

SapientGuardian:
Yeah that should cover the annotations. There were a couple other customizations I had to do in the ansible image to complete the picture (setting fsGroup: 3001 in the reporting operator spec and setting hive.s3.use-instance-credentials=false in the presto config) but as far as the annotations go, this should cover it.

PR #990, opened 2019-10-17
charts,manifests: Support configuring reporting-operator securityContext

chancez:
Partial fix for #947


chancez:
I need to find time to manually test this out.

chancez:
/hold cancel
Works as expected for controlling the securityContext values on reporting-operator.

PR #992, opened 2019-10-17
hack,test,Makefile: Consolidate the e2e and integration testing suites.

timflannagan1:
### Overview
With the addition of the `deployframework` package, which allows us to deploy metering inside of the Go test code, we can consolidate the integration and e2e packages into one package, and remove the deploy and test wrapper hack scripts.

This PR is part of the overall effort towards refactoring our current testing suites and removing any of the now unneeded bash hack scripts. Before, the e2e and integration packages shared a lot of redundancies and both packages tested e2e behavior.

### TODO:
- [x] Remove the cliensets from the `reportingframework` (pass them to constructor)
- [x] Ensure the reporting operator images are overrided properly
- [x] Support passing a list of test functions to the test table
- [x] Reorganize the package
- [x] Support skipping a list of test functions
- [ ] Verify the error message resembles the returned error message (when `ExpectErr: true`)


test/e2e/main_test.go
```diff
@@ -33,6 +38,207 @@ var (
 	reportingOperatorImageTag  string
 
 	defaultTargetPods = 7
+
+	reportTestTimeout         = 5 * time.Minute
+	reportTestOutputDirectory string
+
+	testReportsProduceCorrectDataForInputTestCases = []reportsProduceCorrectDataForInputTestCase{
```
chancez: I think this probably deserves it's own file, and probably next to the testReportsProduceCorrectDataForInput function. Can we add a new file `test/e2e/reports_produce_correct_data_test.go` and move `testReportsProduceCorrectDataForInput` there and `testReportsProduceCorrectDataForInputTestCases` there?

chancez:
Panic in the tests.
- https://storage.googleapis.com/origin-ci-test/pr-logs/pull/operator-framework_operator-metering/992/pull-ci-operator-framework-operator-metering-feature-metering-e2e-aws/75/artifacts/metering-e2e-aws/container-logs/test.log
- Your failing install test has a pointer derefence here that's causing it     `err := validateImageConfig(*cfg.MeteringConfig.Spec.ReportingOperator.Spec.Image)`
- This should have failed the tests, but it didn't. I'm guessing we're getting a 0 exit code from the commands in the TRAP. We need to capture the exit code of the command prior to the TRAP running, then exit using that code after the trap to preserve it.

test/deployframework/context.go
```diff
@@ -71,13 +73,18 @@ func (df *DeployFramework) NewDeployerCtx(
 
 	// validate the reporting-operator image is non-empty when overrided
 	if df.ReportingOperatorImageRepo != "" || df.ReportingOperatorImageTag != "" {
-		err := validateImageConfig(*cfg.MeteringConfig.Spec.ReportingOperator.Spec.Image)
+		reportingOperatorImageConfig := &metering.ImageConfig{
+			Repository: df.ReportingOperatorImageRepo,
+			Tag:        df.ReportingOperatorImageTag,
+		}
+
+		err := validateImageConfig(*reportingOperatorImageConfig)
```
timflannagan1: @chancez I think it may make more sense to build up an `metering.ImageConfig` object with the deployframework reporting-operator repo/tag fields, and validating that object instead of directly validating the `cfg.MeteringConfig...ReportingOperator.Spec.Image` object and having to deal with null checking nested structure fields (like we do with the metering-operator image).

In this case, we can error if the overridden image was improperly managed, which means that metering won't be installed.

test/e2e/main_test.go
```diff
@@ -208,109 +240,56 @@ func testInstall(
...
-	assert.NoError(t, err, "deploying metering should produce no error")
-
-	defer func() {
-		err := deployerCtx.Teardown(df.CleanupScriptPath)
-		assert.NoError(t, err, "capturing logs and uninstalling metering should produce no error")
-	}()
+	for _, testFunction := range testInstallFunctions {
+		if testFunction.Skip {
+			continue
+		}
```
timflannagan1: We can simply skip the loop iteration in the case where we want to skip a test function, e.g.
```go
			InstallSubTests: []InstallTestCase{
				{
					Name:     "testReportingProducesData",
					TestFunc: testReportingProducesData,
					Skip: true,
				},
				{
					Name:     "testReportingProducesCorrectDataForInput",
					TestFunc: testReportingProducesCorrectDataForInput,
					Skip: true,				
				},
			},
```

In the current implementation, I'm not sure how useful this would be but in the future, we could control this via a go test flag for specific tests:
```go
skipTestReportingProducesData := os.Getenv("DISABLE_TEST_REPORTING_PRODUCES_DATA")
...
{
	Name:     "testReportingProducesData",
	TestFunc: testReportingProducesData,
	Skip: skipTestReportingProducesData,
}
```

It may be more useful to be able to skip certain `MeteringConfig` configurations, like installing metering using s3 for storage, than certain, specific test functions.

chancez: Yeah, there's no reason to skip individual tests like this except for when it's about being able to run the tests, rather than wanting to. For example, like using S3, it's disabled if there's no creds for it, or something.

timflannagan1: Alright, removed.

test/e2e/main_test.go
```diff
@@ -212,112 +243,65 @@ func testInstall(
...
 
-	// randomize the namespace to avoid existing namespaces
 	rand.Seed(time.Now().UnixNano())
-	namespace := namespacePrefix + "-" + strconv.Itoa(rand.Intn(50))
 
-	deployerCtx, err := df.NewDeployerCtx(namespace, meteringOperatorImageRepo, meteringOperatorImageTag, reportingOperatorImageRepo, reportingOperatorImageTag, spec, testOutputDir, targetPods)
-	assert.NoError(t, err, "creating a new deployer context should produce no error")
+	for _, testFunction := range testInstallFunctions {
+		// randomize the namespace to avoid existing namespaces
+		namespace := namespacePrefix + "-" + strconv.Itoa(rand.Intn(50))
```
chancez: I believe we include the testcase name here in some way as it is we can still get conflicts and it's only 50 unique combos. Maybe `namespace := fmt.Sprintf("%s-%s-%d", namespacePrefix, testCaseName, rand.Intn(50))` We may need to truncate/limit it to `63` characters as that's the limit on kube NS names.

test/e2e/main_test.go
```diff
@@ -78,19 +78,50 @@ func TestMain(m *testing.M) {
...
 			TargetPods:                defaultTargetPods,
 			MeteringOperatorImageRepo: meteringOperatorImageRepo,
 			MeteringOperatorImageTag:  meteringOperatorImageTag,
+			InstallSubTests: []InstallTestCase{
+				{
+					Name:     "testReportingProducesData",
+					TestFunc: testReportingProducesData,
+				},
+				{
+					Name:     "testReportingProducesCorrectDataForInput",
```
chancez: Im realizing this won't work exactly because `MeteringConfigSpec` needs to be different between these two cases, as `testReportingProducesCorrectDataForInput` should disable metrics importing for reporting-operator. I think the concept of `InstallSubTests` still makes sense, so we can leave it, but we need to move this to it's own parent test case with it's own meteringconfig.

test/e2e/main_test.go
timflannagan1: Good point - we can first try with the combination above, and if the length of that string exceeds the kube ns limit, retry with the original implementation.

I took a look at the kube namespace validation regex, and it only allows lowercase characters and numerics.

test/e2e/main_test.go
```diff
@@ -78,19 +80,46 @@ func TestMain(m *testing.M) {
...
 		{
-			Name:                      "HDFSInstall",
+			Name:                      "DefaultMeteringConfigProducesError",
 			TargetPods:                defaultTargetPods,
 			MeteringOperatorImageRepo: meteringOperatorImageRepo,
 			MeteringOperatorImageTag:  meteringOperatorImageTag,
+			InstallSubTests: []InstallTestCase{
+				{
+					Name:      "testReportingProducesData",
+					TestFunc:  testReportingProducesData,
```
chancez: I think this TestCase doesn't really make a lot of sense. Should `ExpectErr` maybe be outside of InstallSubTests`? Also I think we need a way to verify what the error is. Like we should maybe change `ExpectErr` to a string which we check exists inside the error returned by install.

chancez: It's also semi-weird that ExpectErr` is in `InstallTestCase` because the install failing is mostly about the MeteringConfig and will never differ between different InstallTestCases. This makes me believe that maybe we need to make `InstallSubTests` into a non-slice and move ExpectErr out.

timflannagan1: Depending on the implementation of `InstallSubTests`, we can omit this section entirely and just pass an empty `MeteringConfig.Spec` object, e.g.
```go
rf, err := deployerCtx.Setup()
if testCaseExpectErr {
	require.NotNil(t, err, "(%s) expected an error when deploying metering", testFunction.Name)
	require.Equal(t, err, testCaseExpectErrMsg, "(%s) expected the error returned from ")
} else {
	require.NoError(t, err, "(%s) deploying metering should produce no error", testFunction.Name)
}

if rf != nil && testInstallFunction != nil {
	(run the reporting tests)
	...
}
```

If we still want to support testing multiple functions per a `MeteringConfig` configuration, this would be more difficult to remove as we iterate over that list of functions, install metering, then run reporting tests.

test/e2e/main_test.go
```diff
@@ -78,19 +80,46 @@ func TestMain(m *testing.M) {
 	os.Exit(m.Run())
 }
 
+type InstallTestCase struct {
+	Name      string
+	ExpectErr bool
+	TestFunc  func(t *testing.T, testReportingFramework *reportingframework.ReportingFramework)
+}
+
 func TestInstallMeteringAndReportingProducesData(t *testing.T) {
```
chancez: We'll have to pick a better name for this now that it's doing a ton of different things in subtests

chancez: `TestMeteringManualInstall` maybe?

test/e2e/main_test.go
```diff
@@ -194,130 +229,219 @@ func TestInstallMeteringAndReportingProducesData(t *testing.T) {
...
-		{queryName: "namespace-cpu-utilization"},
-		{queryName: "pod-cpu-request-aws", skip: !runAWSBillingTests, nonParallel: true},
-		{queryName: "pod-memory-request-aws", skip: !runAWSBillingTests, nonParallel: true},
-		{queryName: "aws-ec2-cluster-cost", skip: !runAWSBillingTests, nonParallel: true},
-	}
+		rf, canSafelyRunTest, err := deployerCtx.Setup(testExpectErr)
+		if testExpectErr {
+			assert.NotNil(t, err, "expected the install to fail but it didnt")
+			if err != nil {
+				assert.Contains(t, err.Error(), testExpectErrMsg)
```
timflannagan1: This could be problematic as `err` is just an interface so null checking might not be the correct implementation and would result in a panic.

If we changed the `fmt.Errorf(...)` in the deploy package to `errors.Wrap(err, "...")` we might be able to use `assert.EqualErrors(t, errors.Cause(err), testExpectErrMsg)` to get the last error message in the stack trace, which is safer and eliminates the need to also check if `assert.NotNil(t, err, ...)`.

test/deployframework/helpers.go
```diff
@@ -119,12 +118,6 @@ func (pw *PodWaiter) WaitForPods(namespace string, targetPodsCount int) error {
 			return false, err
 		}
 
-		// TODO(chancez): is this check needed? If so, maybe move outside of
-		// WaitForPods.
-		if len(pods.Items) == 0 {
```
timflannagan1: We definitely don't need this check anymore, especially with the introduction of running the e2e suite locally (as the metering-operator pod would be run locally.)

PR #994, opened 2019-10-17
test/deployframework: Ensure the overrided reporting-operator image isn't empty.

timflannagan1:
As we push towards refactoring the e2e/integration suites, we need to ensure that overriding the reporting-operator images works as intended. 

This PR adds a testhelper function that validates if the image repository or tag are empty, when that image is overrided.

### DIFF
--- a/test/deployframework/context.go
+++ b/test/deployframework/context.go
@@ -38,8 +38,8 @@ type DeployerCtx struct {
 
 // NewDeployerCtx constructs and returns a new DeployerCtx object
 func (df *DeployFramework) NewDeployerCtx(
-	repo,
-	tag,
+	meteringOperatorImageRepo,
+	meteringOperatorImageTag,
 	namespace,
 	outputPath string,
 	targetPods int,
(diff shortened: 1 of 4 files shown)


timflannagan1:
Used the wrong parameters to verify
/hold

chancez:
Cool. Adding validation is a great idea to prevent misuse.

test/deployframework/context.go
```diff
@@ -69,6 +69,14 @@ func (df *DeployFramework) NewDeployerCtx(
 		},
 	}
 
+	// validate the reporting-operator image isn't empty when overrided
+	if df.ReportingOperatorImageRepo != "" || df.ReportingOperatorImageTag != "" {
```
chancez: Should we do the same for meteringOperatorImageRepo and meteringOperatorImageTag ?

timflannagan1: The metering operator images get a default value (defined in the Makefile) so I didn't think we would need to validate this image, but I can add a check that verifies both the image repo and tag are both set, which should be sufficient.

PR #996, opened 2019-10-18
Return correct resource_id for Azure

adberglund:
## Summary

The proposed change here uses a CASE statement to return the correct resource_id depending on the cloud the cluster is installed in. This adds support for Azure to the resource_id column.


chancez:
@adberglund What's an azure resource ID look like anyways?

PR #997, opened 2019-10-18
test/e2e: Fail test install helper function when NewDeployerCtx return an error.

timflannagan1:
If there's an error or the `DeployerCtx` object is nil when running the `NewDeployerCtx` constructor, we need to exit the test in order to avoid a panic call when the Setup() method is run (as the `DeployerCtx` object will be nil.)

In the current architecture, we can't return an error from the `testInstall` helper function, as that function is potentially run in a goroutine (via `t.Run(...)`), and we need to run the defer closure. If the `DeployerCtx` object return is nil, then we don't care about logging resource output or uninstalling metering, but we do care about that once the `Setup` method is run.

### DIFF
--- a/test/e2e/main_test.go
+++ b/test/e2e/main_test.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/sirupsen/logrus"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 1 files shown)


test/e2e/main_test.go
```diff
@@ -211,17 +212,19 @@ func testInstall(
...
 	rand.Seed(time.Now().UnixNano())
 	namespace := df.NamespacePrefix + "-" + strconv.Itoa(rand.Intn(50))
 
 	deployerCtx, err := df.NewDeployerCtx(repo, tag, namespace, testOutputDir, targetPods, spec)
-	assert.NoError(t, err, "creating a new deployer context should produce no error")
+	require.NoError(t, err, "creating a new deployer context should produce no error")
+	require.NotNil(t, deployerCtx, "the DeployerCtx object should not be nil")
 
 	rf, err := deployerCtx.Setup()
 	assert.NoError(t, err, "deploying metering should produce no error")
```
chancez: I feel like both of these should be require

test/e2e/main_test.go
```diff
@@ -211,17 +212,19 @@ func testInstall(
...
+	assert.NotNil(t, rf, "the reportingframework should not be nil")
```
chancez: require?

test/e2e/main_test.go
```diff
@@ -211,17 +212,19 @@ func testInstall(
...
+	require.NoError(t, err, "creating the test case output directory should produce no error")
 
 	// randomize the namespace to avoid existing namespaces
 	rand.Seed(time.Now().UnixNano())
 	namespace := df.NamespacePrefix + "-" + strconv.Itoa(rand.Intn(50))
 
 	deployerCtx, err := df.NewDeployerCtx(repo, tag, namespace, testOutputDir, targetPods, spec)
-	assert.NoError(t, err, "creating a new deployer context should produce no error")
+	require.NoError(t, err, "creating a new deployer context should produce no error")
+	require.NotNil(t, deployerCtx, "the DeployerCtx object should not be nil")
```
chancez: Do we really need both? I feel like deployerCtx should never be non-nil if there's not an error, and it should never be set if there is, so the `require.NoError` should be sufficient.

test/e2e/main_test.go
timflannagan1: I used `require` here, then any statement that evaluates to false will end the test function, and the defer closure won't be run.

test/e2e/main_test.go
timflannagan1: Hmm not really, I'll remove it.

timflannagan1:
Superseded by #992

PR #1005, opened 2019-10-24
Support running the e2e suite locally.

timflannagan1:
### Overview:
This would update the e2e suite to support running e2e locally by deploying both the metering and reporting operators locally.

### TODO:
- [x] Look into breaking up the `deployerCtx.Setup()` method into helper functions
- [x] Stop passing the metering operator image to the `deployerCtx.Setup()` method
- [x] Look into passing a list of environment variables to the `RunReportingOperatorLocal()` method
- [x] Look into whether the directory containing the reporting-operator TLS data should be stored in a tmpdir (in the case where tests are interrupted)
- [x] Add a field/flag that contains the path to the `hack/run-reporting-operator-local.sh` script (this is useful for when directly run via `go test ...`)


test/deployframework/local.go
```diff
@@ -0,0 +1,206 @@
...
+		"REPORTING_OPERATOR_PROMETHEUS_DATASOURCE_MAX_IMPORT_BACKFILL_DURATION=15m",
+		"REPORTING_OPERATOR_PROMETHEUS_METRICS_IMPORTER_INTERVAL=30s",
+		"REPORTING_OPERATOR_PROMETHEUS_METRICS_IMPORTER_CHUNK_SIZE=5m",
+		"REPORTING_OPERATOR_PROMETHEUS_METRICS_IMPORTER_INTERVAL=5m",
+		"REPORTING_OPERATOR_PROMETHEUS_METRICS_IMPORTER_STEP_SIZE=60s",
+		"REPORTING_OPERATOR_TLS_DIR=" + lc.ReportingOperatorTLSPath,
+	}
+
+	// TODO: add this to LocalCtx?
+	scriptPath, err := filepath.Abs("../../hack/run-reporting-operator-local.sh")
```
chancez: yeah probably ideal if we have a flag/variable (defaulting to PWD) that's the path to the repo. I think deployer.Config has similar stuff for manifests. Then we can just use `path.Join(repoDir, "hack/run-reporting-operator-local.sh")`. What do you think?

test/deployframework/local.go
```diff
@@ -0,0 +1,206 @@
...
+}
+
+// RunMeteringOperatorLocal is a method that runs the metering-operator locally
+func (lc *LocalCtx) RunMeteringOperatorLocal() error {
+	dockerCmd := exec.Command(
+		"docker", "run",
+		"--rm",
+		"-u", "0:0",
+		"--name", meteringOperatorContainerName,
+		"-v", lc.KubeConfigPath+":/kubeconfig",
```
chancez: Since you've got `destKubeConfigPath you should probably use that here for `/kubeconfig`

test/deployframework/local.go
```diff
@@ -0,0 +1,206 @@
...
+
+	dockerCmd.Stdout = logFile
+	dockerCmd.Stderr = logFile
+
+	err = dockerCmd.Start()
+	if err != nil {
+		return fmt.Errorf("failed to run the metering-operator container: %v", err)
+	}
+
+	lc.MeteringOperatorPID = dockerCmd.Process.Pid
```
chancez: This is fine I think, but I wonder if we should be storing the `dockerCmd` too, so we can call `cmd.Wait()` during cleanup.

test/deployframework/local.go
```diff
@@ -0,0 +1,206 @@
...
+func (lc *LocalCtx) CleanupLocal() error {
+	var errArr []string
+
+	lc.Logger.Infof("Removing the '%s' reporting-operator TLS directory", lc.ReportingOperatorTLSPath)
+	err := os.RemoveAll(lc.ReportingOperatorTLSPath)
+	if err != nil {
+		errArr = append(errArr, fmt.Sprintf("failed to remove the directory containing the reporting-operator TLS data: %v", err))
+	}
+
+	err = cleanupPID(lc.Logger, lc.MeteringOperatorPID, lc.ReportingOperatorPID)
```
chancez: We should probably only be doing this after `docker rm -f` which should stop the docker container also.

test/deployframework/local.go
chancez: Actually, we should definitely be storing `dockerCmd` and not `MeteringOperatorPID`. Two reasons:
- in cleanupPID you use `FindProcess` to get an `*os.Process` to kill, but `dockerCmd.Process` is already an `*os.Process` so you're unnecessarily looking up the process when we already have it.
- We want to use `cmd.Wait()`

test/deployframework/helpers.go
```diff
@@ -190,3 +192,69 @@ func GetServiceAccountToken(client kubernetes.Interface, namespace, serviceAccou
...
+		"cleanup": pids,
+	}).Infof("Killing the local process IDs")
+
+	for _, pid := range pids {
+		p, err := os.FindProcess(pid)
+		if err != nil {
+			errArr = append(errArr, fmt.Sprintf("failed to find the %d pid: %v", pid, err))
+		}
+
+		err = p.Kill()
```
chancez: Does the original script `kill -9` or just `kill`? If the latter, we should use `p.Signal(signal.Interrupt)` so it's graceful.

timflannagan1: It was just a regular `kill`:
```bash
    if [ "$DEPLOY_METERING_OPERATOR_LOCAL" == "true" ]; then
        echo "Stopping local metering-operator"
        [ -s "$METERING_OPERATOR_PID_FILE" ] && kill "$(cat "$METERING_OPERATOR_PID_FILE")" || true
        rm -f "$METERING_OPERATOR_PID_FILE"
        docker rm -f "$METERING_OPERATOR_CONTAINER_NAME" || true
    fi
```

test/deployframework/local.go
timflannagan1: I think it may make more sense to provide a path to the `hack/` directory instead, as we also need the path to the `hack/run-test-cleanup.sh` script that logs the resource and container logs. 

That way we can remove the cleanup script flag, and only manage the path to the `hack` script directory directly (via `filepath.Join(...)` and const variables).

test/deployframework/local.go
timflannagan1: I was hesitant to use `cmd.Wait()` due to the reconciliation aspect, but never got around to testing the behavior of that. I think what you're proposing makes more sense for both the metering and reporting operator commands used, e.g.

```go
type LocalCtx struct {
    ...
    MeteringOperatorProcess *os.Process
    ReportingOperatorProcess *os.Process
    ...
}
```

timflannagan1: Hmm I don't think we want to use `cmd.Wait()` as cleanup just hangs while waiting for the reporting-operator to finish execution. I guess we just need to be more mindful in cleaning up the resources used by reporting-operator-local ourselves.

timflannagan1: Well I guess we could just kill the process then release the resources, e.g.
```go
err := cmd.Process.Kill()
if err != nil { ... }

err = cmd.Process.Release()
if err != nil { ... }
```

test/deployframework/helpers.go
chancez: Then we should be using `Signal` not `Kill`

test/deployframework/local.go
```diff
@@ -0,0 +1,212 @@
...
+
+	return nil
+}
+
+// RunReportingOperatorLocal is a method that runs the reporting-operator locally
+func (lc *LocalCtx) RunReportingOperatorLocal(apiListenPort, metricsListenPort, pprofListenPort int, token string) error {
+	var err error
+
+	// TODO: should this still be a tmpdir in the case where tests are interrupted early?
+	lc.ReportingOperatorTLSPath, err = ioutil.TempDir(lc.BasePath, reportingOperatorTLSDirName)
```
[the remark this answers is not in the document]
chancez: We don't want to hardcode directories, ever. If this is an issue it should be adapted to support using an existing directory instead of creating a tempdir. 

Additionally, it may be best if we simply replace the logic in `run-reporting-operator-local.sh` with Go code since there's so many background tasks, it would be way simpler to do it in Go.

test/deployframework/local.go
timflannagan1: Yeah I agree - I read up on how release is handled, and I don't think we would want to use that, especially if we can just send an interrupt to the script, wait for it to complete (which releases the all resources). This means we don't need commands to remove the docker container, or command PIDS anymore.

test/deployframework/local.go
chancez: I was thinking the path to the repo (which could default to `PWD`) since we already need the manifest directory, which is in the repo and we could potentially just use one var in tests for paths, and it could just be the path to the repo. What do you think?

test/deployframework/local.go
```diff
@@ -0,0 +1,189 @@
...
+	disableOCPFeatures = false
+)
+
+// LocalCtx holds all the necessary information to run e2e tests locally
+type LocalCtx struct {
+	Namespace                      string
+	BasePath                       string
+	KubeConfigPath                 string
+	HackScriptPath                 string
+	MeteringOperatorImage          string
```
chancez: nit: For consistency with the deployer.Config I think we should keep this separated as `meteringOperatorImageRepo` and `meteringOperatorImageTag` and then combine them when we call `exec.Command` in `RunMeteringOperatorLocal`.

test/deployframework/local.go
```diff
@@ -0,0 +1,189 @@
...
+		"-e", "OPERATOR_NAME="+meteringOperatorName,
+		"-e", "POD_NAME="+meteringOperatorName,
+		"-e", "WATCH_NAMESPACE="+lc.Namespace,
+		"-e", "ENABLE_DEBUG="+strconv.FormatBool(enableDebug),
+		"-e", "DISABLE_OCP_FEATURES="+strconv.FormatBool(disableOCPFeatures),
+		lc.MeteringOperatorImage,
+	)
+
+	lc.Logger.WithFields(logrus.Fields{
+		"command": lc.MeteringOperatorCmd.Args,
```
chancez: I'm kind of thinking this shouldn't be a log field, but maybe just it's own debug log message.

test/deployframework/local.go
timflannagan1: Updated.

test/deployframework/context.go
```diff
@@ -111,14 +160,22 @@ func (ctx *DeployerCtx) Setup() (*reportingframework.ReportingFramework, error)
...
-	// support passing a METERING_CR_FILE yaml file for local testing
-	useHTTPSAPI := true
-	useRouteForReportingAPI := true
-	useKubeProxyForReportingAPI := false
-	reportingAPIURL := ""
+	if ctx.RunTestLocal {
+		useHTTPSAPI = false
+		useRouteForReportingAPI = false
+		useKubeProxyForReportingAPI = false
+		reportingAPIURL = fmt.Sprintf("%s://%s:%d/healthy", httpScheme, localAddr, apiPort)
```
chancez: This still didn't get renamed to reflect that it is not a reporting API URL, just a health endpoint.

test/deployframework/local.go
```diff
@@ -0,0 +1,189 @@
...
+	err = lc.ReportingOperatorCmd.Start()
+	if err != nil {
+		return fmt.Errorf("failed to run the reporting-operator locally: %v", err)
+	}
+
+	return nil
+}
+
+// WaitForReportingOperator runs a GET request for the reporting-operator healthy endpoint
+func (lc *LocalCtx) WaitForReportingOperator(reportingAPIURL string) error {
```
chancez: From what I see, this isn't actually specific to ReportingOperator, it just waits for a specific URL to return 200. Can you rename it to reflect this, and update the logging to also? Also it doesn't really look like this needs to be a method, I think it can probably just be a function which takes a logger and a URL.

test/deployframework/local.go
timflannagan1: After playing around with this a bit, I think it would make the most sense to default to `../../` which assumes you're running the `go test ...` command directly (as hack/e2e.sh sets this to `$ROOT_DIR`).

If you elect to run go test directly, then you would need to provide the absolute path to the operator-metering directory.

test/deployframework/local.go
```diff
@@ -0,0 +1,154 @@
...
+
+	return nil
+}
+
+// CleanupLocal removes all resources that were created while running e2e locally
+func (lc *LocalCtx) CleanupLocal() error {
+	var errArr []string
+
+	if lc.MeteringOperatorCmd != nil && lc.ReportingOperatorCmd != nil {
+		err := cleanupLocalCmds(lc.Logger, *lc.MeteringOperatorCmd, *lc.ReportingOperatorCmd)
```
timflannagan1: @chancez I'm on the fence about consolidating these two commands into a `*exec.Cmd` slice/array, which would be an easier nil check and not change the implementation too much.

timflannagan1: Updated to a list of `exec.Cmd` since last review.

test/deployframework/local.go
```diff
@@ -0,0 +1,153 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to stat the '%s' path: %v", targetScriptDir, err)
+	}
+
+	cmd := exec.Command(targetScriptDir)
+	cmd.Env = append(os.Environ(), envVarArr...)
+
+	lc.Logger.WithFields(logrus.Fields{
+		"command":     cmd.Args,
+		"environment": envVarArr,
```
chancez: This prints secrets

test/deployframework/context.go
```diff
@@ -133,49 +213,45 @@ func (ctx *DeployerCtx) Setup() (*reportingframework.ReportingFramework, error)
 		ctx.MeteringClient,
 	)
 	if err != nil {
-		ctx.Logger.Infof("Failed to construct a reportingframework: %v", err)
 		return nil, fmt.Errorf("failed to construct a reportingframework: %v", err)
 	}
 
+	ctx.Logger.Infof("ReportingFramework Object: %+v", rf)
```
chancez: This prints secrets

timflannagan1: Updated.

PR #1011, opened 2019-11-01
test: Add support for testing an invalid MeteringConfig.

timflannagan1:
This would extend the e2e suite to test for an invalid `MeteringConfig` (missing storage spec), and add another sub-test that would ensure that the current context's `MeteringConfig` was not created.

### TODO:
- [x] Delay the installation error in `ctx.Setup()` when we expect an error.
- [x] Skip the sub-test when there was an error, and we didn't expect an error.
- [x] Assert that there's an error when we expect an installation error.
- [x] Assert that the returned error contains a list of sub-strings when the test expects an installation error.

### DIFF
--- a/pkg/operator/deploy/install.go
+++ b/pkg/operator/deploy/install.go
@@ -35,7 +35,7 @@ func (deploy *Deployer) installNamespace() error {
 
 		_, err := deploy.client.CoreV1().Namespaces().Create(namespaceObj)
 		if err != nil {
-			return fmt.Errorf("Failed to create %s namespace: %v", deploy.config.Namespace, err)
+			return err
 		}
 		deploy.logger.Infof("Created the %s namespace", deploy.config.Namespace)
 	} else if err == nil {
@@ -53,7 +53,7 @@ func (deploy *Deployer) installNamespace() error {
 
 			_, err := deploy.client.CoreV1().Namespaces().Update(namespace)
 			if err != nil {
-				return fmt.Errorf("Failed to add the 'openshift.io/cluster-monitoring' label to the %s namespace: %v", deploy.config.Namespace, err)
+				return fmt.Errorf("failed to add the 'openshift.io/cluster-monitoring' label to the %s namespace: %v", deploy.config.Namespace, err)
 			}
 		} else {
 			deploy.logger.Infof("The %s namespace already exists", deploy.config.Namespace)
@@ -70,7 +70,7 @@ func (deploy *Deployer) installMeteringConfig() error {
 	if apierrors.IsNotFound(err) {
 		_, err = deploy.meteringClient.MeteringConfigs(deploy.config.Namespace).Create(deploy.config.MeteringConfig)
 		if err != nil {
-			return fmt.Errorf("Failed to create the MeteringConfig resource: %v", err)
+			return err
 		}
 		deploy.logger.Infof("Created the MeteringConfig resource")
 	} else if err == nil {
@@ -78,7 +78,7 @@ func (deploy *Deployer) installMeteringConfig() error {
 
 		_, err = deploy.meteringClient.MeteringConfigs(deploy.config.Namespace).Update(mc)
 		if err != nil {
-			return fmt.Errorf("Failed to update the MeteringConfig: %v", err)
+			return fmt.Errorf("failed to update the MeteringConfig: %v", err)
 		}
 		deploy.logger.Infof("The MeteringConfig resource has been updated")
 	} else {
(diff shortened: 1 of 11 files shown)


test/testhelpers/assert.go
```diff
@@ -69,3 +70,13 @@ func AssertReportResultsEqual(t *testing.T, expected, actual []map[string]interf
 		}
 	}
 }
+
+func AssertErrorContainsErrorMsgs(t *testing.T, err error, errMsgArr []string) {
+	require.NotNil(t, err, "expected the error would not be nil")
+
+	strErr := strings.ReplaceAll(err.Error(), "\"", "")
```
chancez: Is this needed?

test/e2e/invalid_meteringconfig_produces_error_tests.go
```diff
@@ -0,0 +1,23 @@
...
+
+func testInvalidMeteringConfigMissingStorageSpec(t *testing.T, rf *reportingframework.ReportingFramework) {
+	require.NotNil(t, rf, "expected the reportingframework object would not be nil")
+	require.NotEmpty(t, rf.Namespace, "expected the reportingframework Namespace field would not be empty")
+	require.NotNil(t, rf.MeteringClient, "expected the reportingframework MeteringClient field would not be nil")
+
+	// we expect that the returned mc.Spec will only have nil fields (e.g. Spec:{DisableOCPFeatures:<nil> LogHelmTemplate:<nil>)
+	mc, err := rf.MeteringClient.MeteringConfigs(rf.Namespace).Get("operator-metering", meta.GetOptions{})
+	assert.Error(t, err, "expected there would be an error trying to query for the invalid meteringconfig")
+	assert.Nil(t, mc.Spec.Storage, "expected the meteringconfig storage spec would be nil")
```
chancez: This is unnecessary since we expect an error.

test/e2e/invalid_meteringconfig_produces_error_tests.go
```diff
@@ -0,0 +1,23 @@
...
+)
+
+func testInvalidMeteringConfigMissingStorageSpec(t *testing.T, rf *reportingframework.ReportingFramework) {
+	require.NotNil(t, rf, "expected the reportingframework object would not be nil")
+	require.NotEmpty(t, rf.Namespace, "expected the reportingframework Namespace field would not be empty")
+	require.NotNil(t, rf.MeteringClient, "expected the reportingframework MeteringClient field would not be nil")
+
+	// we expect that the returned mc.Spec will only have nil fields (e.g. Spec:{DisableOCPFeatures:<nil> LogHelmTemplate:<nil>)
+	mc, err := rf.MeteringClient.MeteringConfigs(rf.Namespace).Get("operator-metering", meta.GetOptions{})
+	assert.Error(t, err, "expected there would be an error trying to query for the invalid meteringconfig")
```
chancez: Maybe instead of assert.Error we verify the error is a kubernetes not found error. Since that's normally determined by a helper, we could do something like `assert.True(t, apierrors.IsNotFound(err), "expected MeteringConfig to not exist")`

test/testhelpers/assert.go
timflannagan1: Yes, there's going to be nested `"` that aren't going to be escaped:
```go
failed to install metering: Failed to create the MeteringConfig resource: Failed to create the MeteringConfig resource: MeteringConfig.metering.openshift.io \"operator-metering\" is invalid: []: Invalid value: map[string]interface {}{\"metadata\":map[string]interface {}{\"name\":\"operator-metering\", \"namespace\":\"tflannag-invalidhdfs-missingstoragespec\", \"creationTimestamp\":\"2019-11-05T21:29:17Z\", \"generation\":1, \"uid\":\"525914c9-0013-11ea-abed-0a3e8aca242c\"}, \"spec\":map[string]interface {}{\"logHelmTemplate\":true}, \"kind\":\"MeteringConfig\", \"apiVersion\":\"metering.openshift.io/v1\"}: validation failure list:\nspec.storage in body is required
```

I can add a comment clarifying the intent and describing that behavior.

chancez: Right, but we aren't using quotes in the errors we're checking for from what I saw, so I wasn't sure if this was actually required or not, since we aren't matching against `"`, it may not matter if they're escaped or not.

test/e2e/invalid_meteringconfig_produces_error_tests.go
timflannagan1: At one point that was in the previous implementation (in the consolidating PR before dropping those commits) but I must have applied the wrong stash - I agree that's a much better check that just asserting there was an error trying to query for the intended `MeteringConfig`.

test/testhelpers/assert.go
timflannagan1: Hmm I've run this a bunch locally and I'm not hitting any errors after removing the aforementioned.

Previously, I would get something similar to ```... could not be applied builtin len()```, when running the `assert.Contains(...)` so that must have been needed in a prior implementation.

timflannagan1:
Cherry-picked the deploy package error formatting changes to this PR.

pkg/operator/deploy/install.go
```diff
@@ -70,15 +70,15 @@ func (deploy *Deployer) installMeteringConfig() error {
 	if apierrors.IsNotFound(err) {
 		_, err = deploy.meteringClient.MeteringConfigs(deploy.config.Namespace).Create(deploy.config.MeteringConfig)
 		if err != nil {
-			return fmt.Errorf("Failed to create the MeteringConfig resource: %v", err)
+			return err
```
EmilyM1: Did we want an error message here?

timflannagan1: This belonged to a cherry-picked commit from master - this just reduced redundant error messages.

PR #1015, opened 2019-11-05
K8s Event Creation Module

EmilyM1:
Ansible Module to create and update Metering Events for debugging.
Moved over from ansible fork.
Need to comment.
Example of JSON user supplied args provided.

1. Need ansible installed
2. Need to be in venv
3. Separate install of `openshift` package required 
pip install openshift
4. To run
python -m <path to module>_mod <path to args>

### DIFF
--- a/images/metering-ansible-operator/roles/meteringconfig/library/k8s_event.py
+++ b/images/metering-ansible-operator/roles/meteringconfig/library/k8s_event.py
@@ -1,0 +1,338 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# Copyright (c) 2019 Emily Moss
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+
+__metaclass__ = type
+
+
+ANSIBLE_METADATA = {
+    "metadata_version": "1.1",
+    "status": ["preview"],
+    "supported_by": "community",
+}
+
+DOCUMENTATION = """
+
+module: k8s_events
+
+short_description: Create Kubernetes Events
+
+version_added: "2.8"
+
+author: Emily Moss for Red Hat
+
+description:
+  - Create Kubernetes Events for Metering
+
+extends_documentation_fragment:
+  - k8s_auth_options
+
+options:
+  resource_definition:
+    description:
+    - A partial YAML definition of the Event object being created/updated. Here you can define Kubernetes
+      Event Resource parameters not covered by this module's parameters.
+    - "NOTE: I(resource_definition) has lower priority than module parameters. If you try to define e.g.
+      I(metadata.namespace) here, that value will be ignored and I(metadata) used instead."
+    aliases:
+    - definition
+    - inline
+    type: dict
+  state:
+    description:
+    - Determines if an object should be created, patched, or deleted. When set to C(present), an object will be
+      created, if it does not already exist. If set to C(absent), an existing object will be deleted. If set to
+      C(present), an existing object will be patched, if its attributes differ from those specified using
+      module options and I(resource_definition).
+    default: present
+    choices:
+    - present
+    - absent
+  force:
+    description:
+    - If set to C(True), and I(state) is C(present), an existing object will be replaced.
+    default: false
+    type: bool
+  merge_type:
+    description:
+    - Whether to override the default patch merge approach with a specific type. By default, the strategic
+      merge will typically be used.
+    - For example, Custom Resource Definitions typically aren't updatable by the usual strategic merge. You may
+      want to use C(merge) if you see "strategic merge patch format is not supported"
+    - See U(https://kubernetes.io/docs/tasks/run-application/update-api-object-kubectl-patch/#use-a-json-merge-patch-to-update-a-deployment)
+    - Requires openshift >= 0.6.2
+    - If more than one merge_type is given, the merge_types will be tried in order
+    - If openshift >= 0.6.2, this defaults to C(['strategic-merge', 'merge']), which is ideal for using the same parameters
+      on resource kinds that combine Custom Resources and built-in resources. For openshift < 0.6.2, the default
+      is simply C(strategic-merge).
+    choices:
+    - json
+    - merge
+    - strategic-merge
+    type: list
+  name:
+    description:
+      - Use to specify a Event object name.
+    required: true
+    type: str
+  namespace:
+    description:
+      - Use to specify a Event object namespace.
+    required: true
+    type: str
+  message:
+    description:
+      - Status for Operation
+    required: true
+    type: str
+  reason:
+    description:
+      - Reason for the transition into the objects current status
+    required: true
+    type: str
+  reportingComponent:
+    description:
+      - Component responsible for event
+    required: true
+    type: str
+  type:
+    description:
+      - Type of Event
+    choices:
+      - Warning
+      - Normal
+  source:
+    description:
+      - Component for reporting this Event
+    required: true
+    type: string
+  involvedObject:
+    description: ObjectReference
+      - Object event is reporting on. ApiVersion, kind, name and namespace are of the involvedObject.
+    - apiVersion
+      required: true
+      type: string
+    - kind
+      required: true
+      type: string
+    - name
+      required: true
+      type: string
+    - namespace
+      required: true
+      type: string
+
+requirements:
+  - python >= 2.7
+  - openshift >= 0.6.2
+"""
+
+EXAMPLES = """
+
+- name: Create Kubernetes Event
+  k8s_events
+    state: present
+    name: test-https-emily109
+    namespace: default
+    message: Event created
+    reason: Created
+    reportingComponent: Reporting components
+    type: Normal
+    source:
+    component: Metering components
+    involvedObject:
+      apiVersion: v1
+      kind: Service
+      name: test-https-emily107
+      namespace: default
+"""
+
+RETURN = """
+result:
+  description:
+  - The created, patched, or otherwise present Event object. Will be empty in the case of a deletion.
+  returned: success
+  type: complex
+  contains:
+     api_version:
+        description: The versioned schema of this representation of an object.
+        returned: success
+        type: str
+     count:
+       description: Count of Event occurance
+       returned: success
+       type: int
+     firstTimestamp:
+       description: Timestamp of first occurance of Event
+       returned: success
+       type: timestamp
+     involvedObject:
+       description: Object event is reporting on. Includes apiVersion, kind, name, namespace, resourceVersion and uid.
+       returned: success
+       type: complex
+     kind:
+       description: Always 'Event'.
+       returned: success
+       type: str
+     lastTimestamp:
+       description: Timestamp of last occurance of Event
+       returned: success
+       type: timestamp
+     message:
+       description: Status for operation
+       returned: success
+       type: str
+     metadata:
+       description: Standard object metadata. Includes name, namespace, annotations, labels, etc.
+       returned: success
+       type: complex
+     reason:
+       description: Reason for the transition into the objects current status
+       returned: success
+       type: str
+     reportingComponent:
+       description: Component responsible for event
+       returned: success
+       type: str
+     source:
+       description: Component for reporting this Event
+       returned: success
+       type: string
+     type:
+       description: Type of Event. Either Normal or Warning
+       returned: success
+       type: string
+
+"""
+
+import copy
+import traceback
+import datetime
+import kubernetes.config.dateutil
+import openshift
+
+from collections import defaultdict
+
+from ansible.module_utils.k8s.common import AUTH_ARG_SPEC
+from ansible.module_utils.k8s.raw import KubernetesRawModule
+
+EVENT_ARG_SPEC = {
+    "state": {"default": "present", "choices": ["present", "absent"]},
+    "name": {"required": True},
+    "namespace": {"required": True},
+    "merge_type": {"type": "list", "choices": ["json", "merge", "strategic-merge"]},
+    "message": {"type": "str", "required": True},
+    "reason": {"type": "str", "required": True},
+    "reportingComponent": {"type": "str", "required": True},
+    "type": {"choices": ["Normal", "Warning"]},
+    "source": {"type": "str", "component": {"type": "str", "required": True}},
+    "involvedObject": {
+        "type": "dict",
+        "apiVersion": {"type": "str", "required": True},
+        "kind": {"type": "str", "required": True},
+        "name": {"type": "str", "required": True},
+        "namespace": {"type": "str", "required": True},
+    },
+}
+
+
+class KubernetesEvent(KubernetesRawModule):
+    def __init__(self, *args, **kwargs):
+        super(KubernetesEvent, self).__init__(*args, k8s_kind="Event", **kwargs)
+
+    @property
+    def argspec(self):
+        """ argspec property builder """
+        argumentSpec = copy.deepcopy(AUTH_ARG_SPEC)
+        argumentSpec.update(EVENT_ARG_SPEC)
+        return argumentSpec
+
+    def execute_module(self):
+        """ Module execution """
+        self.client = self.get_api_client()
+
+        message = self.params.get("message")
+        reason = self.params.get("reason")
+        reporting_component = self.params.get("reportingComponent")
+        event_type = self.params.get("type")
+        source = self.params.get("source")
+
+        definition = defaultdict(defaultdict)
+
+        meta = definition["metadata"]
+        meta["name"] = self.params.get("name")
+        meta["namespace"] = self.params.get("namespace")
+
+        involved_obj = self.params.get("involvedObject")
+
+        resource = self.find_resource("Event", "v1", fail=True)
+
+        prior_event = None
+        try:
+            prior_event = resource.get(
+                name=meta["name"],
+                namespace=meta["namespace"])
+
+        except openshift.dynamic.exceptions.NotFoundError:
+            pass
+
+        prior_count = 1
+        now = datetime.datetime.now()
+        rfc = kubernetes.config.dateutil.format_rfc3339(now)
+        first_timestamp = rfc
+        last_timestamp = rfc
+
+        if prior_event and prior_event["reason"] == reason:
+            prior_count = prior_event["count"] + 1
+            first_timestamp = prior_event["firstTimestamp"]
+            last_timestamp = rfc
+
+        involved_object_resource = self.find_resource(involved_obj["kind"], "v1", fail=True)
+
+        if involved_object_resource:
+            try:
+                api_involved_object = involved_object_resource.get(
+                   name=involved_obj["name"], namespace=involved_obj["namespace"])
+
+                involved_obj["uid"] = api_involved_object["metadata"]["uid"]
+                involved_obj["resourceVersion"] = api_involved_object["metadata"]["resourceVersion"]
+
+            except openshift.dynamic.exceptions.NotFoundError:
+                pass
+
+        event = {
+            "count": prior_count,
+            "eventTime": None,
+            "firstTimestamp": first_timestamp,
+            "involvedObject": involved_obj,
+            "lastTimestamp": last_timestamp,
+            "message": message,
+            "metadata": {
+                "name": meta["name"],
+                "namespace": meta["namespace"]
+            },
+            "reason": reason,
+            "reportingComponent": reporting_component,
+            "reportingInstance": "",
+            "source": {"component": source},
+            "type": event_type,
+        }
+
+        result = self.perform_action(resource, event)
+        self.exit_json(**result)
+
+
+def main():
+    module = KubernetesEvent()
+    try:
+        module.execute_module()
+    except Exception as e:
+        module.fail_json(msg=str(e), exception=traceback.format_exc())
+
+
+if __name__ == "__main__":
+    main()


images/metering-ansible-operator/roles/meteringconfig/library/k8s_event.py
```diff
@@ -0,0 +1,371 @@
...
+        def_meta['name'] = self.params.get('name')
+        def_meta['namespace'] = self.params.get('namespace')
+
+        def_involvedObject = definition['involvedObject']
+        def_involvedObject['namespace'] = self.params.get('namespace')
+        def_involvedObject['apiVersion'] = self.params.get('apiVersion')
+        def_involvedObject['kind'] = self.params.get('kind')
+        def_involvedObject['name'] = self.params.get('name')
+        def_involvedObject['uid'] = self.params.get('uid')
+        def_involvedObject['resourceVersion'] = self.params.get('resourceVersion')
```
timflannagan1: One-liner:

```python
involved_obj = dict(zip(['namespace', 'apiVersion', 'kind', 'name', 'uid', 'resourceVersion'], (self.params.get('namespace'), self.params.get('apiVersion'), self.params.get('kind'), self.params.get('name'), self.params.get('resourceVersion')))
```

images/metering-ansible-operator/roles/meteringconfig/library/k8s_event.py
```diff
@@ -0,0 +1,371 @@
...
+                #priorCount = 1
+                priorCount = priorCount + 1
+                print("the reason has changed", priorCount)
+        except openshift.dynamic.exceptions.NotFoundError:
+            pass
+
+        involvedObject_resourceVersion = "1"
+        involvedObject_uid = "1"
+        try:
+            totalEvent=resource.get(name=def_meta['name'], namespace=def_meta['namespace'])
```
timflannagan1: You're querying for the same event as the previous try/except block, and you're not changing any value of that event, so I think we can combine these two try/except blocks, e.g.:

```python
event = resource.get(name=def_meta['name'], namespace=def_meta['namespace'])

if event is not None:
    if event['reason'] != reason:
        # do all the event counting and timestamp tasks here
    
    # do all the involvedObj tasks here
```

images/metering-ansible-operator/roles/meteringconfig/library/k8s_event.py
```diff
@@ -0,0 +1,371 @@
...
+        try:
+            priorEvent=resource.get(name=def_meta['name'],
+                 namespace=def_meta['namespace'])
+            priorReason=priorEvent['reason']
+            print(priorReason)
+            print("current reason is %s" % reason)
+            priorCount = priorEvent['count']
+            print("the count from the prior event is %i" % priorCount)
+
+            if priorEvent is not None and priorReason != reason:
```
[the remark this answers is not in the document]
EmilyM1: The reason is what determines if it's a new event. So the compound 'if' is there.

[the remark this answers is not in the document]
timflannagan1: I'm not suggesting you change the logic of the conditional, but change the usage from `priorReason` to `prior_event['reason']` (as the contain the same value.)

images/metering-ansible-operator/roles/meteringconfig/library/k8s_event.py
EmilyM1: This threw a syntax error.

EmilyM1:
Comments address in ea418ec44a3b6aea09db233e66693be0e49952b9

images/metering-ansible-operator/roles/meteringconfig/library/k8s_event.py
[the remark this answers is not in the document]
EmilyM1: Fixed

images/metering-ansible-operator/roles/meteringconfig/library/k8s_event.py
EmilyM1: I considered this. Ultimately I really hate such nesting for readability, so I prefer to keep as is for clarity and separation.

timflannagan1: https://en.wikipedia.org/wiki/Don%27t_repeat_yourself

EmilyM1: See: Arrowhead Antipattern.

images/metering-ansible-operator/roles/meteringconfig/library/k8s_event.py
```diff
@@ -0,0 +1,382 @@
...
+
+        resource = self.find_resource(
+            "Event", "v1", fail=True
+        )  # finds resource or gets it from the client api, looks for resrouce
+        # second call to find resource for involved object, set kind to arg of involved OBject kind
+        # rnewesource = self.find_resource('involvedObjectKind', 'v1', fail=True)
+
+        # this just establishes tha vars and seperates out the reason and count from returned cli object
+        priorCount = 1
+        try:
```
chancez: All 2 of these `try` `except` blocks can be consolidated. You're querying the same thing multiple times unnecessarily. It would be simpler if you just did something like this:

```
priorEvent = None
# Query the API for the existing event. If it doesn't exist, priorEvent will be None.
try:
    priorEvent = resource.get(
        name=meta["name"],
        namespace=meta["namespace"]
    )
except openshift.dynamic.exceptions.NotFoundError:
    pass

# set some default values for timestamps/count for when priorEvent doesn't exist
firstTimestamp = now
lastTimestamp = now
count = 1

# handle if priorEvent exists
if priorEvent:
   # do stuff when reason changed
   if priorEvent["reason"] != reason:
       # do stuff related count when reason changed
       count = 1

       # do stuff related timestamps when reason changed
       firstTimestamp = ....
       lastTimestamp = ...
   else: # do stuff when reason doesn't change
       # do stuff related count when reason is the same
       count = priorEvent["count"] + 1
       # do stuff related timestamps when reason is the same
       firstTimestamp = ....
       lastTimestamp = ...
```

images/metering-ansible-operator/roles/meteringconfig/library/k8s_event.py
```diff
@@ -0,0 +1,382 @@
...
+                # priorCount = 1
+                priorCount +=1
+                print("the reason has changed", priorCount)
+        except openshift.dynamic.exceptions.NotFoundError:
+            pass
+
+        involvedObject_resourceVersion = "1"
+        involvedObject_uid = "1"
+        try:
+            totalEvent = resource.get(
```
chancez: Your querying the wrong resource here. Your querying the event, not the involved object. You need to use self.find_resource to be able to query the right type of resource for the involvedObject.

For example:

```
involvedObjectResource = self.find_resource(involvedObject["kind"], "v1", fail=True)
# if the "kind" for involvedObject exists
if involvedObjectResource:
    try:
        # get the involvedObject if it exists
        apiInvolvedObject = involvedObjectResource.get(name=involvedObject["name"], namespace=involvedObject["namespace"])
        # sets the involvedObject uid and resourceVersion from the values if the object exists
        # retrieved from the API
        involvedObject["uid"] = apiInvolvedObject["metadata"]["uid"]
        involvedObject["resourceVersion"] = apiInvolvedObject["metadata"]["resourceVersion"]
    except openshift.dynamic.exceptions.NotFoundError:
        # involvedObject doesn't exist, so simply do not set uid/resourceVersion on it
        pass
```

images/metering-ansible-operator/roles/meteringconfig/library/k8s_event.py
chancez: There's no more or less nesting in the example provided, it's simply that you are currently spreading the same conditional logic in multiple places.

images/metering-ansible-operator/roles/meteringconfig/library/k8s_event.py
```diff
@@ -0,0 +1,382 @@
...
+        namespace_args = self.params.get("namespace")
+        message = self.params.get("message")
+        reason = self.params.get("reason")
+        reporting_component = self.params.get("reportingComponent")
+        event_type = self.params.get("type")
+        source = self.params.get("source")
+
+        definition = defaultdict(defaultdict)
+
+        def_meta = definition["metadata"]
```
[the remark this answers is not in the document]
EmilyM1: Will remove the prefix 👍

images/metering-ansible-operator/roles/meteringconfig/library/k8s_event.py
```diff
@@ -0,0 +1,460 @@
...
+- name: Create Kubernetes Event
+  k8s_events
+  state: present
+  name: test-https-emily109
+  namespace: default
+  message: message here
+  reason: reason is now different againnnnnnnnnnnn
+  reportingComponent: reportingComponents here
+  type: Normal
+  source:
```
timflannagan1: Not a priority right now, but this example is mis-intended:
```yaml
- name: Create a k8s event
  k8s_events:
    state: present
    ...
```

images/metering-ansible-operator/roles/meteringconfig/library/k8s_event.py
```diff
@@ -0,0 +1,352 @@
...
+            "apiVersion": "v1",
+            "count": priorCount,
+            "eventTime": None,
+            "firstTimestamp": firstTimestamp,
+            "involvedObject": {
+                "apiVersion": involved_obj["apiVersion"],
+                "kind": involved_obj["kind"],
+                "namespace": involved_obj["namespace"],
+                "name": involved_obj["name"],
+                "resourceVersion": involvedObject_resourceVersion,
```
timflannagan1: It looks like you would need to validate that this/these value(s) aren't `None` as you initialize it in a try/except block, and there could be an error attempting to get the involved_object, which means that these values contain their default, assigned values (`None`).

images/metering-ansible-operator/roles/meteringconfig/library/k8s_event.py
```diff
@@ -0,0 +1,352 @@
...
+    reason: reason is now different againnnnnnnnnnnn
+    reportingComponent: reportingComponents here
+    type: Normal
+    source:
+    component: Metering components
+    involvedObject:
+    apiVersion: v1
+    kind: Event
+    name: involvedObject event names
+    namespace: default
```
timflannagan1: This block (lines 149-153) is still misindented:
```yaml
involvedObject:
  apiVersion: v1
  kind: Event
  name: ...
  namespace: default
```

images/metering-ansible-operator/roles/meteringconfig/library/k8s_event.py
EmilyM1: It's there. See line 298

images/metering-ansible-operator/roles/meteringconfig/library/k8s_event.py
EmilyM1: Will fix

images/metering-ansible-operator/roles/meteringconfig/library/k8s_event.py
timflannagan1: On line 292, you initializing the variable to `None`:
```python
involvedObject_resourceVersion = None
```
If `involvedObjectResource` evaluates to false, or there's an error while running:
```python
try:
                apiInvolvedObject = resource.get(
                    name=involved_obj["name"], namespace=involved_obj["namespace"])
```
then `involvedObject_resourceVersion` won't be re-initialized, and you would referencing the original value.

I'm not sure what happens if you pass `"resourceVersion": involvedObject_resourceVersion,`
to the event dictionary when `involvedObject_resourceVersion == None` and whether or not that would produce an error.

EmilyM1: At every major code change I test:
New Event,
New Event /same reason
New Event /different reason
Deletion and re-creation of event
I have not hit an error in any of these.

images/metering-ansible-operator/roles/meteringconfig/library/k8s_event.py
```diff
@@ -0,0 +1,352 @@
...
+                "kind": involvedObj["kind"],
+                "namespace": involvedObj["namespace"],
+                "name": involvedObj["name"],
+                "resourceVersion": involvedObjectResourceVersion,
+                "uid": involvedObjectUid,
+            },
+            "kind": "Event",
+            "lastTimestamp": lastTimestamp,
+            "message": message,
+            "metadata": {"name": meta["name"], "namespace": "default",},
```
timflannagan1: The `metadata.namespace` field is hardcoded to `default`. I think it should be:
```python
"metadata": { ... "namespace": meta["namespace"] }
```

EmilyM1: Will fix

images/metering-ansible-operator/roles/meteringconfig/library/k8s_event.py
```diff
@@ -0,0 +1,352 @@
...
+        argumentSpec.update(EVENT_ARG_SPEC)
+        return argumentSpec
+
+    def execute_module(self):
+        """ Module execution """
+        #auth
+        self.client = self.get_api_client()
+
+        # retrieve params
+        apiVersion = "v1"
```
timflannagan1: Double-check but this appears to be unused (you use a hardcoded value of "apiVersion": "v1" in the event dictionary).

images/metering-ansible-operator/roles/meteringconfig/library/k8s_event.py
```diff
@@ -0,0 +1,352 @@
...
+
+            except openshift.dynamic.exceptions.NotFoundError:
+                pass
+
+        event = {
+            "apiVersion": "v1",
+            "count": priorCount,
+            "eventTime": None,
+            "firstTimestamp": firstTimestamp,
+            "involvedObject": {
```
timflannagan1: Check if you can just pass `"involvedObject": involved_obj` as it looks like this should be possible.

images/metering-ansible-operator/roles/meteringconfig/library/k8s_event.py
```diff
@@ -0,0 +1,349 @@
...
+        # event counter
+        priorCount = 1
+        print("the count is no event exists %i" % priorCount)
+        now = datetime.datetime.now()
+        rfc = kubernetes.config.dateutil.format_rfc3339(now)
+        firstTimestamp = rfc
+        lastTimestamp = rfc
+
+        if priorEvent:
+            if priorEvent["reason"] != reason:
```
chancez: This entire branch is redundant now since this case is just setting the values to the same as what they're initialized to. We can just change this to handle only when reason is the same since the initial values handle when it's different and we don't need redundantly set them.

Eg:
```
        # default values are the same values to use as when the reason changes.
        priorCount = 1
        now = datetime.datetime.now()
        rfc = kubernetes.config.dateutil.format_rfc3339(now)
        firstTimestamp = rfc
        lastTimestamp = rfc

        # reason didn't change
        if priorEvent and priorEvent["reason"] == reason:
            priorCount = priorEvent["count"] + 1
            firstTimestamp = priorEvent["firstTimestamp"]
            lastTimestamp = rfc
```

images/metering-ansible-operator/roles/meteringconfig/library/k8s_event.py
```diff
@@ -0,0 +1,349 @@
...
+                firstTimestamp = rfc
+                print("the value of the firstTimestamp new event is", firstTimestamp)
+                lastTimestamp = rfc
+                print("the value of the lastTimestamp new event is", lastTimestamp)
+            else:
+                priorCount = priorEvent["count"] + 1
+                print("If event exists and reason is the same %i" % priorCount)
+                firstTimestamp = priorEvent["firstTimestamp"]
+                print("the value of the firstTimestamp event exists same reason is", firstTimestamp)
+                lastTimestamp = firstTimestamp
```
chancez: This should be `rfc`

images/metering-ansible-operator/roles/meteringconfig/library/k8s_event.py
```diff
@@ -0,0 +1,349 @@
...
+                "kind": involvedObj["kind"],
+                "namespace": involvedObj["namespace"],
+                "name": involvedObj["name"],
+                "resourceVersion": involvedObjectResourceVersion,
+                "uid": involvedObjectUid,
+            },
+            "kind": "Event",
+            "lastTimestamp": lastTimestamp,
+            "message": message,
+            "metadata": {"name": meta["name"], "namespace": meta["namespace"],},
```
chancez: Remove the extra comma after `meta["namespace"]`

images/metering-ansible-operator/roles/meteringconfig/library/k8s_event.py
```diff
@@ -0,0 +1,349 @@
...
+            "reportingInstance": "",
+            "source": {"component": source},
+            "type": eventType,
+        }
+
+        print("count from CURRENT is %i" % event["count"])
+
+        #passes from metadata
+        definition = self.set_defaults(
+            resource, definition
```
chancez: These can be on the same line as `self.set_defaults(`

images/metering-ansible-operator/roles/meteringconfig/library/k8s_event.py
chancez: @timflannagan1 that's expected. These values will be unset if the involvedObject doesn't exist since we can't get them, but we still want to create the event, even if we can't get all the details of the involvedObject.

images/metering-ansible-operator/roles/meteringconfig/library/k8s_event.py
timflannagan1: I should've been clearer, I still think we want this as a variable, but replace all the "v1" references with `api_version` to avoid hardcoding any values.

images/metering-ansible-operator/roles/meteringconfig/library/k8s_event.py
timflannagan1: I realize that this implies that we need to manually set/update the resource version and uid if the event already exists.

chancez: Ah yes. Ill make a comment above on that.

images/metering-ansible-operator/roles/meteringconfig/library/k8s_event.py
```diff
@@ -0,0 +1,349 @@
...
+        involvedObjectUid = None
+        involvedObjectResource = self.find_resource(
+            involvedObj["kind"], "v1", fail=True)
+
+        if involvedObjectResource:
+            try:
+                apiInvolvedObject = resource.get(
+                    name=involvedObj["name"], namespace=involvedObj["namespace"])
+
+                involvedObjectUid = apiInvolvedObject["metadata"]["uid"]
```
chancez: This should set the correct fields in `involvedObj` instead of using new variables since it would remove a bunch of extra code you have below?

Eg:

```
involvedObject["uid"] = apiInvolvedObject["metadata"]["uid"]
involvedObject["resourceVersion"] = apiInvolvedObject["metadata"]["resourceVersion"]
```

images/metering-ansible-operator/roles/meteringconfig/library/k8s_event.py
chancez: Yeah change this to `involvedObject": involved_obj` once you've removed `involvedObjectResourceVersion` and `involvedObjectResourceVersion` like I mentioned above.

images/metering-ansible-operator/roles/meteringconfig/library/k8s_event.py
```diff
@@ -0,0 +1,329 @@
...
+        argumentSpec = copy.deepcopy(AUTH_ARG_SPEC)
+        argumentSpec.update(EVENT_ARG_SPEC)
+        return argumentSpec
+
+    def execute_module(self):
+        """ Module execution """
+        #auth
+        self.client = self.get_api_client()
+
+        # retrieve params
```
timflannagan1: I think it would be cleaner to initialize the constants here:
```python
api_version = 'v1'
event_kind_name = 'Event'
prior_event = None
prior_count = 1
```

What do you think?

images/metering-ansible-operator/roles/meteringconfig/library/k8s_event.py
```diff
@@ -0,0 +1,329 @@
...
+
+        meta = definition["metadata"]
+        meta["name"] = self.params.get("name")
+        meta["namespace"] = self.params.get("namespace")
+
+        involved_obj = self.params.get("involvedObject")
+        print(involved_obj["kind"], involved_obj["namespace"], involved_obj["name"], involved_obj["apiVersion"])
+
+        resource = self.find_resource(
+            "Event", "v1", fail=True)
```
timflannagan1:
```python
resource = self.find_resource(event_kind_name, api_version, fail=True)
```

images/metering-ansible-operator/roles/meteringconfig/library/k8s_event.py
```diff
@@ -0,0 +1,329 @@
...
+            prior_count = prior_event["count"] + 1
+            print("If event exists and reason is the same %i" % prior_count)
+            first_timestamp = prior_event["firstTimestamp"]
+            print("the value of the firstTimestamp event exists same reason is", first_timestamp)
+            last_timestamp = rfc
+            print("the value of the lastTimestamp event exists same reason is", last_timestamp)
+
+        # handle involvedObject
+        involved_object_resource = self.find_resource(
+            involved_obj["kind"], "v1", fail=True)
```
timflannagan1:
```python
involved_object_resource = self.involved_obj["kind"], api_version, fail=True)
```

images/metering-ansible-operator/roles/meteringconfig/library/k8s_event.py
```diff
@@ -0,0 +1,329 @@
...
+                    name=involved_obj["name"], namespace=involved_obj["namespace"])
+
+                involved_obj["uid"] = api_involved_object["metadata"]["uid"]
+                involved_obj["resourceVersion"] = api_involved_object["metadata"]["resourceVersion"]
+
+            except openshift.dynamic.exceptions.NotFoundError:
+                pass
+
+        event = {
+            "apiVersion": "v1",
```
timflannagan1:
```python
"apiVersion": api_version,
```

images/metering-ansible-operator/roles/meteringconfig/library/k8s_event.py
```diff
@@ -0,0 +1,329 @@
...
+            "count": prior_count,
+            "eventTime": None,
+            "firstTimestamp": first_timestamp,
+            "involvedObject": involved_obj,
+            "kind": "Event",
```
timflannagan1:
```python
"kind": event_kind_name,
```

images/metering-ansible-operator/roles/meteringconfig/library/k8s_event.py
```diff
@@ -0,0 +1,329 @@
...
+            "lastTimestamp": last_timestamp,
+            "message": message,
+            "metadata": {"name": meta["name"], "namespace": meta["namespace"]},
```
timflannagan1: I think this would be easier to read if the `metadata` key was multi-lined:
```python
            "metadata": {
                "name": meta["name"],
                "namespace": meta["namespace"],
             },
```

images/metering-ansible-operator/roles/meteringconfig/library/k8s_event.py
timflannagan1: I'm opened to suggestions on: `event_kind_name`

chancez: I don't particularly think we even raelly need to use variables for the apiVersion and kind. It doesn't make it any clearer, that's for sure. The amount of repetition is more since the value is shorter than the variable, and not every object shares an apiVersion so the re-use is minimal anyways.

images/metering-ansible-operator/roles/meteringconfig/library/k8s_event.py
chancez: I don't really think we need this

images/metering-ansible-operator/roles/meteringconfig/library/k8s_event.py
chancez: I don't really think we need this

images/metering-ansible-operator/roles/meteringconfig/library/k8s_event.py
timflannagan1: Fair enough, but we reference these values multiple times throughout the module, which is why I suggested we declare a variable for it.

EmilyM1: Going to keep ordering as is. It's clear. Retrieval parameters in one spot.

images/metering-ansible-operator/roles/meteringconfig/library/k8s_event.py
```diff
@@ -0,0 +1,328 @@
...
+    namespace: default
+    message: message here
+    reason: reason is now different againnnnnnnnnnnn
+    reportingComponent: reportingComponents here
+    type: Normal
+    source:
+    component: Metering components
+    involvedObject:
+      apiVersion: v1
+      kind: Event
```
chancez: This is a poor example of an involvedObject, can you pick a different `kind` like pod, service, etc and use a valid name? Names cannot have spaces in them so your example isn't valid.

images/metering-ansible-operator/roles/meteringconfig/library/k8s_event.py
```diff
@@ -0,0 +1,328 @@
...
+      type: string
+
+requirements:
+  - python >= 2.7
+  - openshift >= 0.6.2
+"""
+
+EXAMPLES = """
+
+- name: Create Kubernetes Event
```
chancez: Maybe generally for this example try to reproduce a real event that's generated in a real controller. A lot of the example values here aren't representative of a properly defined Event. Reason is a single word, CamelCase, message is often a useful description. Right now this has mostly place-holders so it's hard to figure out the proper conventions. Just do a `kubectl get events --all-namespaces -o yaml` and find one that seems like a good basis to try to recreate using the module.

images/metering-ansible-operator/roles/meteringconfig/library/k8s_event.py
```diff
@@ -0,0 +1,328 @@
...
+       type: str
+     kind:
+       description: Always 'Event'.
+       returned: success
+       type: str
+     metadata:
+       description: Standard object metadata. Includes name, namespace, annotations, labels, etc.
+       returned: success
+       type: complex
+     spec:
```
chancez: Events don't have a spec.

images/metering-ansible-operator/roles/meteringconfig/library/k8s_event.py
```diff
@@ -0,0 +1,328 @@
...
+       description: Specific attributes of the object. Will vary based on the I(api_version) and I(kind).
+       returned: success
+       type: complex
+     status:
```
chancez: Events don't have a status.

images/metering-ansible-operator/roles/meteringconfig/library/k8s_event.py
```diff
@@ -0,0 +1,328 @@
...
+    component: Metering components
+    involvedObject:
+      apiVersion: v1
+      kind: Event
+      name: involvedObject event names
+      namespace: default
+"""
+
+RETURN = """
+result:
```
chancez: This is missing most of the fields on an event. Also be careful when actually defining this since it may not match the actual API definition since the ansible code may change how fields are defined.

images/metering-ansible-operator/roles/meteringconfig/library/k8s_event.py
```diff
@@ -0,0 +1,328 @@
...
+        argumentSpec = copy.deepcopy(AUTH_ARG_SPEC)
+        argumentSpec.update(EVENT_ARG_SPEC)
+        return argumentSpec
+
+    def execute_module(self):
+        """ Module execution """
+        #auth
+        self.client = self.get_api_client()
+
+        # retrieve params
```
chancez: Unnecessary comment.

images/metering-ansible-operator/roles/meteringconfig/library/k8s_event.py
```diff
@@ -0,0 +1,328 @@
...
+        prior_event = None
+        try:
+            prior_event = resource.get(
+                name=meta["name"],
+                namespace=meta["namespace"])
+
+        except openshift.dynamic.exceptions.NotFoundError:
+            pass
+
+        # event counter
```
chancez: Unnecessary comment. Doesn't provide any additional clarity.

images/metering-ansible-operator/roles/meteringconfig/library/k8s_event.py
```diff
@@ -0,0 +1,328 @@
...
+
+        if prior_event and prior_event["reason"] == reason:
+            prior_count = prior_event["count"] + 1
+            print("If event exists and reason is the same %i" % prior_count)
+            first_timestamp = prior_event["firstTimestamp"]
+            print("the value of the firstTimestamp event exists same reason is", first_timestamp)
+            last_timestamp = rfc
+            print("the value of the lastTimestamp event exists same reason is", last_timestamp)
+
+        # handle involvedObject
```
chancez: Unnecessary comment. Doesn't provide any additional clarity.

images/metering-ansible-operator/roles/meteringconfig/library/k8s_event.py
```diff
@@ -0,0 +1,328 @@
...
+            "count": prior_count,
+            "eventTime": None,
+            "firstTimestamp": first_timestamp,
+            "involvedObject": involved_obj,
+            "lastTimestamp": last_timestamp,
+            "message": message,
+            "metadata": {
+                "name": meta["name"],
+                "namespace": meta["namespace"]
+                },
```
chancez: This bracket should match the indentation of `metadata`

images/metering-ansible-operator/roles/meteringconfig/library/k8s_event.py
```diff
@@ -0,0 +1,328 @@
...
+            "reason": reason,
+            "reportingComponent": reporting_component,
+            "reportingInstance": "",
+            "source": {"component": source},
+            "type": event_type,
+        }
+
+        print("count from CURRENT is %i" % event["count"])
+
+        #passes from metadata
```
chancez: Unnecessary comment. Doesn't provide any additional clarity.

images/metering-ansible-operator/roles/meteringconfig/library/k8s_event.py
```diff
@@ -0,0 +1,328 @@
...
+        definition = self.set_defaults(
```
chancez: This variable isn't used.

images/metering-ansible-operator/roles/meteringconfig/library/k8s_event.py
timflannagan1: misindented

images/metering-ansible-operator/roles/meteringconfig/library/k8s_event.py
timflannagan1:
```python
        definition = self.set_defaults(resource, definition)
```

images/metering-ansible-operator/roles/meteringconfig/library/k8s_event.py
EmilyM1: Should this match the EVENT_ARG_SPEC fields?

images/metering-ansible-operator/roles/meteringconfig/library/k8s_event.py
```diff
@@ -0,0 +1,339 @@
...
+    def execute_module(self):
+        """ Module execution """
+        #auth
+        self.client = self.get_api_client()
+
+        message = self.params.get("message")
+        reason = self.params.get("reason")
+        reporting_component = self.params.get("reportingComponent")
+        event_type = self.params.get("type")
+        source = self.params.get("source")
```
timflannagan1: @chancez I feel like we should define these in `__init__` - what do you think?

images/metering-ansible-operator/roles/meteringconfig/library/k8s_event.py
```diff
@@ -0,0 +1,339 @@
...
+    @property
+    def argspec(self):
+        """ argspec property builder """
+        argumentSpec = copy.deepcopy(AUTH_ARG_SPEC)
+        argumentSpec.update(EVENT_ARG_SPEC)
+        return argumentSpec
+
+    def execute_module(self):
+        """ Module execution """
+        #auth
```
timflannagan1: remove this comment

images/metering-ansible-operator/roles/meteringconfig/library/k8s_event.py
timflannagan1: This could also be relegated to a follow-up PR

images/metering-ansible-operator/roles/meteringconfig/library/k8s_event.py
```diff
@@ -0,0 +1,339 @@
...
+import kubernetes.config.dateutil
+import openshift
+
+from collections import defaultdict
+
+from ansible.module_utils.k8s.common import AUTH_ARG_SPEC
+from ansible.module_utils.k8s.raw import KubernetesRawModule
+
+# user args
+EVENT_ARG_SPEC = {
```
timflannagan1: I think we should remove this comment too, as this pattern of defining a dictionary of arguments is pretty common in modules.

images/metering-ansible-operator/roles/meteringconfig/library/k8s_event.py
EmilyM1: Follow up

chancez: What's here is fine

chancez:
This looks good to me. Can you squash these to a single commit?

PR #1016, opened 2019-11-05
[release-4.2] bug 1766341: hack: Fix mirroring scripts now that internal image registry is different

chancez:
Images from brew are now stored in quay and accessed through an internal
proxy.

### DIFF
--- a/hack/mirror-ose-image.sh
+++ b/hack/mirror-ose-image.sh
@@ -2,27 +2,25 @@
 
 set -e
 
-IMAGE_NAME=$1
-CLUSTER_REGISTRY_URL=$2
+SOURCE_IMAGE=$1
+OUTPUT_IMAGE=$2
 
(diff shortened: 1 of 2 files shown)


chancez:
/hold cancel
False alarm, was using the wrong branch when testing this again. Everything seems to work fine.

derekwaynecarr:
this doesnt appear to impact actual function, but CI only.  is this update already in master branch?  if so, can you link to it?

timflannagan1:
@derekwaynecarr https://github.com/operator-framework/operator-metering/pull/968

chancez:
@derekwaynecarr Correct, but it's used by QE to verify the changes since we are shipped via OLM. There's an arduous process to actually use the yet-to-be-shipped OLM packages and images which this basically automates. 

The reason it needs updating is because way images were named/published by ART all changed post-code freeze which broke the scripts.

PR #1017, opened 2019-11-05
MC-720 Making Prometheus connector available to Presto

bentito:
~WIP~
See Jira ticket for steps being encoded. But basically, updating templates to allow Presto to have default settings to access Prometheus; update CRD for same; update permissions via CRB/CR for same.


charts/openshift-metering/templates/presto/presto-prometheus-auth.yaml
```diff
@@ -0,0 +1,29 @@
...
+    app: presto
+roleRef:
+  kind: ClusterRole
+  name: {{ .Release.Namespace }}-presto-service-account
+  apiGroup: rbac.authorization.k8s.io
+subjects:
+  - kind: ServiceAccount
+    name: presto
+    namespace: {{ .Release.Namespace }}
+{{- end }}
```
timflannagan1: missing newline after line 29

images/metering-ansible-operator/roles/meteringconfig/tasks/reconcile_presto.yml
```diff
@@ -49,6 +49,9 @@
       - template_file: templates/presto/presto-serviceaccount.yaml
         apis: [ {kind: serviceaccount} ]
         prune_label_value: presto-serviceaccount
+      - template_file: templates/presto/presto-prometheus-auth.yaml
+        apis: [ {kind: serviceaccount} ]
+        prune_label_value: presto-prometheus-auth
```
timflannagan1: I think we may need a `create` field if this template file isn't going to be always used.

images/metering-ansible-operator/roles/meteringconfig/defaults/main.yml
```diff
@@ -362,6 +362,15 @@ _ocp_enabled_overrides:
       config:
         prometheus:
           url: "https://prometheus-k8s.openshift-monitoring.svc:9091/"
+  presto:
+    spec:
+      config:
+        connectors:
+          prometheus:
+            enabled: true
+            uri: "https://prometheus-k8s.openshift-monitoring.svc:9091/"
```
timflannagan1: I think we may need to add a validation task that ensures this field isn't empty. That being said, we do validate the `reporting-operator.spec.config.prometheus.url` isn't empty when the ocp-only features are disabled.

charts/openshift-metering/templates/presto/presto-prometheus-auth.yaml
bentito: Done in ab1d035

charts/openshift-metering/templates/presto/_presto_helpers.tpl
```diff
@@ -45,6 +45,35 @@ connector.name=blackhole
 connector.name=memory
 {{ end }}
 
+{{- define "presto-prometheus-catalog-properties" -}}
```
timflannagan1: Should this be wrapped in a if condition, e.g.:
```yaml
{{- if .Values.presto.spec.config.connectors.prometheus.enabled }}
{{- define ... }}
```

images/metering-ansible-operator/roles/meteringconfig/tasks/reconcile_presto.yml
bentito: Done in 1a067bf

images/metering-ansible-operator/roles/meteringconfig/tasks/reconcile_presto.yml
```diff
@@ -49,6 +49,10 @@
...
+        create: "{{ meteringconfig_create_presto_prometheus_auth }}"
```
timflannagan1: Hmm, I don't see this defined anywhere - if that's the case then just add this variable to the `defaults/main.yml` file somewhere above the `_ocp_enabled_overrides` dictionary, e.g.:
```yaml
meteringconfig_create_presto_prometheus_auth: "{{ _presto_spec.config.connectors.prometheus.enabled | default(false) }}"
```

charts/openshift-metering/templates/presto/_presto_helpers.tpl
bentito: yeah, seems like otherwise it will be there doing nothing when not enabled. Done in ce5e2ad

bentito: Hmm, but with a change like that, helm template is puking on the location of the define now...

bentito: Swapped order of define/if, helm template seems good with that. Done in 6315e14

images/metering-ansible-operator/roles/meteringconfig/tasks/reconcile_presto.yml
bentito: Done in 622ca1e, still passes `make verify`

images/metering-ansible-operator/roles/meteringconfig/defaults/main.yml
bentito: I think doing this only in the CRD was what we decided in Slack, right, so I am going to do that work on branch: `MC-720-crd-manifests-only`

timflannagan1: Yeah moving the validation from ansible-side to the CRD-side.

charts/metering-ansible-operator/templates/crds/meteringconfig.crd.yaml
```diff
@@ -1549,6 +1549,47 @@ spec:
...
+                                chunkSizeDuration:
+                                  description: Default size of each Prometheus query chunk.
+                                  type: string
+                                maxQueryRangeDuration:
+                                  description: Default range for Prometheus query, divided into chunkSizeDuration chunks.
+                                  type: string
+                                cacheDuration:
+                                  description: Time to live for Presto checking connector properties values.
+                                  type: string
+                                bearer:
```
bentito: Copying this discussion over so it can be addressed in this PR:
timflannagan1 4 days ago Member
@chancez do you think this should just bearerTokenFile for the time being, or do you think there will be more authentication/bearer configurations in the future that warrants it's own section?

 @chancez
chancez 4 days ago Member
Good question. I guess generally speaking, being able to specify a path directly isn't very useful actually since it'll be coming either from the service account secret, or an arbitrary user supplied secret.

In the case of a user supplied secret, they wouldn't likely have much choice in the location of the file, and we would just mount it to a known location.

So perhaps we should have an. A useServiceAccountToken bool option and eventually (not now) we could add a secretName config for supplying your own credentials.

WDYT?

 @chancez
chancez 4 days ago Member
Good question. I guess generally speaking, being able to specify a path directly isn't very useful actually since it'll be coming either from the service account secret, or an arbitrary user supplied secret.

In the case of a user supplied secret, they wouldn't likely have much choice in the location of the file, and we would just mount it to a known location.

So perhaps we should have an. A useServiceAccountToken bool option and eventually we could add a secretName config for supplying your own credentials.

images/metering-ansible-operator/roles/meteringconfig/defaults/main.yml
```diff
@@ -362,6 +365,15 @@ _ocp_enabled_overrides:
       config:
         prometheus:
           url: "https://prometheus-k8s.openshift-monitoring.svc:9091/"
+  presto:
```
timflannagan1: nit: can you add a newline between the reporting-operator prometheus URL section and the following presto section?

bentito: Fixed in acfb7a89

charts/metering-ansible-operator/templates/crds/meteringconfig.crd.yaml
bentito: Made the chart changes for this in: c64a8ec1
Thinking it's likely we don't want to check `useServiceAccountToken` on the ansible validation side until we provide ability to specify an alternative token location.
Should we have a story for that so we don't forget to do it?

images/metering-ansible-operator/roles/meteringconfig/defaults/main.yml
```diff
@@ -363,6 +366,16 @@ _ocp_enabled_overrides:
...
 
+  presto:
+    spec:
+      config:
+        connectors:
+          prometheus:
+            enabled: true
+            uri: "https://prometheus-k8s.openshift-monitoring.svc:9091/"
+            bearer:
+              tokenFile: "/var/run/secrets/kubernetes.io/serviceaccount/token"
```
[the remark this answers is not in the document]
bentito: Done in 18abd7a6

charts/openshift-metering/values.yaml
```diff
@@ -544,6 +544,14 @@ presto:
...
 
         extraConnectorFiles: []
+        prometheus:
+          enabled: false
+          uri: ""
+          chunkSizeDuration: null
+          maxQueryRangeDuration: null
+          cacheDuration: null
+          bearer:
+            tokenFile: null
```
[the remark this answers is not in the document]
bentito: Done in 18abd7a6

charts/openshift-metering/templates/presto/_presto_helpers.tpl
```diff
@@ -45,6 +45,37 @@ connector.name=blackhole
...
+max-query-range-duration=1d
+{{- end }}
+{{- if .cacheDuration }}
+cache-duration={{ .cacheDuration }}
+{{- else }}
+cache-duration=30s
+{{- end }}
+{{- if .bearer.tokenFile }}
+bearer-token-file={{ .bearer.tokenFile }}
+{{- end }} {{- /* end-if */ -}}
```
timflannagan1: Hmm, we probably want to update this (and the *-values.yaml) to remove this conditional and use an `emptyDir` to mount the applicable service account when `{{- .useServiceAccountToken }}` is set to true.

bentito: Done in 21d5c091

charts/openshift-metering/values.yaml
```diff
@@ -578,9 +587,6 @@ presto:
         logLevel: info
         taskMaxWorkerThreads: null
         taskMinDrivers: null
-        logLevel: info
-        taskMaxWorkerThreads: null
-        taskMinDrivers: null
```
timflannagan1: Why were these coordinator config options removed?

charts/openshift-metering/values.yaml
```diff
@@ -544,6 +544,15 @@ presto:
             useInstanceCredentials: null
 
         extraConnectorFiles: []
+        prometheus:
+          enabled: true
+          uri: ""
+          chunkSizeDuration: null
+          maxQueryRangeDuration: null
+          cacheDuration: null
```
timflannagan1: I think we should nest lines 550-552 under `config`, (and update the charts) i.e.:

```yaml
config:
  chunkSizeDuration: null
  maxQueryRangeDuration: null
  cacheDuration: null
```

wdyt?

images/metering-ansible-operator/roles/meteringconfig/defaults/main.yml
```diff
@@ -363,6 +366,17 @@ _ocp_enabled_overrides:
...
+            useServiceAccountToken: true
```
timflannagan1:
```yaml
auth:
  bearerTokenFile: "/var/run/..."
  useServiceAccountToken: true
```

I'm not sure we need to specify the `useServiceAccountToken` either as the default value is set to `true`.

charts/openshift-metering/values.yaml
bentito: they're dupes, linting found them, see right above?

charts/openshift-metering/values.yaml
bentito: well if we do that it should include `uri` too, as with that one under config: we'd have the full grouping of presto-prom connector configs, it would call them out as separate from `enabled`, which is not.

charts/openshift-metering/values.yaml
timflannagan1: Oh wow, I guess those were introduced when we moved the hive configuration to it's section.

images/metering-ansible-operator/roles/meteringconfig/defaults/main.yml
bentito: in `values.yaml` right? If so, then yes, I agree we don't need to set it here.

charts/openshift-metering/values.yaml
bentito: Done in 04273d69

images/metering-ansible-operator/roles/meteringconfig/defaults/main.yml
timflannagan1: Yes, in `values.yaml`

charts/metering-ansible-operator/templates/crds/meteringconfig.crd.yaml
timflannagan1: Yeah let's hold off on creating a story for now.

images/metering-ansible-operator/roles/meteringconfig/defaults/main.yml
timflannagan1:
```suggestion
            auth:
              bearerTokenFile: "/var/run/secrets/kubernetes.io/serviceaccount/token"
```

images/metering-ansible-operator/roles/meteringconfig/defaults/main.yml
bentito: Done in last several commits.

charts/openshift-metering/values.yaml
```diff
@@ -544,6 +544,16 @@ presto:
             useInstanceCredentials: null
 
         extraConnectorFiles: []
+        prometheus:
+          enabled: true
```
timflannagan1: Hmm the more I think about this, I feel that we should disable this connector by default until the connector is a supported configuration option. This would also make it easier to mock in the e2e suite as we'd be explicitly enabling the connector.

bentito: I think it's easier to test initial functionality, like we're doing right now, if it's supposed to be enabled and working by default. Also the connector is the only way to get info from Prometheus, so `not enabled` is exactly the same as `not present`, so I'm not sure I see the benefits for mocking?

timflannagan1: Okay, that's understandable, and I realize that there are many steps to actually enable the workflow, and this PR just exposes another connector to the metering stack. With that said, I think we can keep enabling it by default then.

charts/metering-ansible-operator/templates/crds/meteringconfig.crd.yaml
```diff
@@ -1590,6 +1590,9 @@ spec:
                                     tokenFile:
```
timflannagan1: We now need to update the MeteringConfig CRD values for the Prometheus connector now that we've decided on a general layout (then run `make metering-manifests` to re-generate the manifests.)

images/metering-ansible-operator/roles/meteringconfig/defaults/main.yml
bentito: fixed in: 3b42b241

charts/openshift-metering/templates/presto/_presto_helpers.tpl
```diff
@@ -45,6 +45,37 @@ connector.name=blackhole
...
+max-query-range-duration={{ .config.maxQueryRangeDuration }}
+{{- else }}
+max-query-range-duration=1d
+{{- end }}
+{{- if .config.cacheDuration }}
+cache-duration={{ .config.cacheDuration }}
+{{- else }}
+cache-duration=30s
+{{- end }}
+{{- if .auth.useServiceAccountToken }}
```
timflannagan1: Hmm, this seems off now too, as we can set the bearer-token-file through the `auth.useServiceAccountToken` as-well-as explicitly defining the path to the token file via `auth.bearerTokenFile`.

Should it be something like this:
```yaml
{{- if or .auth.useServiceAccountToken .auth.bearerTokenFile }}
bearer-token-file={{ .auth.bearerTokenFile }}
```

bentito: Fixed in: 0c29d3ad

charts/metering-ansible-operator/templates/crds/meteringconfig.crd.yaml
bentito: updated in 7dc46406

charts/metering-ansible-operator/templates/crds/meteringconfig.crd.yaml
```diff
@@ -1571,23 +1571,26 @@ spec:
...
+                                      description: Time to live for Presto checking connector properties values.
+                                      type: string
+                                auth:
                                   type: object
                                   properties:
-                                    tokenFile:
+                                    bearerTokenFile:
                                       description: Path to token file for bearer authentication to Prometheus.
                                       type: string
                                 useServiceAccountToken:
```
timflannagan1: This needs to be under `prometheus.auth`.

bentito: Fixed in: 438ae4ea

charts/openshift-metering/templates/presto/_presto_helpers.tpl
```diff
@@ -45,6 +45,37 @@ connector.name=blackhole
 connector.name=memory
 {{ end }}
 
+{{- define "presto-prometheus-catalog-properties" -}}
+{{- if .Values.presto.spec.config.connectors.prometheus.enabled }}
```
timflannagan1: After taking a look at how we manage other connectors, any "empty" (or unused if I understand correctly) have a `<connector-name>.properties` file, that contains the name of the connector. We could modify the above to be something like this, to conform with the rest of the connectors:

```yaml
{{- define "presto-prometheus-catalog-properties" -}}
connector.name=prometheus
{{- if .Values.presto.spec.config.connectors.prometheus.enabled }}
...
```

wdyt?

images/metering-ansible-operator/roles/meteringconfig/defaults/main.yml
```diff
@@ -363,6 +366,17 @@ _ocp_enabled_overrides:
...
 
+  presto:
+    spec:
+      config:
+        connectors:
+          prometheus:
+            enabled: true
+            config:
+              uri: "https://prometheus-k8s.openshift-monitoring.svc:9091/"
+            auth:
```
timflannagan1: nitpick: let's create a newline between the `config` and `auth` sections.

charts/openshift-metering/templates/presto/_presto_helpers.tpl
bentito: So that we still `prometheus.properties` file with just `prometheus` in it even when the connector is not enabled? I don't think that is right from a Presto standpoint. If the connector is not enabled I think there would be no file present in `.../etc/catalog/`

image

images/metering-ansible-operator/roles/meteringconfig/defaults/main.yml
bentito: Looking around just in `main.yml` I don't see that convention in use throughout. Should we try to start having some standard for vertical whitespace in yaml? We could start by making that whole file conform. I'm up for that.

Best would be if we had some tool that could format and enforce whatever we want so all the yaml kinda starts looking more the same at least within our own project. Not just whitespace, but ordering of fields might be helpful too.

The Presto project does this for Java code via a Maven plugin that gets run in the verify phase of compile. For us it could be part of the make driven stuff I'd think.

charts/openshift-metering/templates/presto/_presto_helpers.tpl
timflannagan1: Hmm, I'm seeing this when I disable the prometheus connector:

```bash
tflannag@localhost operator-metering [bentito-MC-720]
└─ $ ▶ k exec -it presto-coordinator-0 -- /bin/bash
Defaulting container name to presto.
Use 'kubectl describe pod/presto-coordinator-0 -n tflannag' to see all of the containers in this pod.
bash-4.2$ cat /opt/presto/presto-server/etc/
catalog/                                config.properties                       jvm.config                              log.properties                          node.properties                         prometheus-connector-bearer-token-file  
bash-4.2$ cat /opt/presto/presto-server/etc/catalog/
..2019_12_11_18_26_16.752759064/ ..data/                          blackhole.properties             hive.properties                  jmx.properties                   memory.properties                prometheus.properties            tpcds.properties                 tpch.properties                  
bash-4.2$ cat /opt/presto/presto-server/etc/catalog/prometheus.properties 
connector.name=prometheus
prometheus-uri=https://prometheus-k8s.openshift-monitoring.svc:9091/
query-chunk-size-duration=1h
max-query-range-duration=1d
cache-duration=30s
bearer-token-file=/var/run/secrets/kubernetes.io/serviceaccount/tokenbash-4.2$ cat /opt/presto/presto-server/etc/catalog/memory.properties     
connector.name=memory
bash-4.2$ 
bash-4.2$ 
bash-4.2$ cat /opt/presto/presto-server/etc/catalog/memory.properties 
connector.name=memory
bash-4.2$ exit
exit
tflannag@localhost operator-metering [bentito-MC-720]
└─ $ ▶ bat $METERING_CR_FILE
───────┬────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
       │ File: custom_config/openshift.yaml
───────┼────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
   1   │ apiVersion: metering.openshift.io/v1
   2   │ kind: MeteringConfig
   3   │ metadata:
   4   │   name: "operator-metering"
   5   │ spec:
   6   │   unsupportedFeatures:
   7   │     enableHDFS: true
   8   │ 
   9   │   storage:
  10   │     type: "hive"
  11   │     hive:
  12   │       type: "hdfs"
  13   │       hdfs:
  14   │         namenode: "hdfs-namenode-0.hdfs-namenode:9820"
  15   │ 
  16   │   presto:
  17   │     spec:
  18   │       config:
  19   │         connectors:
  20   │           prometheus:
  21   │             enabled: false
───────┴────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
tflannag@localhost operator-metering [bentito-MC-720]
└─ $ ▶ 
```

timflannagan1: The `prometheus.properties` file is still being populated despite being disabled. Also, It looks like we may need to add a trailing newline to their `prometheus.properties` file.

images/metering-ansible-operator/roles/meteringconfig/defaults/main.yml
timflannagan1: I was under the impression that we did have some sort of convention used throughout the default values file as we do in the MeteringConfig values.yaml. I think going forward we may need to clean up this ever-growing default values file.

bentito: anyways.. I just added the one space for now. Done in 018461af

charts/openshift-metering/templates/presto/_presto_helpers.tpl
bentito: is this just a result of ocp default being enabled?

charts/openshift-metering/values.yaml
timflannagan1: If we're enabling the Prometheus-connector by default for OCP, and set `enabled: true` in the OCP enabled overrides dictionary, then we should disable this connector by default, in the MeteringConfig values.

charts/openshift-metering/templates/presto/_presto_helpers.tpl
```diff
@@ -45,6 +45,37 @@ connector.name=blackhole
...
+cache-duration={{ .config.cacheDuration }}
+{{- else }}
+cache-duration=30s
+{{- end }}
+{{- if or .auth.useServiceAccountToken .auth.bearerTokenFile }}
+bearer-token-file={{ .auth.bearerTokenFile }}
+{{- end }} {{- /* end-if */ -}}
+{{- end }} {{- /* end-with */ -}}
+{{- end }} {{- /* end-if-enabled */ -}}
+{{- end }} {{- /* end-define */ -}}
```
timflannagan1: When viewing the `prometheus.properties` in the presto-coordinator container, there's no trailing newline like the other connector properties files. Replacing `{{- end }}` with `{{ end }}` from lines 74 to 77 will fix this problem.

charts/openshift-metering/templates/presto/_presto_helpers.tpl
bentito: Added prom conn to metering_config.go so the makefile deploy will track other changes in this PR in: 966820f6

charts/openshift-metering/values.yaml
bentito: Done in 8962ea0b

charts/openshift-metering/templates/presto/_presto_helpers.tpl
bentito: Done in d98d9846.

PR #1019, opened 2019-11-06
bug 1755606: Verify that the s3 credentials secret exists.

timflannagan1:


images/metering-ansible-operator/roles/meteringconfig/tasks/configure_storage.yml
```diff
@@ -24,15 +24,22 @@
...
         state: present
         name: "{{ meteringconfig_storage_s3_bucket_name }}"
         region: "{{ meteringconfig_storage_s3_bucket_region }}"
         aws_access_key: "{{ operator_s3_credentials_secret.resources[0].data['aws-access-key-id'] | b64decode | default(omit) }}"
         aws_secret_key: "{{ operator_s3_credentials_secret.resources[0].data['aws-secret-access-key'] | b64decode | default(omit) }}"
-      when: operator_s3_credentials_secret.resources | length > 0
-
+      when: s3_credentials_secret_exists
+  vars:
+    s3_credentials_secret_exists: "{{ operator_s3_credentials_secret.resources and operator_s3_credentials_secret.resources | length > 0 }}"
```
chancez: It's weird to me this works because `operator_s3_credentials_secret` is defined in the inner scope of the block that this `var` uses. But it works because it's lazily evaluated. ¯\_(ツ)_/¯ idk what to think of this.

chancez: I guess this is fine. Anything is is just more verbose.

PR #1021, opened 2019-11-07
Add ability to set Prometheus Presto connector configuration options in MeteringConfig CRD

bentito:


chancez:
/approve

waiting on LGTM from @timflannagan1

manifests/deploy/ocp-testing/olm/bundle/4.3/meteringconfig.crd.yaml
```diff
@@ -1548,6 +1548,29 @@ spec:
...
+                                maxQueryRangeDuration:
+                                  description: Default range for Prometheus query, divided into chunkSizeDuration chunks.
+                                  type: string
+                                cacheDuration:
+                                  description: Time to live for Presto checking connector properties values.
+                                  type: string
+                                bearer:
+                                  type: object
+                                  properties:
+                                    tokenFile:
```
timflannagan1: @chancez do you think this should just `bearerTokenFile` for the time being, or do you think there will be more authentication/bearer configurations in the future that warrants it's own section?

manifests/deploy/ocp-testing/olm/bundle/4.3/meteringconfig.crd.yaml
```diff
@@ -1548,6 +1548,29 @@ spec:
                                   properties:
                                     useInstanceCredentials:
                                       type: boolean
+                            prometheus:
+                              type: object
+                              properties:
+                                enabled:
+                                  type: boolean
```
timflannagan1: I think it would be helpful to add a description field for this property too.

timflannagan1:
None of my comments really justify holding off merging this
/lgtm

manifests/deploy/ocp-testing/olm/bundle/4.3/meteringconfig.crd.yaml
chancez: Good question. I guess generally speaking, being able to specify a path directly isn't very useful actually since it'll be coming either from the service account secret, or an arbitrary user supplied secret.

In the case of a user supplied secret, they wouldn't likely have much choice in the location of the file, and we would just mount it to a known location.

So perhaps we should have an. A `useServiceAccountToken` bool option and eventually (not now) we could add a `secretName` config for supplying your own credentials.

WDYT?

chancez: Good question. I guess generally speaking, being able to specify a path directly isn't very useful actually since it'll be coming either from the service account secret, or an arbitrary user supplied secret.

In the case of a user supplied secret, they wouldn't likely have much choice in the location of the file, and we would just mount it to a known location.

So perhaps we should have an. A `useServiceAccountToken` bool option and eventually we could add a `secretName` config for supplying your own credentials.

charts/metering-ansible-operator/templates/crds/meteringconfig.crd.yaml
```diff
@@ -1549,6 +1549,29 @@ spec:
...
                                     useInstanceCredentials:
                                       type: boolean
+                            prometheus:
+                              type: object
+                              properties:
+                                enabled:
+                                  type: boolean
+                                uri:
+                                  description: URI for Prometheus.
+                                  type: string
```
timflannagan1: `format: uri`

charts/metering-ansible-operator/templates/crds/meteringconfig.crd.yaml
```diff
@@ -1549,6 +1549,29 @@ spec:
...
+                                cacheDuration:
+                                  description: Time to live for Presto checking connector properties values.
+                                  type: string
+                                bearer:
+                                  type: object
+                                  properties:
+                                    tokenFile:
+                                      description: Path to token file for bearer authentication to Prometheus.
+                                      type: string
```
timflannagan1: We may be able to get away with just validating that the `uri` property is set when `enabled: true`, and not set when `enabled: false`, e.g.:

```yaml
                              oneOf:
                              - required:
                                - uri
                                properties:
                                  enabled:
                                    enum:
                                    - true
                              - allOf:
                                - not:
                                    required:
                                    - uri
                                properties:
                                  enabled:
                                    enum:
                                    - false
```

manifests/deploy/ocp-testing/olm/bundle/4.3/meteringconfig.crd.yaml
bentito: Done in 42681d7

charts/metering-ansible-operator/templates/crds/meteringconfig.crd.yaml
bentito: Done in: 42681d7

charts/metering-ansible-operator/templates/crds/meteringconfig.crd.yaml
bentito: In 42681d7 we add another subschema to match enabled: false with there not being a uri defined. We also define format uri for uri.

manifests/deploy/ocp-testing/olm/bundle/4.3/meteringconfig.crd.yaml
bentito: Copied this comment set to https://github.com/operator-framework/operator-metering/pull/1017 comments.

PR #1026, opened 2019-11-12
images: Update the k8s_status module to better handle merging conditions.

timflannagan1:
### Description: 
When attempting to patch existing conditions, the `k8s_status` module will merge existing conditions with the conditions introduced in the most recent reconcile period. 

In order to do this, the module iterates over the new conditions list, checking if the current condition iteration's type matches an index in the existing conditions list, and if true, returns the corresponding index in the existing conditions list. 

This can be problematic as the `get_condition_idx` function returns the first match, and in the case where there are multiple existing conditions with the same type, returns a list containing multiple conditions with the same type.

This PR would update the merging process by filtering out any existing conditions that match the current, new condition when iterating over the list of new conditions. This ensures that there will only be a single condition type when the merging process has finished.

### DIFF
--- a/images/metering-ansible-operator/roles/meteringconfig/library/k8s_status.py
+++ b/images/metering-ansible-operator/roles/meteringconfig/library/k8s_status.py
@@ -328,19 +328,24 @@ class KubernetesAnsibleStatusModule(KubernetesAnsibleModule):
         merged = copy.deepcopy(old_conditions)
 
         for condition in new_conditions:
-            idx = self.get_condition_idx(merged, condition['type'])
-            if idx is not None:
-                merged[idx] = condition
-            else:
-                merged.append(condition)
+          new_merged_list = self.filter_existing_conditions(merged, condition['type'])
+          new_merged_list.append(condition)
+
+          merged = new_merged_list
+
         new['conditions'] = merged
         return new
 
-    def get_condition_idx(self, conditions, name):
-        for i, condition in enumerate(conditions):
-            if condition.get('type') == name:
-                return i
-        return None
+    def filter_existing_conditions(self, conditions, condition_type):
+        new_conditions = []
+
+        for condition in conditions:
+            if condition.get('type') == condition_type:
+                continue
+
+            new_conditions.append(condition)
+
+        return new_conditions
 
     def object_contains(self, obj, subset):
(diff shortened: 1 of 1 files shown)


images/metering-ansible-operator/roles/meteringconfig/library/k8s_status.py
```diff
@@ -328,14 +328,25 @@ def merge_status(self, old, new):
...
+
+          merged = new_merged_list
+
         new['conditions'] = merged
         return new
 
+    def filter_existing_conditions(self, conditions, type):
+        new_conditions = []
+
+        for _, condition in enumerate(conditions):
```
chancez: I don't think we need enumerate at this point?

images/metering-ansible-operator/roles/meteringconfig/library/k8s_status.py
```diff
@@ -328,14 +328,25 @@ def merge_status(self, old, new):
...
-                merged.append(condition)
+          new_merged_list = self.filter_existing_conditions(merged, condition['type'])
+          new_merged_list.append(condition)
+
+          merged = new_merged_list
+
         new['conditions'] = merged
         return new
 
+    def filter_existing_conditions(self, conditions, type):
```
chancez: `type` is a reserved word, maybe `conditionType` to avoid that

images/metering-ansible-operator/roles/meteringconfig/library/k8s_status.py
```diff
@@ -328,14 +328,25 @@ def merge_status(self, old, new):
...
+
+        for _, condition in enumerate(conditions):
+            if condition.get('type') == type:
+                continue
+
+            new_conditions.append(condition)
+
+        return new_conditions
+
     def get_condition_idx(self, conditions, name):
```
chancez: This might be unused now?

timflannagan1: Yeah, it's unused now, I wasn't sure if there's a need to keep this function around in the toolset.

chancez: Prob safe to remove

chancez:
lgtm, can you squash the last 3 commits together?

PR #1027, opened 2019-11-13
bug 1763303: Fix handling of AdditionalArtifactKeys in AWS billing report manifests

chancez:
When testing previous fixes for bug 1763303 on a new billing bucket I
discovered another bug related to how we expect the AdditionalArtifactKeys
field to be represented.

This was not caught previously due to not enabling any additional features in
the AWS billing reports we didn't need, but were set in the new billing report
being used which causes this field to be present in the AWS billing report JSON
manifests.

Also adds a commit to improve logging since I was having trouble figuring out
these issues originally without the improved logging.

### DIFF
--- a/pkg/aws/client.go
+++ b/pkg/aws/client.go
@@ -10,6 +10,7 @@ import (
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/s3"
 	"github.com/aws/aws-sdk-go/service/s3/s3iface"
+	log "github.com/sirupsen/logrus"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


pkg/aws/client.go
```diff
@@ -58,13 +61,24 @@ func (r *manifestRetriever) RetrieveManifests() ([]*Manifest, error) {
...
 		prefix += "/"
 	}
+	logger := r.logger.WithFields(log.Fields{
+		"bucket": r.bucket,
+		"prefix": prefix,
+	})
 
 	var manifests []*Manifest
 	var manifestErr error
+	page := 0
```
timflannagan1: `var page int`

chancez: Fixed

chancez:
/hold
Working on our CI configs.

PR #1032, opened 2019-11-14
Makefile: Remove the integration targets.

timflannagan1:
### Description:

This would remove the integration-related Makefile targets as those are no longer used after the e2e rewrite has been merged.


timflannagan1:
Need to wait for https://github.com/openshift/release/pull/5925 to merge first.
/hold

PR #1037, opened 2019-11-19
cmd,pkg,test: Move the deploy package down a directory level.

timflannagan1:
This has always been referred to as `pkg/deploy` in PR's, and the deploy package makes more sense to me living outside of the operator-specific package code.

This presumably would also require updating any existing deploy-metering local binaries via `make bin/deploy-metering`.


bentito:
I don't full understand the PR comment: ```This presumably would also require updating any existing deploy-metering binaries via make bin/deploy-metering.``` Does this mean work that should be done in some other PR? It almost sounds like if we merge this PR it does not have enough to allow said binaries to continue to be built correctly?

timflannagan1:
No, the `make bin/deploy-metering` would need to be run locally. The deploy-metering CLI tool is also not the preferred way of installing metering over the hack install scripts at the moment (and the CLI option is also not currently documented either), so I'm not too worried about making this change.

timflannagan1:
Just to be clear, we don't manage any CLI binaries in the operator-metering repo but build them locally when necessary, e.g. `make run-reporting-operator-local` will build the `bin/reporting-operator-local` binary.

PR #1045, opened 2019-12-02
Update repository to use release 4.4.

### LINKED ISSUE #1044: Please prepare bundle for OCP 4.4
Now that we are fully branched for 4.4, please prepare your operator to supply a 4.4 bundle, so that 4.4 operator publishing works and doesn't overwrite 4.3 bundles. This means at least updating the package.yaml under https://github.com/operator-framework/operator-metering/tree/master/manifests/deploy/openshift/olm/bundle
@bentito @timflannagan1

timflannagan1:
This would update the charts/manifests/Documentation etc. to the correct release after the branch cut completed a couple of weeks ago.

This would fix #1044


sosiouxme:
thanks, i would say LGTM but i wouldn't want it to roll through without real peer review :) @bentito @EmilyM1 how about it?

PR #1046, opened 2019-12-02
Dockerfile,images: Manage the MeteringConfig status directly.

timflannagan1:
### Description

This would update the metering-ansible-operator to manage the `MeteringConfig` status directly, as opposed to the ansible-operator doing it for us. That would involve setting the `manageStatus: false` option in the watches.yml file, and using the `k8s_status` periodically throughout the role.

Also included in the PR is an update to the metering-ansible-operator Dockerfile that ensures the Ansible version is set to 2.8 for now as there was a regression in 2.9 where the `ansible_failed_result` and `ansible_failed_task` variables were undefined in a block/rescue task, which is not ideal.

### DIFF
--- a/Dockerfile.metering-ansible-operator
+++ b/Dockerfile.metering-ansible-operator
@@ -25,7 +25,10 @@ RUN ln -f -s /usr/bin/oc /usr/bin/kubectl
 RUN ln -f -s /tini /usr/bin/tini
 
 RUN pip install --no-cache-dir --upgrade openshift boto3 cryptography
-RUN pip install --upgrade ansible>=2.8
+# In order to use the 'ansible_failed_result' and 'ansible_failed_task' variables in a block/rescue,
+# we need to ensure that the 2.8 version is being used while this is fixed in 2.9 upstream.
+# TODO: revert this change once the issue mentioned above is resolved.
+RUN pip install --upgrade ansible==2.8
 
 ENV HOME /opt/ansible
 ENV HELM_CHART_PATH ${HOME}/charts/openshift-metering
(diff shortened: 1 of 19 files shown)


Dockerfile.metering-ansible-operator
```diff
@@ -25,7 +25,7 @@ RUN ln -f -s /usr/bin/oc /usr/bin/kubectl
 RUN ln -f -s /tini /usr/bin/tini
 
 RUN pip install --no-cache-dir --upgrade openshift boto3 cryptography
-RUN pip install --upgrade ansible>=2.8
+RUN pip install --upgrade ansible==2.8
```
bentito: Might be good to add the commit message as a comment here. Later viewers of this segment might not know why we're locking it to this version.

timflannagan1: Yeah, I was thinking about adding a TODO comment, but I like this better.

EmilyM1:
Can you squash?

timflannagan1:
@EmilyM1 all of those commits are distinct changes that deserve their own commit imo.

images/metering-ansible-operator/roles/meteringconfig/tasks/validate_reporting_operator_tls.yml
```diff
@@ -10,6 +10,17 @@
       that:
         - meteringconfig_spec['reporting-operator'].spec.config.presto.tls.caCertificate != ""
       msg: "reporting-operator.spec.config.presto.tls.caCertificate cannot be empty if createSecret: true and secretName != ''"
+  rescue:
```
bentito: Ansible noob, so bear with me: Would it be possible to pull a lot of these tasks under one `block:` statement so that the `rescue:` could apply to all and not be repeated so much?

timflannagan1: In this case, that isn't something that needs to be reflected in the status, and only in the case of a failure (i.e. the `rescue` block) do we care about what happened.

In other cases, unfortunately not - it requires a lot more work to be able to track what's currently happening, and what went wrong (especially when looping over resources).

PR #1047, opened 2019-12-02
test/deployframework: Fix the logging message for where the metering-operator local logs are stored.

timflannagan1:
Before, the testing suite would log the full `os.File` object returned from creating a file, instead of the path to the file created which contains a lot of unnecessary artifacts.


EmilyM1:
Looks good. Any reason to not throw this change in with the 1046 open PR?

PR #1052, opened 2019-12-05
manifests: Stop maintaining a 4.3 bundle for the release-4.4.

timflannagan1:
### Description:

This would remove all the 4.3 OLM bundle manifests for the master/release-4.4 branches as we can track a single OLM manifest bundle to the applicable release-branch, e.g. 4.3 -> release-4.3, and we most likely don't want to maintain multiple sets of bundles per a release.


bentito:
checked out. seems like clean 4.3 removal to me

PR #1057, opened 2019-12-10
charts,manifests: Update the csv's container image field to use the 4.4 tag instead of latest.

timflannagan1:
All the images listed in a given CSV use the appropriate release tag, and the `containerImage` field was using `latest` which may be disrupting ART from substituting the image contained in this field to the correct image-registry ones in distgit.


timflannagan1:
This passed e2e locally, so force merging.

PR #1062, opened 2019-12-11
OWNERS: Update the release-4.2 owners.

timflannagan1:


timflannagan1:
@bparees just to clarify, this doesn't require a BZ to be filed to be merged, right?

bparees:
> @bparees just to clarify, this doesn't require a BZ to be filed to be merged, right?

correct, i'll manually apply the "valid-bz" label to bypass it.

PR #1063, opened 2019-12-16
hack,Makefile,test: Support running the e2e suite in a more dev-friendly configuration.

timflannagan1:
As we continue working with the refactored e2e suite, adding any additional e2e manual install tests can be cumbersome. This is because it takes a while to deploy a full install, wait for those resources to become ready to run reporting tests, then teardown any provisioned resources created. In the case where an error has occurred, then you would need to re-run the process outlined above, which is not ideal.

This PR would introduce the option to skip any tasks that uninstall the metering stack, or deletes namespaces containing the e2e label. With this change, you can re-run any failed tests with an existing namespace that already contains ready pods, which leads to quicker turnaround when adding more e2e tests.


timflannagan1:
TODO:
- [ ] Lower the readiness initial probe
- [ ] Investigate skipping all existing tests

PR #1064, opened 2019-12-17
Initial e2e for prom connector

bentito:


hack/e2e_dev.sh
```diff
@@ -0,0 +1,70 @@
...
+
+ROOT_DIR=$(dirname "${BASH_SOURCE[0]}")/..
+source "${ROOT_DIR}/hack/common.sh"
+source "${ROOT_DIR}/hack/lib/tests.sh"
+
+function cleanup() {
+    exit_status=$?
+
+#    echo "Removing namespaces with the 'name=metering-testing-ns' label"
+#    kubectl delete ns -l "name=metering-testing-ns" || true
```
bentito: This is just a copy of `hack/e2e.sh` that avoids deleting the project -- for dev trialing of a new test.

There is a also a conditional on Teardown section set to `false` here: https://github.com/operator-framework/operator-metering/pull/1064/commits/5174f648ed72ef82ffead8eeea8e26e4fe7edd99#diff-e8664074431f2653c7da0fc9019a1935R482

test/e2e/presto_prometheus_connector_up_test.go
```diff
@@ -0,0 +1,50 @@
...
+		"--keystore-path", "/opt/presto/tls/keystore.pem",
+		"--execute",
+		"SELECT * FROM prometheus.default.up where timestamp > (NOW() - INTERVAL '10' second)",
+	)
+	var queryResults bytes.Buffer
+	cmd.Stderr = os.Stderr
+	cmd.Stdout = &queryResults
+	err = cmd.Run()
+	queryResultsStr := string(queryResults.Bytes())
+	require.Containsf(t, queryResultsStr, "endpoint=metrics", "query output didn't contain `endpoint=metrics`")
```
timflannagan1: I think we should just pass the `string(queryResultsStr.Bytes())` directly to `require` (like we do on line 26) w/o declaring a variable.

timflannagan1: Let's add a check for a nil error too, e.g.:
```go
require.Nil(t, err, "expected that querying the prometheus catalog would produce no error)
```

pkg/apis/metering/v1/metering_config.go
```diff
@@ -453,9 +452,6 @@ type PrestoConnectorHiveConfig struct {
 	MetastoreTimeout       string     `json:"metastoreTimeout,omitempty"`
 	TLS                    *TLSConfig `json:"tls,omitempty"`
 }
-type PrestoConnectorPrometheusConfig struct {
-	Enabled *bool `json:"enabled,omitempty"`
-}
```
timflannagan1: Why was this deleted?

test/e2e/presto_prometheus_connector_up_test.go
```diff
@@ -0,0 +1,50 @@
...
+	"testing"
+)
+
+func testPrometheusConnectorWorks(t *testing.T, rf *reportingframework.ReportingFramework) {
+	t.Helper()
+	out, err := exec.Command(
+		"/usr/local/bin/kubectl",
+		"-n", rf.Namespace,
+		"get", "pods",
+		"-l", "app=presto",
```
timflannagan1: Hmm while we know that this will always return the `presto-coordinator` pod as we know the context of the manual install, but there are no guarantees that there are no presto-workers pods.

I think we're more looking for:
```go
"-l", "app=presto,presto=coordinator",
```

This way, we're guaranteed to get back the presto coordinator pod in the current namespace.

test/e2e/presto_prometheus_connector_up_test.go
bentito: I thought about that originally, but you can't get `Containsf` without that also being true.

bentito: Oh maybe, could you get some kind of err but still get results on the query? Perhaps, but it might be a warning or something. If you get the results and they Containsf that string, I think testing the other could unnecessarily fail the test.

pkg/apis/metering/v1/metering_config.go
bentito: accident, putting it back. I got confused that we wanted the default case, but if default then no Prom connector.

test/e2e/presto_prometheus_connector_up_test.go
bentito: ah, okay. done.

test/e2e/presto_prometheus_connector_up_test.go
timflannagan1: I feel like disregarding an error, despite the query returning results, may open us up to bugs in the future. I'm not sure though, as your point about `Containsf` is valid.

pkg/apis/metering/v1/metering_config.go
```diff
@@ -442,9 +442,8 @@ type PrestoWorkerSpec struct {
 	Config                        *PrestoServerConfig          `json:"config,omitempty"`
 }
 type PrestoConnectorConfig struct {
-	Hive              *PrestoConnectorHiveConfig       `json:"hive,omitempty"`
-	Prometheus        *PrestoConnectorPrometheusConfig `json:"prometheus,omitempty"`
```
timflannagan1: I'm still seeing this deleted

bentito: fixed. added back

test/e2e/presto_prometheus_connector_up_test.go
bentito: I think to an extent, looser is better on an e2e test. To my mind e2e tests tell you if the system "generally" works, could still have some bugs, doesn't try to rule them all out. Especially on an e2e with a heavy setup/teardown time like all of these, I think it's better to get them to err of the side of working most of the time.

I don't feel the same about unit or integration tests, fwiw.

test/e2e/presto_prometheus_connector_up_test.go
```diff
@@ -0,0 +1,51 @@
...
+		"--user", "root",
+		"--catalog", "prometheus",
+		"--schema", "default",
+		"--keystore-path", "/opt/presto/tls/keystore.pem",
+		"--execute",
+		"SELECT * FROM prometheus.default.up where timestamp > (NOW() - INTERVAL '10' second)",
+	)
+	var queryResults bytes.Buffer
+	cmd.Stderr = os.Stderr
+	cmd.Stdout = &queryResults
```
timflannagan1: Just saw this in the e2e logs:
```go
    --- FAIL: TestManualMeteringInstall/PrometheusConnectorWorks (403.48s)
        --- FAIL: TestManualMeteringInstall/PrometheusConnectorWorks/testPrometheusConnectorWorks (0.00s)
            require.go:765: 
                	Error Trace:	presto_prometheus_connector_up_test.go:23
                	            				main_test.go:475
                	Error:      	Expected nil, but got: &os.PathError{Op:"fork/exec", Path:"/usr/local/bin/kubectl", Err:0x2}
                	Test:       	TestManualMeteringInstall/PrometheusConnectorWorks/testPrometheusConnectorWorks
                	Messages:   	expected querying for the presto pod name would produce no error
```

I think this implies that we should be logging the `queryResults`, e.g.
```go
logrus.Debugf("Query Results: %s", string(queryResults.Bytes())
```

timflannagan1: Whoops, I see that this is for the first command (querying for the presto-coordinator pod name). Maybe we would want to also redirect stdout/stderr for that command like we do for the presto query command.

bentito: Hopefully we'll see from changes in ed27074

PR #1072, opened 2020-01-07
Support querying unmanaged, existing Presto table metadata. 

timflannagan1:
Currently, `spec.columns` are a required value in the `PrestoTable` CRD which is not always ideal. In the case of unmanaged tables, you would still need to specify a `spec.columns` configuration despite that table schema already existing in Presto. This would update the way that the reporting-operator handles unmanaged tables, namely natively querying the Presto go-client for the `spec.tableName` table_name, which eliminates the need to specify a column definition.

### DIFF
--- a/charts/metering-ansible-operator/templates/crds/prestotable.crd.yaml
+++ b/charts/metering-ansible-operator/templates/crds/prestotable.crd.yaml
@@ -35,7 +35,7 @@ spec:
           type: object
           description: |
             PrestoTableSpec is the desired specification of a PrestoTable custom resource.
-            Required fields: unmanaged, catalog, schema, tableName, and columns.
+            Required fields: unmanaged, catalog, schema, tableName. Note: columns is required when unmanaged is set to false.
             Optional fields: query, view, createTableAs, properties, and comment.
             More info: https://github.com/operator-framework/operator-metering/blob/master/Documentation/prestotables.md
           required:
@@ -43,7 +43,6 @@ spec:
           - catalog
           - schema
           - tableName
-          - columns
           properties:
             unmanaged:
               type: boolean
@@ -123,43 +122,3 @@ spec:
                 Note: this is an optional field.
                 More info: https://prestosql.io/docs/current/overview/concepts.html#query
               minLength: 1
-          oneOf:
-          - required:
-            - createTableAs
-            - query
-            properties:
-              createTableAs:
-                enum:
-                - true
-              unmanaged:
-                enum:
-                - false
-            allOf:
-            - not:
-                required:
-                - view
-          - required:
-            - view
-            - query
-            properties:
-              view:
-                enum:
-                - true
-              unmanaged:
-                enum:
-                - false
-            allOf:
-            - not:
-                required:
-                - createTableAs
-          - properties:
-              unmanaged:
-                enum:
-                - true
-            allOf:
-            - not:
-                required:
-                - view
-            - not:
-                required:
-                - createTableAs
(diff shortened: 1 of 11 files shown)


manifests/deploy/ocp-testing/metering-ansible-operator/prestotable.crd.yaml
```diff
@@ -35,15 +35,14 @@ spec:
           type: object
           description: |
             PrestoTableSpec is the desired specification of a PrestoTable custom resource.
-            Required fields: unmanaged, catalog, schema, tableName, and columns.
+            Required fields: unmanaged, catalog, schema, tableName. Note: columns is required when unmanaged is set to false.
```
bentito: Shouldn't we add a validation to enforce what the note says being the case?

bentito: Oh and then I looked up the page and say we had that, nvm

charts/metering-ansible-operator/templates/crds/prestotable.crd.yaml
```diff
@@ -35,15 +35,14 @@ spec:
           type: object
           description: |
             PrestoTableSpec is the desired specification of a PrestoTable custom resource.
-            Required fields: unmanaged, catalog, schema, tableName, and columns.
+            Required fields: unmanaged, catalog, schema, tableName. Note: columns is required when unmanaged is set to false.
```
bentito: Couldn't what the note says still be enforced?

pkg/presto/util.go
```diff
@@ -77,6 +77,28 @@ func DropView(queryer db.Queryer, catalog, schema, viewName string, ignoreNotExi
...
+// table name to determine that table's column information.
+func QueryMetadata(queryer db.Queryer, catalog, schema, tableName string) ([]Column, error) {
+	rows, err := ExecuteSelect(queryer, fmt.Sprintf("DESCRIBE %s", FullyQualifiedTableName(catalog, schema, tableName)))
+	if err != nil {
+		return nil, fmt.Errorf("failed to query the %s Presto table's metadata: %v", tableName, err)
+	}
+
+	var cols []Column
+
+	for _, row := range rows {
```
bentito: What if code in here errors? We're still returning nil for the error.

pkg/presto/util.go
```diff
@@ -77,6 +77,28 @@ func DropView(queryer db.Queryer, catalog, schema, viewName string, ignoreNotExi
...
 }
 
+// QueryMetadata executes a "DESCRIBE" Presto query against an existing, fully-qualified
+// table name to determine that table's column information.
+func QueryMetadata(queryer db.Queryer, catalog, schema, tableName string) ([]Column, error) {
+	rows, err := ExecuteSelect(queryer, fmt.Sprintf("DESCRIBE %s", FullyQualifiedTableName(catalog, schema, tableName)))
+	if err != nil {
+		return nil, fmt.Errorf("failed to query the %s Presto table's metadata: %v", tableName, err)
+	}
+
```
bentito: nit: the extra vertical spacing doesn't add much value and doesn't match with the surrounding code.

pkg/presto/util.go
timflannagan1: I think that's a fair point, but I can't see a reason where an error would occur. Type assertion in Go asserts that the row["Column"] element is not nil, and the value is of type string.

timflannagan1: Actually I believe you're right as that implies that if either of those checks fails - it panics which is not ideal.

PR #1075, opened 2020-01-14
Bug 1782504: Update the default reporting-operator addresses to work with dual-stack IP versions.

timflannagan1:
With the upcoming need to support dual ipv4/ipv6 stack configurations, this would proactively update the default address parameters to use the ipv6 notation as we punt these addresses to `svc.ListenAndServe()` which can support a dual-stack configuration.


timflannagan1:
/retitle Bug 1782504: Update the default reporting-operator addresses to work with dual-stack IP versions.

timflannagan1:
As mentioned in the BZ tracking this, it doesn't appear that this change is necessary to work on IPV6 environments.

PR #1076, opened 2020-01-20
Improve S3 bucket usage docs

bentito:
I think this helps by splitting the docs for S3 into subheads, one for each path: Create Bucket or Use Existing Bucket. Also clarifies the wording.

### DIFF
--- a/Documentation/configuring-storage.md
+++ b/Documentation/configuring-storage.md
@@ -9,18 +9,16 @@ Additionally, Hive metastore requires storage for it's database containing metad
 > Changing these settings after installation will result in broken and unexpected behavior.
 
 ## Storing data in Amazon S3
-
+### Create a S3 bucket
 To use Amazon S3 for storage, edit the `spec.storage` section in the example [s3-storage.yaml][s3-storage-config] configuration.
 Set the `spec.storage.hive.s3.bucket`, `spec.storage.hive.s3.region` and `spec.storage.hive.s3.secretName` values.
 
-The `bucket` and `region` fields should be the name, and optionally the path within the bucket you wish to store Metering data at, and the region in which you wish to create your bucket in, respectively.
-
-If you want to provide an existing S3 bucket, or do not want to provide IAM credentials that have CreateBucket permissions, set `spec.storage.hive.s3.createBucket` to `false` and provide the name of a pre-existing bucket for `spec.storage.hive.s3.bucket`.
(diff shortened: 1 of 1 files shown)


Documentation/configuring-storage.md
```diff
@@ -9,18 +9,16 @@ Additionally, Hive metastore requires storage for it's database containing metad
...
-
-If you want to provide an existing S3 bucket, or do not want to provide IAM credentials that have CreateBucket permissions, set `spec.storage.hive.s3.createBucket` to `false` and provide the name of a pre-existing bucket for `spec.storage.hive.s3.bucket`.
-The `spec.storage.hive.s3.secretName` should be the name of a secret in the Metering namespace containing the AWS credentials in the `data.aws-access-key-id` and `data.aws-secret-access-key` fields.
+The `bucket` field defines the bucket name, and if desired, the path for storing Metering data.
+`region` is the AWS region in which you wish to create your bucket in.
 
 > **Note**: The values of the `aws-access-key-id` and `aws-secret-access-key` must be base64 encoded.
 
-For example:
+For example, a file to be loaded with `kubectl create -f <aws-secret-file.yaml> --namespace <aws-based-cluster>`:
```
timflannagan1: Hmm, I'm not sure what's more clear:
- Altering the `Secret` definition below to contain a `metadata.namespace` field
- Using the kubectl command directly

We still want to be consistent with how we handle namespacing throughout the rest of the documentation, so I think at the very least we should alter the above with:
`kubectl -n $METERING_NAMESPACE create -f <aws-secret-file.yaml>`

Documentation/configuring-storage.md
```diff
@@ -9,18 +9,16 @@ Additionally, Hive metastore requires storage for it's database containing metad
 > Changing these settings after installation will result in broken and unexpected behavior.
 
 ## Storing data in Amazon S3
-
+### Create a bucket
```
timflannagan1:
```suggestion
### Creating an S3 bucket
```

Documentation/configuring-storage.md
```diff
@@ -30,6 +28,11 @@ data:
   aws-access-key-id: "dGVzdAo="
   aws-secret-access-key: "c2VjcmV0Cg=="
 ```
+`spec.storage.hive.s3.secretName` must then be set to: `metadata.name` from the Secret.
+### Use existing bucket
```
timflannagan1:
```suggestion
### Using an existing S3 bucket
```

Documentation/configuring-storage.md
bentito: Let's keep mention of the namespace with what you suggest as a change. If we leave out namespace, it's pretty opaque to the user when the secret is added to some other namespace (like their default).

PR #1077, opened 2020-01-21
Bug 1793543: Allow support for using S3 buckets in an IPV6 cluster.

timflannagan1:
AWS exposes a REST API endpoint for dual-stack s3 bucket hostnames. After playing around with both the virtual hosted-style and path-style requests, it appears that we still need to specify the fs.defaultFS and hive.metastore.warehouse.dir in order to talk with Hive Metastore correctly. This implies that we need to specify the s3 bucket name as the defaultFS/warehouseDir components, and use the path-style endpoint (i.e. s3.dualstack.us-east-1.amazonaws.com/<bucket name>) to hit the correct s3 bucket host.

Link to AWS dualstack docs: https://docs.aws.amazon.com/AmazonS3/latest/dev/dual-stack-endpoints.html

### DIFF
--- a/charts/openshift-metering/templates/hadoop/_hadoop_config_helpers.tpl
+++ b/charts/openshift-metering/templates/hadoop/_hadoop_config_helpers.tpl
@@ -57,9 +57,20 @@
 {{- define "core-site-xml" -}}
 <configuration>
   <property>
-      <name>fs.defaultFS</name>
-      <value>{{ .Values.hadoop.spec.config.defaultFS }}</value>
+    <name>fs.defaultFS</name>
+    <value>{{ .Values.hadoop.spec.config.defaultFS }}</value>
   </property>
+{{- if and .Values.hadoop.spec.config.aws.region .Values.useIPV6Networking }}
+  <property>
+    <name>fs.s3a.endpoint</name>
+    <value>https://s3.dualstack.{{ .Values.hadoop.spec.config.aws.region }}.amazonaws.com</value>
+  </property>
+  <property>
+    <name>fs.s3a.path.style.access</name>
+    <value>true</value>
+    <description>Enable S3 path style access.</description>
+  </property>
+{{- end }}
 {{- if .Values.hadoop.spec.config.s3Compatible.endpoint }}
   <property>
     <name>fs.s3a.impl</name>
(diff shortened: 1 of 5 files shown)


charts/openshift-metering/templates/hadoop/_hadoop_config_helpers.tpl
```diff
@@ -57,8 +57,17 @@
...
 <configuration>
   <property>
-      <name>fs.defaultFS</name>
-      <value>{{ .Values.hadoop.spec.config.defaultFS }}</value>
+    <name>fs.defaultFS</name>
+    <value>{{ .Values.hadoop.spec.config.defaultFS }}</value>
+  </property>
+  <property>
+    <name>fs.s3a.endpoint</name>
+    <value>https://s3.dualstack.us-east-1.amazonaws.com</value>
```
bentito: This would lock in the region, versus probably it should come in as a region variable: 

from the doc?:
`bucketname.s3.dualstack.aws-region.amazonaws.com`

charts/openshift-metering/templates/presto/_presto_helpers.tpl
```diff
@@ -7,6 +7,10 @@ hive.compression-codec=SNAPPY
 hive.hdfs.authentication.type=NONE
 hive.metastore.authentication.type=NONE
 hive.collect-column-statistics-on-write=true
+# TODO(tflannag): temporary fix, need to add templating for when
+# s3 was specified in the MeteringConfig CR
+hive.s3.endpoint=https://s3.dualstack.us-east-1.amazonaws.com
```
bentito: Same comment about region as variable.

charts/openshift-metering/templates/presto/presto-coordinator-config.yaml
```diff
@@ -30,7 +30,7 @@ data:
     internal-communication.https.truststore.path=/opt/presto/tls/truststore.pem
 {{- else }}
     http-server.http.port=8080
-    discovery.uri=http://presto:8080
+    discovery.uri=http://presto-coordinator-0:8080
```
bentito: Wait I thought from the Slack discussion that moving this host name, while it seems like it fixes the Presto announce problem, won't actually work for the cluster members reaching Presto?

timflannagan1: I threw this PR solely for reference and getting a baseline for Metering installs. Going forward, we need to add some templating around the hive/presto s3 configurations (and avoid hardcoding any bucket regions) and figure out the DNS issues around the Presto discovery URI.

bentito: 👍

timflannagan1:
/retitle Bug 1793543: Allow support for using S3 buckets in an IPV6 cluster.

timflannagan1:
Note: there's still some work needed resolving the Presto hostname.
/hold cancel
/retest

PR #1043, opened 2019-11-26
Add logging of resource installs to k8s_event_module

EmilyM1:

PR #1078, opened 2020-01-21
Revert "Add logging of resource installs to k8s_event_module"

timflannagan1:
Reverts operator-framework/operator-metering#1043

I've seen this module break a couple of Metering installs so let's play it safe and revert this PR until we wrap up the ipv6-related support.


timflannagan1:
I haven't seen any more issues related to this, so reverting this is probably not needed anymore.

PR #1079, opened 2020-01-21
Add support for auto-detecting single-stack IPV4/IPV6 Openshift clusters.

timflannagan1:
Dockerfile(s): Add the python `netaddr` module to allow the usage of the `ipv4` Jinja2 filter.

Charts/: This would add a top-level variable that enables IPV6 networking to values.yaml.

Images/: Update the MeteringConfig Ansible role to allow support for inspecting the Networking config.openshift.io object, iterating over the serviceNetwork array and checking if there are any IPV4 addresses. In the case where there's a mixture of IPV4 and IPV6 addresses, treat that configuration as a single-stack IPV4 Openshift cluster. Else, treat it as a single-stack IPV6 Openshift cluster.

### DIFF
--- a/charts/metering-ansible-operator/values.yaml
+++ b/charts/metering-ansible-operator/values.yaml
@@ -36,6 +36,13 @@ operator:
       verbs:
       - get
       - list
+    # grants access to view the networking config
+    - apiGroups:
+      - config.openshift.io
+      resources:
+      - networks
+      verbs:
+      - list
     # used by openshift auth-proxy
     - apiGroups:
       - authorization.k8s.io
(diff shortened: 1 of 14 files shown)


timflannagan1:
I realize I only tested this in isolation and that the metering-operator service account does not permissions to view cluster-wide resources.
/hold

timflannagan1:
Updated the charts/manifests to allow the metering-operator service account the requisite permissions to view the cluster's networking configuration.
/hold cancel

charts/metering-ansible-operator/values.yaml
```diff
@@ -36,6 +36,13 @@ operator:
       verbs:
       - get
       - list
+    # grants access to view the networking config
```
bentito: oof, so many places to add the exact same lines.

timflannagan1: I know, we manage too many sets of manifest files.

PR #1080, opened 2020-01-22
images: Remove the task name for ending the role during an error.

timflannagan1:
When viewing the metering-operator "operator" container logs, there's no field that details whether a task is skipped or not [1], and in most cases we skip the fail module for this task. This can lead to confusion, so removing this task name can help improve clarity.

[1] {"level":"info","ts":1579721726.2411647,"logger":"logging_event_handler","msg":"[playbook task]","name":"operator-metering","namespace":"openshift-metering","gvk":"metering.openshift.io/v1, Kind=MeteringConfig","event_type":"playbook_on_task_start","job":"3510942875414458836","EventData.Name":"meteringconfig : End role due to failure"}


timflannagan1:
The E2E keeps flaking due to AMI/terraform errors and this doesn't change any existing functionality, so force-merging.

PR #1081, opened 2020-01-27
Bug 1795241: Add hostAliases with presto -> ipv6 loopback

bentito:


charts/openshift-metering/templates/presto/presto-coordinator-statefulset.yaml
```diff
@@ -56,6 +56,11 @@ spec:
 {{ toYaml .Values.presto.spec.annotations | indent 8 }}
 {{- end }}
     spec:
+      hostAliases:
```
timflannagan1: Let's add some templating around the `useIPV6Networking` key:
```yaml
if {{- .Values.useIPV6Networking }}
...
{{- end }}
```
 
While I don't think adding a host alias for the ipv6 localhost will affect any ipv4-based installs, I still think we should separate single-stack ipv4 and ipv6 until we have a better idea of how to tackle dual-stack.

charts/openshift-metering/values.yaml
```diff
@@ -1,11 +1,14 @@
 global:
   ownerReferences: []
+  useIPV6Networking: false
```
timflannagan1: remove this

charts/openshift-metering/templates/presto/presto-coordinator-statefulset.yaml
```diff
@@ -51,11 +51,15 @@ spec:
 {{- end }}
 {{- if and .Values.presto.spec.config.connectors.hive.tls.enabled .Values.presto.spec.config.connectors.hive.tls.createSecret }}
         presto-hive-metastore-tls-hash: {{ include (print $.Template.BasePath "/presto/presto-hive-metastore-tls-secrets.yaml") . | sha256sum }}
-{{- end }}
-{{- if .Values.presto.spec.annotations }}
-{{ toYaml .Values.presto.spec.annotations | indent 8 }}
```
timflannagan1: This needs to be undone too

charts/openshift-metering/values.yaml
```diff
@@ -1,11 +1,14 @@
...
 
 unsupportedFeatures:
   enableHDFS: false
 
 disableOCPFeatures: false
 
+useIPV6Networking: false
```
timflannagan1: Hmm, this is being rendered weird for me as it looks like you're adding this to the values.yaml file but that's already present upstream: https://github.com/operator-framework/operator-metering/blob/master/charts/openshift-metering/values.yaml#L9

PR #1082, opened 2020-01-27
Bug 1795267: Allow the metering-operator to get the network openshift config object.

timflannagan1:
This would add the 'get' verb to the cluster-role RBAC rules as the 'list' verb was insufficient permissions.


timflannagan1:
This is breaking existing installs, and CI keeps flaking so I'm going to force-merge this.

PR #1086, opened 2020-01-29
Makefile: Remove the test2json dependency and unused PHONY targets.

timflannagan1:


EmilyM1:
Will be interested to see if test data for sample reports populates and deserializes without package.

timflannagan1:
After some light investigation, it appears that pre-1.10 Go releases didn't support printing test results as json, and I think we can safely get rid of this dependency now.

/hold cancel

PR #1088, opened 2020-01-29
Bug 1787745: Remove the RSA's md5 algorithm from the Ansible openssl modules.

timflannagan1:
When attempting to run Metering on a FIPS-enabled cluster, we fail role execution while generating user TLS certificates calling the 'openssl_privatekey' module. This module utilizes the module_utils/crypto.py helper library. This library iterates over a list of hashlib algorithms, which contains "FIPS-compliant" algorithms but still contains the 'md5' algorithm.

This would remove that algorithm from the list of available algorithms, and allow role execution to progress past the TLS-related tasks.

### DIFF
--- a/images/metering-ansible-operator/roles/meteringconfig/module_utils/crypto.py
+++ b/images/metering-ansible-operator/roles/meteringconfig/module_utils/crypto.py
@@ -1,0 +1,2009 @@
+# -*- coding: utf-8 -*-
+#
+# (c) 2016, Yanis Guenane <yanis+ansible@guenane.org>
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+#
+# ----------------------------------------------------------------------
+# A clearly marked portion of this file is licensed under the BSD license
+# Copyright (c) 2015, 2016 Paul Kehrer (@reaperhulk)
+# Copyright (c) 2017 Fraser Tweedale (@frasertweedale)
+# For more details, search for the function _obj2txt().
+# ---------------------------------------------------------------------
+# A clearly marked portion of this file is extracted from a project that
+# is licensed under the Apache License 2.0
+# Copyright (c) the OpenSSL contributors
+# For more details, search for the function _OID_MAP.
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+
+import sys
+from distutils.version import LooseVersion
+
+try:
+    import OpenSSL
+    from OpenSSL import crypto
+except ImportError:
+    # An error will be raised in the calling class to let the end
+    # user know that OpenSSL couldn't be found.
+    pass
+
+try:
+    import cryptography
+    from cryptography import x509
+    from cryptography.hazmat.backends import default_backend as cryptography_backend
+    from cryptography.hazmat.primitives.serialization import load_pem_private_key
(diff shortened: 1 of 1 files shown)


timflannagan1:
CC @pruan-rht

images/metering-ansible-operator/roles/meteringconfig/module_utils/crypto.py
```diff
@@ -0,0 +1,2015 @@
...
+def get_fingerprint_of_bytes(source):
+    """Generate the fingerprint of the given bytes."""
+
+    fingerprint = {}
+
+    try:
+        algorithms = hashlib.algorithms
+    except AttributeError:
+        try:
+            algorithms = hashlib.algorithms_guaranteed
```
pruan-rht: @timflannagan1 the #guraanteed method returns a set, so I think you can just do algorithms.remove('md5') w/o transforming it into a `list`

timflannagan1: Weirdly enough, when I import the hashlib in the python interpreter locally, I see the same thing:

```bash
tflannag@localhost operator-metering [docker-src-updates] ▶ python
Python 3.7.5 (default, Oct 17 2019, 12:16:48) 
[GCC 9.2.1 20190827 (Red Hat 9.2.1-1)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> 
>>> import hashlib
>>> 
>>> type(hashlib.algorithms_guaranteed)
<class 'set'>
>>> 
```

But when I was running this through the metering-operator, that method was being interpreted as a tuple, and not a set/list.

timflannagan1: Resolved through slack - `hashlib.algorithms` returns a tuple, and `hashlib.algorithms_guaranteed` returns a set. We were assuming that `hashlib.algorithms_guaranteed` was always going to be returned as `hashlib.algorithms` was depreciated in python3. 

Metering uses a python2 build, so there's a fair chance that `hashlib.algorithms` will be available, and in the case it's not, `hashlib.algorithms_guaranteed` will be chosen. In either case, both get translated to a list so we should have consistent behavior.

images/metering-ansible-operator/roles/meteringconfig/module_utils/crypto.py
```diff
@@ -0,0 +1,2015 @@
...
+    digest = None
+    if digest_string == 'sha256':
+        digest = hashes.SHA256()
+    elif digest_string == 'sha384':
+        digest = hashes.SHA384()
+    elif digest_string == 'sha512':
+        digest = hashes.SHA512()
+    elif digest_string == 'sha1':
+        digest = hashes.SHA1()
+    elif digest_string == 'md5':
```
EmilyM1: Does this need to be removed?

images/metering-ansible-operator/roles/meteringconfig/module_utils/crypto.py
```diff
@@ -0,0 +1,2015 @@
...
+    '1.2.840.113533.7.66.12': ('pbeWithMD5AndCast5CBC', ),
+    '1.2.840.113533.7.66.13': ('password based MAC', 'id-PasswordBasedMAC'),
+    '1.2.840.113533.7.66.30': ('Diffie-Hellman based MAC', 'id-DHBasedMac'),
+    '1.2.840.113549': ('RSA Data Security, Inc.', 'rsadsi'),
+    '1.2.840.113549.1': ('RSA Data Security, Inc. PKCS', 'pkcs'),
+    '1.2.840.113549.1.1': ('pkcs1', ),
+    '1.2.840.113549.1.1.1': ('rsaEncryption', ),
+    '1.2.840.113549.1.1.2': ('md2WithRSAEncryption', 'RSA-MD2'),
+    '1.2.840.113549.1.1.3': ('md4WithRSAEncryption', 'RSA-MD4'),
+    '1.2.840.113549.1.1.4': ('md5WithRSAEncryption', 'RSA-MD5'),
```
EmilyM1: Unneeded?

images/metering-ansible-operator/roles/meteringconfig/module_utils/crypto.py
timflannagan1: Nope, the openssl Ansible modules don't interact with this function.

PR #1092, opened 2020-02-03
Bug 1797801: Update the Hive and reporting-operator charts to work with single-stack ipv4 and ipv6.

timflannagan1:
This is a follow-up to PR #1077 which introduced support for using S3 as the storage backend in an IPV6 cluster. The changes in this PR update the Hive ghostunnel configuration needed to support both a single-stack configuration, dependent on the value of the top-level .useIPV6Networking.

In the case of the pkg/reporting-operator/cmd.go flags, we typically override the default values through the corresponding environment variable, which is reflected in the reporting-operator deployment changes. It's important to note that this means we need to be more careful when interacting with the CLI explicitly (i.e. in the case of running the reporting-operator locally, or in the e2e suite.)

### DIFF
--- a/charts/openshift-metering/templates/hive/hive-configmap.yaml
+++ b/charts/openshift-metering/templates/hive/hive-configmap.yaml
@@ -31,23 +31,33 @@ data:
       </property>
       <property>
         <name>hive.server2.thrift.bind.host</name>
-{{- if .Values.hive.spec.server.config.tls.enabled }}
+{{- if and .Values.hive.spec.server.config.tls.enabled .Values.useIPV6Networking }}
+        <value>[::1]</value>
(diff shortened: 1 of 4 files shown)


timflannagan1:
/retitle Bug 1797801: Update the Hive and reporting-operator charts to work with single-stack ipv4 and ipv6.

timflannagan1:
This needs some updating
/hold

timflannagan1:
Not entirely sure what's happening here. Both of the metering-ansible-operator Dockerfiles build locally fine, and the previous PR that added the `botocore` package also built fine. I'm going to nuke the namespace in CI and see if it resolves itself.
/retest

PR #1094, opened 2020-02-05
Bug 1790601: Fix description of runImmediately

bentito:
Changes docs and logging around reports `spec.runImmediately: true` to reflect what actually is required and what happens.

### DIFF
--- a/Documentation/reports.md
+++ b/Documentation/reports.md
@@ -242,9 +242,9 @@ spec:
 ```
 
 ### runImmediately
+When `runImmediately` is set to `true`, the report will be run immediately. This behavior ensures that the report is immediately processed and queued without requiring additional scheduling parameters.
 
-Set `runImmediately` to `true` to run the report immediately with all available data, regardless of the `reportingStart` or `reportingEnd` values, and without checking if there is any data for the report period.
-For reports with a schedule set, it will not wait for each period's reportingEnd to elapse before processing and all reportPeriods between `reportingStart` and `reportingEnd`.
+> *Note*: When `runImmediately` is set to `true` you must set a `reportingEnd` and `reportingStart` value.
 
 ### Inputs
 
--- a/pkg/operator/reports.go
+++ b/pkg/operator/reports.go
@@ -486,7 +486,7 @@ func (op *Reporting) runReport(logger log.FieldLogger, report *metering.Report) 
 	var runningMsg, runningReason string
 	if report.Spec.RunImmediately {
 		runningReason = meteringUtil.RunImmediatelyReason
-		runningMsg = fmt.Sprintf("Report %s scheduled: runImmediately=true bypassing reporting period [%s to %s].", report.Name, reportPeriod.periodStart, reportPeriod.periodEnd)
+		runningMsg = fmt.Sprintf("Report %s scheduled: runImmediately=true and reporting period [%s to %s].", report.Name, reportPeriod.periodStart, reportPeriod.periodEnd)
 	} else {
 		// Check if it's time to generate the report
 		if reportPeriod.periodEnd.After(now) {


pkg/operator/reports.go
```diff
@@ -486,7 +486,7 @@ func (op *Reporting) runReport(logger log.FieldLogger, report *metering.Report)
 	var runningMsg, runningReason string
 	if report.Spec.RunImmediately {
 		runningReason = meteringUtil.RunImmediatelyReason
-		runningMsg = fmt.Sprintf("Report %s scheduled: runImmediately=true bypassing reporting period [%s to %s].", report.Name, reportPeriod.periodStart, reportPeriod.periodEnd)
+		runningMsg = fmt.Sprintf("Report %s scheduled: runImmediately=true and reporting period [%s to %s].", report.Name, reportPeriod.periodStart, reportPeriod.periodEnd)
```
timflannagan1: Now that I've had some time to think about it, "bypassing" might be a decent description. We may just want to take it a step further and log something like "bypassing validation checks from reporting period ...". WDYT?

Documentation/reports.md
```diff
@@ -242,9 +242,8 @@ spec:
 ```
 
 ### runImmediately
-
-Set `runImmediately` to `true` to run the report immediately with all available data, regardless of the `reportingStart` or `reportingEnd` values, and without checking if there is any data for the report period.
-For reports with a schedule set, it will not wait for each period's reportingEnd to elapse before processing and all reportPeriods between `reportingStart` and `reportingEnd`.
+Set `runImmediately` to `true` to run the report immediately, this is a simple way to have a report run right away without setting schedule values that would cause it to do so.
```
timflannagan1: When `runImmediately` is set to `true`, the report will be run immediately. This is a simple way to have a report get processed and queued right away without setting schedule values that could cause it to get requeued.

**Note**: You must set a `reportingEnd` value when `runImmediately` is set to `true`. If you do not specify a `reportingStart` value, the current time will be used.

Set `runImmediately` to `true` to run the report immediately with all available data, regardless of the `reportingStart` or `reportingEnd` values, and without checking if there is any data for the report period.

For reports with a schedule specified, it will not wait for each period's `reportingEnd` to elapse before processing and all the reporting periods between `reportingStart` and `reportingEnd`.

timflannagan1: I was thinking something along the lines of that, but I'm not sure how to incorporate a description that handles both a run-once and a scheduled report so it's clear enough to the user what the functionality of `runImmediately` does.

pkg/operator/reports.go
bentito: But setting it true doesn't ignore, or bypass, the reportingEnd marker though. It bypasses `spec.schedule` values and I think the logging as changed reflects that and shows what reporting period is in use. The phrasing, as it was, continues the confusion that the reportingEnd might not be needed, IMHO.

timflannagan1: Hmm okay, that's a fair point. Maybe the best solution at this point is to hold off on making any more updates to this logging message.

Documentation/reports.md
```diff
@@ -242,8 +242,9 @@ spec:
 ```
 
 ### runImmediately
-Set `runImmediately` to `true` to run the report immediately, this is a simple way to have a report run right away without setting schedule values that would cause it to do so.
-You must set a `reportingEnd` value. If you do not set a `reportingStart` value, the current time will be used.
+When `runImmediately` is set to `true`, the report will be run immediately. This is a simple way to have a report get processed and queued right away without setting schedule values that could cause it to get requeued.
+
+> *Note*: You must set a `reportingEnd` value when `runImmediately` is set to `true`. If you do not specify a `reportingStart value`, the current time will be used.
```
timflannagan1: s/`reportingStart value`/`reportingStart` value.

bentito:
/retitle BZ 1790601 : Fix description of runImmediately

bentito:
/retitle Bug 1790601 : Fix description of runImmediately

Documentation/reports.md
```diff
@@ -242,9 +242,9 @@ spec:
 ```
 
 ### runImmediately
+When `runImmediately` is set to `true`, the report will be run immediately. This is a simple way to have a report get processed and queued right away without setting schedule values that could cause it to get requeued.
 
-Set `runImmediately` to `true` to run the report immediately with all available data, regardless of the `reportingStart` or `reportingEnd` values, and without checking if there is any data for the report period.
-For reports with a schedule set, it will not wait for each period's reportingEnd to elapse before processing and all reportPeriods between `reportingStart` and `reportingEnd`.
+> *Note*: You must set a `reportingEnd` value when `runImmediately` is set to `true`. If you do not specify a `reportingStart' value, the current time will be used.
```
pruan-rht: @bentito it's good to have the note, but FYI, if the user leaves that value of, the validation will flag it 

> pruan@MacBook-Pro ~/workspace/v3-testfiles/metering (master●)$ oc create -f run_now_example.yaml
> The Report "namespace-memory-usage-now" is invalid:
> * : Invalid value: "": "spec" must validate at least one schema (anyOf)
> * spec.reportingEnd: Required value

timflannagan1:
```suggestion
> *Note*: You must set a `reportingEnd` value when `runImmediately` is set to `true`. If you do not specify a `reportingStart` value, the current time will be used.
```

lamek:
@timflannagan1 lgtm. I can update downstream.

lamek:
@bentito This has been updated downstream and merged to the 4.2, 4.3, and 4.4 branches.
https://github.com/openshift/openshift-docs/pull/19642

PR #1098, opened 2020-02-07
Doc fix

EmilyM1:


Documentation/dev/developer-guide.md
```diff
@@ -31,7 +31,7 @@ $ pip3 install operator-courier
 
 It is important that you clone the project to the right location in your $GOPATH.
 
-The repository much be located at:
+The repository path should look like this:
```
pruan-rht: I think that was just a typo... how about just subbing `much` with `must`

timflannagan1: +1

EmilyM1: Oh is that what it was. Sure.

PR #1099, opened 2020-02-10
Bug 1795267: Support updating existing clusterrole/clusterrolebindings when install Metering.

timflannagan1:
When changes are made to the clusterrole and/or clusterrolebinding charts, existing Metering installs can break if existing clusterroles/clusterrolebindings don't have sufficient permissions to access certain k8s resources that the newer charts do.

### DIFF
--- a/pkg/deploy/install.go
+++ b/pkg/deploy/install.go
@@ -246,7 +246,11 @@ func (deploy *Deployer) installMeteringClusterRoleBinding() error {
 		}
 		deploy.logger.Infof("Created the metering cluster role binding")
 	} else if err == nil {
-		deploy.logger.Infof("The metering cluster role binding already exists")
+		_, err = deploy.client.RbacV1().ClusterRoleBindings().Update(res)
+		if err != nil {
+			return fmt.Errorf("failed to update the metering clusterrolebinding: %v", err)
+		}
+		deploy.logger.Infof("The metering cluster role binding has been updated")
 	} else {
 		return err
 	}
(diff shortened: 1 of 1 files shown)


pkg/deploy/install.go
```diff
@@ -246,7 +246,11 @@ func (deploy *Deployer) installMeteringClusterRoleBinding() error {
 		}
 		deploy.logger.Infof("Created the metering cluster role binding")
 	} else if err == nil {
-		deploy.logger.Infof("The metering cluster role binding already exists")
+		_, err = deploy.client.RbacV1().ClusterRoleBindings().Update(res)
```
bentito: Can we take the opportunity to refactor a bad variable naming? `res` -> `clusterRoleBinding`, for instance, seems like `clusterRole` is used in the following function fix up and it makes it more readable.

timflannagan1: I'd prefer to do that in a different PR as that variable naming is used throughout the package:
```bash
tflannag@localhost operator-metering [update-clusterroles] ▶ grep -rnw pkg/deploy/ -e "res" | wc -l
52
tflannag@localhost operator-metering [update-clusterroles] ▶ 
```

pkg/deploy/install.go
```diff
@@ -259,15 +263,21 @@ func (deploy *Deployer) installMeteringClusterRole() error {
...
+	clusterRole, err := deploy.client.RbacV1().ClusterRoles().Get(res.Name, metav1.GetOptions{})
 	if apierrors.IsNotFound(err) {
 		_, err := deploy.client.RbacV1().ClusterRoles().Create(res)
 		if err != nil {
 			return err
 		}
 		deploy.logger.Infof("Created the metering cluster role")
 	} else if err == nil {
-		deploy.logger.Infof("The metering cluster role already exists")
+		clusterRole.Rules = res.Rules
```
timflannagan1: We could do something like a deep equality check between the current `ClusterRole` object and the current Metering ClusterRole manifest (that gets encoded as a Go object), but I don't think that's worth the cost. This would instead always swap what's present in the YAML manifest with the existing Metering `ClusterRole` k8s object.

timflannagan1: We may also just want to handle this as we do with the cluster role binding update and always pass the `res` variable (the YAML manifest -> Go object) and call the `.Update()` method.

pkg/deploy/install.go
bentito: ah okay

PR #1101, opened 2020-02-13
Bug 1802638: Allow support for creating an S3 bucket on an IPV6 cluster.

timflannagan1:
This is needed as the s3_bucket Ansible module does not support dual-stack endpoints, which is problematic is metering is run in an ipv6 single-stack configuration. In order to use the aws_s3 module which exposes a `dualstack` option, but has a dependency on the `botocore` python package. This commit would install that package so we can utilize this module to support both single-stack ipv4 and single-stack ipv6 configurations.

### DIFF
--- a/Dockerfile.metering-ansible-operator
+++ b/Dockerfile.metering-ansible-operator
@@ -24,7 +24,11 @@ RUN ln -f -s /usr/bin/oc /usr/bin/kubectl
 # put tini into our path
 RUN ln -f -s /tini /usr/bin/tini
 
-RUN pip install --no-cache-dir --upgrade openshift boto3 cryptography netaddr
+# List of python packages:
+# 1. botocore and boto3 are used by the aws-related modules (aws_s3)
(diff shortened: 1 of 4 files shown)


Dockerfile.metering-ansible-operator
```diff
@@ -24,7 +24,11 @@ RUN ln -f -s /usr/bin/oc /usr/bin/kubectl
 # put tini into our path
 RUN ln -f -s /tini /usr/bin/tini
 
-RUN pip install --no-cache-dir --upgrade openshift boto3 cryptography netaddr
+# List of python packages:
+# 1. botocore and boto3 are used by the aws-related modules (aws_s3)
+# 2. netaddr is needed to use the ipv4/ipv6 jinja filter
+# 3. cryptography is used by the openssl_* modules for TLS/authentication purposes
```
pruan-rht: @timflannagan1, LGTM...I think we need to test it with FIPS on

PR #1104, opened 2020-02-18
charts,manifests,pkg: Support create unmanaged PrestoTables from a ReportDataSource.

timflannagan1:
This would add the ability to create a ReportDataSource that references an existing table in Presto. Before, you would need to first create an unmanaged PrestoTable that referenced an existing table in Presto, then create a ReportDataSource using the spec.prestoTable sub-type. With initial support for the presto-prometheus connector being integrated, this change would be needed as Presto would read directly from Prometheus and store relevant data into tables which the ReportDataSources can then reference.


manifests/deploy/ocp-testing/olm/bundle/4.4/reportdatasource.crd.yaml
```diff
@@ -133,6 +133,16 @@ spec:
...
+            linkExistingTable:
+              type: object
+              required:
+              - tableName
+              properties:
+                tableName:
+                  description: |
+                    TableName is the fully-qualified table name (i.e. catalog.schema.table_name) of an existing Presto table.
+                  type: string
+                  minLenght: 1
```
bentito: typo: minLength

manifests/deploy/openshift/metering-ansible-operator/reportdatasource.crd.yaml
```diff
@@ -133,6 +133,16 @@ spec:
...
+            linkExistingTable:
+              type: object
+              required:
+              - tableName
+              properties:
+                tableName:
+                  description: |
+                    TableName is the fully-qualified table name (i.e. catalog.schema.table_name) of an existing Presto table.
+                  type: string
+                  minLenght: 1
```
bentito: typo

charts/metering-ansible-operator/templates/crds/reportdatasource.crd.yaml
```diff
@@ -133,6 +133,16 @@ spec:
...
+            linkExistingTable:
+              type: object
+              required:
+              - tableName
+              properties:
+                tableName:
+                  description: |
+                    TableName is the fully-qualified table name (i.e. catalog.schema.table_name) of an existing Presto table.
+                  type: string
+                  minLenght: 1
```
bentito: typo: minLength

manifests/deploy/ocp-testing/metering-ansible-operator/reportdatasource.crd.yaml
```diff
@@ -133,6 +133,16 @@ spec:
...
+            linkExistingTable:
+              type: object
+              required:
+              - tableName
+              properties:
+                tableName:
+                  description: |
+                    TableName is the fully-qualified table name (i.e. catalog.schema.table_name) of an existing Presto table.
+                  type: string
+                  minLenght: 1
```
bentito: typo

manifests/deploy/openshift/telemeter/list.yaml
```diff
@@ -3520,6 +3522,16 @@ objects:
                 required:
                 - source
                 type: object
+              linkExistingTable:
+                properties:
+                  tableName:
+                    description: |
+                      TableName is the fully-qualified table name (i.e. catalog.schema.table_name) of an existing Presto table.
+                    minLenght: 1
```
bentito: typo

manifests/deploy/upstream/metering-ansible-operator/reportdatasource.crd.yaml
```diff
@@ -133,6 +133,16 @@ spec:
...
+            linkExistingTable:
+              type: object
+              required:
+              - tableName
+              properties:
+                tableName:
+                  description: |
+                    TableName is the fully-qualified table name (i.e. catalog.schema.table_name) of an existing Presto table.
+                  type: string
+                  minLenght: 1
```
bentito: typo

manifests/deploy/upstream/olm/bundle/4.4/reportdatasource.crd.yaml
```diff
@@ -133,6 +133,16 @@ spec:
...
+            linkExistingTable:
+              type: object
+              required:
+              - tableName
+              properties:
+                tableName:
+                  description: |
+                    TableName is the fully-qualified table name (i.e. catalog.schema.table_name) of an existing Presto table.
+                  type: string
+                  minLenght: 1
```
bentito: typo

manifests/deploy/ocp-testing/olm/bundle/4.4/reportdatasource.crd.yaml
timflannagan1: Good catch - fixed.

pkg/operator/datasources.go
```diff
@@ -440,6 +443,62 @@ func (op *Reporting) handlePrestoTableDataSource(logger log.FieldLogger, dataSou
 	return nil
 }
 
+// TODO: investigate alternative names. exposeUnmanagedTable?
+func (op *Reporting) handleLinkExistingTable(logger log.FieldLogger, dataSource *metering.ReportDataSource) error {
+	if dataSource.Spec.LinkExistingTable.TableName == "" {
+		return fmt.Errorf("invalid configuration passed: spec.linkExistingTable.tableName field cannot be empty")
+	}
+	inputs := strings.Split(dataSource.Spec.LinkExistingTable.TableName, ".")
+	if len(inputs) != 3 {
```
bentito: What is this magic number 3?

pkg/operator/datasources.go
```diff
@@ -440,6 +443,62 @@ func (op *Reporting) handlePrestoTableDataSource(logger log.FieldLogger, dataSou
...
+		}
+		logger.Infof("existing LinkExistingTable ReportDataSource discovered, tableName: %s", tableName)
+		return nil
+	}
+
+	catalog := inputs[0]
+	schema := inputs[1]
+	tableName := inputs[2]
+
+	// using the FQTN from the resource, verify we can query the existing table's properties
```
bentito: I don't think this is a common abbreviation, probably just write it out

pkg/operator/datasources.go
```diff
@@ -440,6 +443,62 @@ func (op *Reporting) handlePrestoTableDataSource(logger log.FieldLogger, dataSou
...
+	// for it's metadata. We can then use that information and create a PrestoTable resource
```
bentito: typo: its

pkg/operator/datasources.go
timflannagan1: We're hoping that the `spec.LinkExistingTable.TableName` in the form of a fully-qualified tables name: `catalog.schema.table_name`, and when that string is split by the `.` delimiter and should have three elements. That said, you could argue this needs a constant variable but I'm not sure how much extra value that offers.

bentito:
```suggestion
	if len(inputs) != ENOUGH_DOTS_FOR_FQTN {
```
Idk? Something like that, magic numbers slow me down reading the code, because they make me stop and make sure I understand what's going on.

timflannagan1: Updated to use a constant:
```go
	// given a table_name in the form of "catalog.schema.table_name", we
	// expect that splitting this overall string by the `.` delimiter
	// will yield an array of three string elements
	expectedArrSplitElementsFQTN = 3
```

pkg/operator/datasources.go
```diff
@@ -440,6 +447,67 @@ func (op *Reporting) handlePrestoTableDataSource(logger log.FieldLogger, dataSou
...
+	schema := inputs[1]
+	tableName := inputs[2]
+
+	// using the fully-qualified table name from the resource, verify we can query the existing table's
+	// properties for its metadata. We can then use that information and create a PrestoTable resource.
+	cols, err := op.prestoTableManager.QueryMetadata(catalog, schema, tableName)
+	if err != nil {
+		return fmt.Errorf("failed to query the %s.%s.%s Presto table metadata: %v", catalog, schema, tableName, err)
+	}
+	prestoTable, err := op.createPrestoTableCR(dataSource, metering.ReportDataSourceGVK, catalog, schema, tableName, cols, true, false, "")
```
bentito: The first `true` is actually fairly important to note, it's setting `unmanaged` bool. Perhaps worth a comment or constant. Had to look up the func definition to know that.

timflannagan1: Done - added both a comment and some variables:
```go
	var createView bool
	catalog := inputs[0]
	schema := inputs[1]
	tableName := inputs[2]
	unmanagedTable := true

	// using the fully-qualified table name from the resource, verify we can query the existing table's
	// properties for its metadata. We can then use that information and create a PrestoTable resource.
	cols, err := op.prestoTableManager.QueryMetadata(catalog, schema, tableName)
	if err != nil {
		return fmt.Errorf("failed to query the %s.%s.%s Presto table metadata: %v", catalog, schema, tableName, err)
	}
	// attempt to create an unmanaged PrestoTable CR as we're linking an existing table in Presto to
	// this particular ReportDataSource and don't need the reporting-operator to create this table for us.
	prestoTable, err := op.createPrestoTableCR(dataSource, metering.ReportDataSourceGVK, catalog, schema, tableName, cols, unmanagedTable, createView, "")
	if err != nil {
		return fmt.Errorf("failed to create the PrestoTable for the %s ReportDataSource: %v", dataSource.Name, err)
	}
```

PR #1105, opened 2020-02-18
Bug 1799054: Add quotations around the cluster-monitoring CSV annotation.

timflannagan1:


charts/metering-ansible-operator/values.yaml
```diff
@@ -314,8 +314,8 @@ olm:
         version: v1
 
     annotations:
-      operatorframework.io/suggested-namespace: openshift-metering
-      operatorframework.io/cluster-monitoring: true
+      "operatorframework.io/suggested-namespace": openshift-metering
+      "operatorframework.io/cluster-monitoring": "true"
```
pruan-rht: @timflannagan1 I see you quoted the keys here, but not the other places.

timflannagan1: Fixed.

PR #1106, opened 2020-02-19
proxy enablement

EmilyM1:
This handles changes to reporting-operator and Presto

### DIFF
--- a/charts/openshift-metering/templates/hadoop/_hadoop_config_helpers.tpl
+++ b/charts/openshift-metering/templates/hadoop/_hadoop_config_helpers.tpl
@@ -108,5 +108,40 @@
       <name>fs.gs.reported.permissions</name>
       <value>733</value>
   </property>
+{{- if and .Values.networking.useGlobalProxyNetworking .Values.presto.spec.config.tls.enabled }}
+  <property>
+      <name>fs.s3a.proxy.host</name>
+      <value>{{ .Values.networking.proxy.config.https_proxy.hostname }}</value>
+  </property>
+  <property>
+      <name>fs.s3a.proxy.port</name>
+      <value>{{ .Values.networking.proxy.config.https_proxy.port }}</value>
+  </property>
+  <property>
+      <name>fs.s3a.proxy.username</name>
+      <value>{{ .Values.networking.proxy.config.https_proxy.username }}</value>
+  </property>
+  <property>
+      <name>fs.s3a.proxy.password</name>
+      <value>{{ .Values.networking.proxy.config.https_proxy.password }}</value>
+  </property>
+{{- else if .Values.networking.useGlobalProxyNetworking }}
+  <property>
+      <name>fs.s3a.proxy.host</name>
+      <value>{{ .Values.networking.proxy.config.http_proxy.hostname }}</value>
+  </property>
+  <property>
+      <name>fs.s3a.proxy.port</name>
+      <value>{{ .Values.networking.proxy.config.http_proxy.port }}</value>
+  </property>
+  <property>
+      <name>fs.s3a.proxy.username</name>
+      <value>{{ .Values.networking.proxy.config.http_proxy.username }}</value>
+  </property>
+  <property>
+      <name>fs.s3a.proxy.password</name>
+      <value>{{ .Values.networking.proxy.config.http_proxy.password }}</value>
+  </property>
+{{- end }}
 </configuration>
 {{- end }}
--- a/charts/openshift-metering/templates/presto/presto-coordinator-config.yaml
+++ b/charts/openshift-metering/templates/presto/presto-coordinator-config.yaml
@@ -59,6 +59,17 @@ data:
 {{- end }}
 
   jvm.config: |
+{{- if and .Values.networking.useGlobalProxyNetworking .Values.presto.spec.config.tls.enabled }}
+    -Dhttps.proxyHost={{ .Values.networking.proxy.config.https_proxy.hostname }}
+    -Dhttps.proxyPort={{ .Values.networking.proxy.config.https_proxy.port }}
+    -Dhttps.proxyUser={{ .Values.networking.proxy.config.https_proxy.username }}
+    -Dhttps.proxyPassword={{ .Values.networking.proxy.config.https_proxy.password }}
+{{- else if .Values.networking.useGlobalProxyNetworking }}
+    -Dhttp.proxyHost={{ .Values.networking.proxy.config.http_proxy.hostname }}
+    -Dhttp.proxyPort={{ .Values.networking.proxy.config.http_proxy.port }}
+    -Dhttp.proxyUser={{ .Values.networking.proxy.config.http_proxy.username }}
+    -Dhttp.proxyPassword={{ .Values.networking.proxy.config.http_proxy.password }}
+{{- end }}
     -server
     -XX:+UseContainerSupport
     -XX:+UseG1GC
--- a/charts/openshift-metering/templates/reporting-operator/reporting-operator-deployment.yaml
+++ b/charts/openshift-metering/templates/reporting-operator/reporting-operator-deployment.yaml
@@ -367,6 +367,11 @@ spec:
           name: prometheus-bearer-token
           subPath: token
 {{- end }}
+{{- if and .Values.networking.useGlobalProxyNetworking .Values.networking.proxy.config.trusted_ca_bundle }}
+        - mountPath: /etc/pki/ca-trust/extracted/pem
+          name: reporting-operator-trusted-ca-bundle
+          readOnly: true
+{{- end }}
 {{- if $operatorValues.spec.authProxy.enabled }}
       - name: reporting-operator-auth-proxy
         image: "{{ include "reporting-operator-auth-proxy-image" . }}"
@@ -441,6 +446,14 @@ spec:
 {{- end }}
 {{- end }}{{/* end of authProxy.enabled */}}
       volumes:
+{{- if and .Values.networking.useGlobalProxyNetworking .Values.networking.proxy.config.trusted_ca_bundle }}
+     - name: reporting-operator-trusted-ca-bundle
+       configMap:
+         name: reporting-operator-trusted-ca-bundle
+         items:
+         - key: ca-bundle.crt
+           path: ca-bundle.crt
+{{- end }}
 {{- if $operatorValues.spec.config.hive.tls.enabled }}
       - name: hive-tls
         secret:
--- a/images/metering-ansible-operator/roles/meteringconfig/tasks/configure_networking.yml
+++ b/images/metering-ansible-operator/roles/meteringconfig/tasks/configure_networking.yml
@@ -68,4 +68,8 @@
     _ocp_enabled_global_proxy_http_proxy: "{{ _cluster_proxy_config.status.httpProxy }}"
     _ocp_enabled_global_proxy_https_proxy: "{{ _cluster_proxy_config.status.httpsProxy }}"
   no_log: true
-  when: cluster_proxy_config is defined and cluster_proxy_config.resources | length > 0 and cluster_proxy_config.status is defined
+  when:
+  - cluster_proxy_config is defined
+  - cluster_proxy_config.resources | length > 0
+  - _cluster_proxy_config.status is defined
+  - _cluster_proxy_config.status | length > 0
--- a/charts/openshift-metering/values.yaml
+++ b/charts/openshift-metering/values.yaml
@@ -7,7 +7,6 @@ unsupportedFeatures:
 disableOCPFeatures: false
 
 useIPV6Networking: false
-useGlobalProxyNetworking: false
 
 networking:
   useGlobalProxyNetworking: false


cmd/reporting-operator/start.go
```diff
@@ -234,3 +234,4 @@ func newLogger() log.FieldLogger {
 	return logger
 
 }
+
```
timflannagan1: We don't need this change.

charts/openshift-metering/templates/reporting-operator/reporting-operator-deployment.yaml
```diff
@@ -367,6 +367,8 @@ spec:
           name: prometheus-bearer-token
           subPath: token
 {{- end }}
+        - mountPath: /var/run/reporting-operator/proxy/user-ca-bundle
```
timflannagan1: Mounting this volume is contingent on whether or not a cluster-wide Proxy is enabled, and whether the administrator specified an additional CA trust bundle (i.e. `spec.trustedCA`).

charts/openshift-metering/templates/reporting-operator/reporting-operator-deployment.yaml
```diff
@@ -441,6 +443,9 @@ spec:
 {{- end }}
 {{- end }}{{/* end of authProxy.enabled */}}
       volumes:
+      - name: proxy-user-ca-bundle
```
timflannagan1: Same situation describe above, we need to add templating around this volume.

charts/openshift-metering/templates/presto/presto-coordinator-config.yaml
```diff
@@ -59,6 +59,17 @@ data:
...
+{{- if .Values.presto.spec.config.tls.enabled }}
+    -Dhttp.proxyHost=networking.useGlobalProxyNetworking.proxy.config.http_proxy.hostname
+    -Dhttp.proxyPort=networking.useGlobalProxyNetworking.proxy.config.http_proxy.port
+    -Dhttp.proxyUser=networking.useGlobalProxyNetworking.proxy.config.http_proxy.username
+    -Dhttp.proxyPassword=networking.useGlobalProxyNetworking.proxy.config.http_proxy.password
+{{ else }}
+    -Dhttps.proxyHost=networking.useGlobalProxyNetworking.proxy.config.https_proxy.hostname
+    -Dhttps.proxyPort=networking.useGlobalProxyNetworking.proxy.config.https_proxy.port
+    -Dhttps.proxyUser=networking.useGlobalProxyNetworking.proxy.config.https_proxy.username
+    -Dhttps.proxyPassword=networking.useGlobalProxyNetworking.proxy.config.https_proxy.password
```
timflannagan1: A couple of things to note about this current iteration of the jvm configuration:
1.) We only want to set these properties when a cluster-wide Proxy is enabled, so we do care about whether or not TLS is enabled for Presto, but in the overall context of the Proxy being enabled. In more concrete terms:
```
if proxy is enabled:
  if tls is enabled:
    set `-Dhttps.proxy*` configurations
  else:
    set `-Dhttp.proxy*` configurations
```
2.) Without templating (`{{ ... }}`), you're essentially just passing this string to the jvm configuration, and not injecting any values from values.yaml.

charts/openshift-metering/templates/reporting-operator/reporting-operator-deployment.yaml
```diff
@@ -367,6 +367,11 @@ spec:
           name: prometheus-bearer-token
           subPath: token
 {{- end }}
+{{- if .Values.networking.useGlobalProxyNetworking.true }}
+{{ - if .Values.networking.proxy.config.trusted_ca_bundle != "" }}
```
timflannagan1:
```suggestion
{{- if .Values.networking.proxy.config.trusted_ca_bundle != "" }}
```

charts/openshift-metering/templates/reporting-operator/reporting-operator-deployment.yaml
```diff
@@ -440,7 +445,13 @@ spec:
           name: authenticated-emails-secret
 {{- end }}
 {{- end }}{{/* end of authProxy.enabled */}}
+{{- if .Values.networking.useGlobalProxyNetworking.true }}
+{{ - if .Values.networking.proxy.config.trusted_ca_bundle != "" }}
```
timflannagan1:
```suggestion
{{- if .Values.networking.proxy.config.trusted_ca_bundle != "" }}
```

charts/openshift-metering/templates/reporting-operator/reporting-operator-deployment.yaml
```diff
@@ -440,7 +445,13 @@ spec:
           name: authenticated-emails-secret
 {{- end }}
 {{- end }}{{/* end of authProxy.enabled */}}
+{{- if .Values.networking.useGlobalProxyNetworking.true }}
```
timflannagan1:
```suggestion
{{- if .Values.networking.useGlobalProxyNetworking }}
```

charts/openshift-metering/templates/presto/presto-coordinator-config.yaml
```diff
@@ -59,6 +59,18 @@ data:
 {{- end }}
 
   jvm.config: |
+{{- if .Values.networking.useGlobalProxyNetworking.true }}
```
timflannagan1:
```suggestion
{{- if .Values.networking.useGlobalProxyNetworking }}
```

charts/openshift-metering/templates/presto/presto-coordinator-config.yaml
```diff
@@ -59,6 +59,18 @@ data:
...
+{{- if .Values.presto.spec.config.tls.enabled }}
+    -Dhttp.proxyHost=networking.useGlobalProxyNetworking.proxy.config.http_proxy.hostname
+    -Dhttp.proxyPort=networking.useGlobalProxyNetworking.proxy.config.http_proxy.port
+    -Dhttp.proxyUser=networking.useGlobalProxyNetworking.proxy.config.http_proxy.username
+    -Dhttp.proxyPassword=networking.useGlobalProxyNetworking.proxy.config.http_proxy.password
+{{ else }}
+    -Dhttps.proxyHost=networking.useGlobalProxyNetworking.proxy.config.https_proxy.hostname
+    -Dhttps.proxyPort=networking.useGlobalProxyNetworking.proxy.config.https_proxy.port
+    -Dhttps.proxyUser=networking.useGlobalProxyNetworking.proxy.config.https_proxy.username
+    -Dhttps.proxyPassword=networking.useGlobalProxyNetworking.proxy.config.https_proxy.password
```
timflannagan1: (this applies to the entire block of code, this is an invalid path)
```suggestion
    -Dhttps.proxyPassword=networking.proxy.config.https_proxy.password
```

charts/openshift-metering/templates/hadoop/_hadoop_config_helpers.tpl
```diff
@@ -108,5 +108,46 @@
       <name>fs.gs.reported.permissions</name>
       <value>733</value>
   </property>
+{{- if .Values.presto.spec.config.tls.enabled }}
```
timflannagan1:
```suggestion
{{- if .Values.networking.useGlobalProxyNetworking }}
{{- if .Values.presto.spec.config.tls.enabled }}
```

charts/openshift-metering/templates/presto/presto-coordinator-config.yaml
```diff
@@ -59,6 +59,19 @@ data:
...
+{{- if .Values.presto.spec.config.tls.enabled }}
+    -Dhttp.proxyHost=networking.proxy.config.http_proxy.hostname
+    -Dhttp.proxyPort=networking.proxy.config.http_proxy.port
+    -Dhttp.proxyUser=networking.proxy.config.http_proxy.username
+    -Dhttp.proxyPassword=networking.proxy.config.http_proxy.password
+{{ else }}
+    -Dhttps.proxyHost=networking.proxy.config.https_proxy.hostname
+    -Dhttps.proxyPort=networking.proxy.config.https_proxy.port
+    -Dhttps.proxyUser=networking.proxy.config.https_proxy.username
+    -Dhttps.proxyPassword=networking.proxy.config.https_proxy.password
```
timflannagan1: You need to add templating to inject the value of these properties
(this applies to this entire jvm configuration code block)
```suggestion
    -Dhttps.proxyPassword={{ .Values.networking.proxy.config.https_proxy.password }}
```

charts/openshift-metering/templates/reporting-operator/reporting-operator-deployment.yaml
```diff
@@ -440,7 +444,12 @@ spec:
           name: authenticated-emails-secret
 {{- end }}
 {{- end }}{{/* end of authProxy.enabled */}}
+{{- if and .Values.networking.useGlobalProxyNetworking .Values.networking.proxy.config.trusted_ca_bundle }}
       volumes:
```
timflannagan1: You need to move the `volumes` key outside of the templating. In the case where either `useGlobalProxyNetworking` or `trusted_ca_bundle` evaluate to false, the `volumes` key will be undefined, which will result in a templating error.

charts/openshift-metering/templates/reporting-operator/reporting-operator-deployment.yaml
```diff
@@ -77,7 +77,7 @@ spec:
       containers:
       - name: reporting-operator
         image: "{{ include "reporting-operator-image" . }}"
-        imagePullPolicy: {{ $operatorValues.spec.image.pullPolicy }}
+        llPolicy: {{ $operatorValues.spec.image.pullPolicy }}
```
timflannagan1: huh?

charts/openshift-metering/templates/reporting-operator/reporting-operator-deployment.yaml
```diff
@@ -337,7 +337,7 @@ spec:
         livenessProbe:
 {{ toYaml $operatorValues.spec.livenessProbe | indent 10 }}
 {{- end }}
-        volumeMounts:
+        Mounts:
```
timflannagan1: revert

charts/openshift-metering/templates/reporting-operator/reporting-operator-deployment.yaml
```diff
@@ -426,7 +430,7 @@ spec:
           containerPort: 8081
         resources:
 {{ toYaml $operatorValues.spec.authProxy.resources | indent 10 }}
-        volumeMounts:
+        Mounts:
```
timflannagan1: revert

charts/openshift-metering/templates/reporting-operator/reporting-operator-deployment.yaml
timflannagan1: *revert

charts/openshift-metering/templates/presto/presto-coordinator-config.yaml
timflannagan1: This still needs to be addressed.

charts/openshift-metering/templates/reporting-operator/reporting-operator-deployment.yaml
```diff
@@ -441,6 +445,11 @@ spec:
 {{- end }}
 {{- end }}{{/* end of authProxy.enabled */}}
       volumes:
+{{- if and (.Values.networking.useGlobalProxyNetworking) (not eq .Values.networking.proxy.config.trusted_ca_bundle "") }}
```
timflannagan1: I'm not sure this templating logic is necessary (or valid). Why can't we just do the original,
```yaml
{{- if and .Values.networking.useGlobalProxyNetworking .Values.networking.proxy.config.trusted_ca_bundle }}
```

charts/openshift-metering/templates/hadoop/_hadoop_config_helpers.tpl
```diff
@@ -108,5 +108,40 @@
       <name>fs.gs.reported.permissions</name>
       <value>733</value>
   </property>
+  <property>
+      <name>fs.s3a.proxy.host</name>
+{{- if and .Values.networking.useGlobalProxyNetworking .Values.presto.spec.config.tls.enabled }}
```
timflannagan1, line 113:
```suggestion
  <property>
{{- if and .Values.networking.useGlobalProxyNetworking .Values.presto.spec.config.tls.enabled }}
      <name>fs.s3a.proxy.host</name>
```

charts/openshift-metering/templates/hadoop/_hadoop_config_helpers.tpl
```diff
@@ -108,5 +108,40 @@
...
+      <value>{{ .Values.networking.proxy.config.https_proxy.username }}</value>
+  </property>
+  <property>
+      <name>fs.s3a.proxy.password</name>
+      <value>{{ .Values.networking.proxy.config.https_proxy.password }}</value>
+  </property>
+  <property>
+      <name>fs.s3a.proxy.host</name>
+{{- else if .Values.networking.useGlobalProxyNetworking }}
+      <value>{{ .Values.networking.proxy.config.http_proxy.hostname }}</value>
```
timflannagan1, line 131:
```suggestion
{{- else if .Values.networking.useGlobalProxyNetworking }}
      <name>fs.s3a.proxy.host</name>
      <value>{{ .Values.networking.proxy.config.http_proxy.hostname }}</value>
```

charts/openshift-metering/templates/hadoop/_hadoop_config_helpers.tpl
```diff
@@ -108,5 +108,40 @@
...
+      <name>fs.s3a.proxy.username</name>
+      <value>{{ .Values.networking.proxy.config.https_proxy.username }}</value>
+  </property>
+  <property>
+      <name>fs.s3a.proxy.password</name>
+      <value>{{ .Values.networking.proxy.config.https_proxy.password }}</value>
+  </property>
+  <property>
+{{- else if .Values.networking.useGlobalProxyNetworking }}
+      <name>fs.s3a.proxy.host</name>
```
timflannagan1, line 130: You need to swap lines 128 with 129. The `{{- else if ... }}` conditional is splitting up this property configuration.
```suggestion
{{- else if .Values.networking.useGlobalProxyNetworking }}
  <property>
      <name>fs.s3a.proxy.host</name>
```

charts/openshift-metering/templates/reporting-operator/reporting-operator-deployment.yaml
```diff
@@ -441,6 +445,11 @@ spec:
 {{- end }}
 {{- end }}{{/* end of authProxy.enabled */}}
       volumes:
+{{- if and .Values.networking.useGlobalProxyNetworking .Values.networking.proxy.config.trusted_ca_bundle }}
+        - name: proxy-user-ca-bundle
+          configMap:
+            name: user-ca-bundle
+{{- end }}
```
timflannagan1, line 452: This is mis-idented compared to the other `volumes` listed here.

charts/openshift-metering/templates/reporting-operator/reporting-operator-deployment.yaml
```diff
@@ -441,6 +445,11 @@ spec:
 {{- end }}
 {{- end }}{{/* end of authProxy.enabled */}}
       volumes:
+{{- if and .Values.networking.useGlobalProxyNetworking .Values.networking.proxy.config.trusted_ca_bundle }}
+        - name: proxy-user-ca-bundle
+          configMap:
+            name: user-ca-bundle
```
timflannagan1, line 451: We most likely don't want the ConfigMap we're referencing to also match the `user-ca-bundle` ConfigMap in the `openshift-config` namespace. Let's add something like a `reporting-operator-*` or `metering-operator-*` prefix to distinguish between the two.

charts/openshift-metering/templates/reporting-operator/reporting-operator-deployment.yaml
```diff
@@ -446,9 +446,9 @@ spec:
 {{- end }}{{/* end of authProxy.enabled */}}
       volumes:
 {{- if and .Values.networking.useGlobalProxyNetworking .Values.networking.proxy.config.trusted_ca_bundle }}
-        - name: proxy-user-ca-bundle
-          configMap:
-            name: user-ca-bundle
+      - name: proxy-user-ca-bundle
+        configMap:
+          name: user-ca-bundle
```
timflannagan1, line 451:
```suggestion
          name: reporting-operator-trusted-ca-bundle
```

charts/openshift-metering/templates/reporting-operator/reporting-operator-deployment.yaml
```diff
@@ -441,6 +445,11 @@ spec:
 {{- end }}
 {{- end }}{{/* end of authProxy.enabled */}}
       volumes:
+{{- if and .Values.networking.useGlobalProxyNetworking .Values.networking.proxy.config.trusted_ca_bundle }}
+      - name: proxy-user-ca-bundle
```
timflannagan1, line 449: `name: reporting-operator-trusted-ca-bundle`

charts/openshift-metering/templates/reporting-operator/reporting-operator-deployment.yaml
```diff
@@ -441,6 +445,11 @@ spec:
...
+        configMap:
+          name: user-ca-bundle
```
timflannagan1, line 451: We also don't need to mount the entire ConfigMap as we're only interested in the `data["ca-bundle.crt"]` item.

timflannagan1: We may also want to mark this as `readOnly: true` as there shouldn't be any write options needed.

charts/openshift-metering/templates/reporting-operator/reporting-operator-deployment.yaml
```diff
@@ -367,6 +367,10 @@ spec:
           name: prometheus-bearer-token
           subPath: token
 {{- end }}
+{{- if and .Values.networking.useGlobalProxyNetworking .Values.networking.proxy.config.trusted_ca_bundle }}
+        - mountPath: /etc/pki/ca-trust/extracted/pem
+          name: reporting-operator-trusted-ca-bundle
```
timflannagan1, line 372:
```suggestion
        - mountPath: /etc/pki/ca-trust/extracted/pem
          name: reporting-operator-trusted-ca-bundle
          readOnly: true
```

charts/openshift-metering/templates/reporting-operator/reporting-operator-deployment.yaml
EmilyM1: That is the only item in the user-ca-bundle

charts/openshift-metering/templates/reporting-operator/reporting-operator-deployment.yaml
EmilyM1: Good call!

charts/openshift-metering/templates/reporting-operator/reporting-operator-deployment.yaml
```diff
@@ -441,6 +446,13 @@ spec:
 {{- end }}
 {{- end }}{{/* end of authProxy.enabled */}}
       volumes:
+{{- if and .Values.networking.useGlobalProxyNetworking .Values.networking.proxy.config.trusted_ca_bundle }}
+      - name: reporting-operator-trusted-ca-bundle
+        valueFrom:
+          configMapKeyRef:
+            name: reporting-operator-trusted-ca-bundle
+            key: data["ca-bundle.crt"]
```
timflannagan1, line 454:
```suggestion
        configMap:
          name: reporting-operator-trusted-ca-bundle
          items:
            - key: ca-bundle.crt
              path: ca-bundle.crt
```

timflannagan1: https://kubernetes.io/docs/concepts/storage/volumes/#configmap

images/metering-ansible-operator/roles/meteringconfig/tasks/configure_networking.yml
```diff
@@ -68,4 +68,4 @@
     _ocp_enabled_global_proxy_http_proxy: "{{ _cluster_proxy_config.status.httpProxy }}"
     _ocp_enabled_global_proxy_https_proxy: "{{ _cluster_proxy_config.status.httpsProxy }}"
   no_log: true
-  when: cluster_proxy_config is defined and cluster_proxy_config.resources | length > 0 and cluster_proxy_config.status is defined
+  when: cluster_proxy_config is defined and (cluster_proxy_config.resources | length) and _cluster_proxy_config.status is defined
```
timflannagan1, line 71: Something is off about this conditional (from the metering-operator Pod logs):
```
TASK [meteringconfig : Check if the cluster-wide Proxy configured] *************
[1;30mtask path: /opt/ansible/roles/meteringconfig/tasks/configure_networking.yml:33[0m
Wednesday 08 April 2020  21:17:58 +0000 (0:00:00.048)       0:00:07.940 ******* 
[0;32mok: [localhost] => {"changed": false, "resources": [{"apiVersion": "config.openshift.io/v1", "kind": "Proxy", "metadata": {"creationTimestamp": "2020-04-08T20:38:32Z", "generation": 1, "name": "cluster", "resourceVersion": "456", "selfLink": "/apis/config.openshift.io/v1/proxies/cluster", "uid": "09160338-f4e6-4f1f-81bb-ecf077de1713"}, "spec": {"trustedCA": {"name": ""}}, "status": {}}]}[0m

TASK [meteringconfig : set_fact] ***********************************************
[1;30mtask path: /opt/ansible/roles/meteringconfig/tasks/configure_networking.yml:48[0m
Wednesday 08 April 2020  21:17:58 +0000 (0:00:00.866)       0:00:08.806 ******* 
[0;31mfatal: [localhost]: FAILED! => {"censored": "the output has been hidden due to the fact that 'no_log: true' was specified for this result"}[0m
```

charts/openshift-metering/templates/reporting-operator/reporting-operator-deployment.yaml
EmilyM1: wouldn't the key still be data?

images/metering-ansible-operator/roles/meteringconfig/tasks/configure_networking.yml
timflannagan1, line 71: Looking back, this check isn't sufficient. Currently, we check whether or not this cluster-wide, proxy/cluster object exists, which is always the case when deploying Metering on OCP (unless a cluster-admin accidentally deletes this object), and whether or not this same object has a `.status` dictionary key, which also should also always be the case.

The problem, however, is when you're running a Metering installation against an OCP cluster that _hasn't_ setup a cluster-wide proxy, i.e. the proxy/cluster object will exist, but the `.status` field will be empty. We need to update the conditional to properly reflect that last point, as the current implementation only checks whether or not this dictionary key exists, and not verifying whether this key is non-empty.

charts/openshift-metering/templates/reporting-operator/reporting-operator-deployment.yaml
EmilyM1: Nah, this needs the configMapKeyRef

charts/openshift-metering/templates/reporting-operator/reporting-operator-deployment.yaml
```diff
@@ -441,6 +446,13 @@ spec:
 {{- end }}
 {{- end }}{{/* end of authProxy.enabled */}}
       volumes:
+{{- if and .Values.networking.useGlobalProxyNetworking .Values.networking.proxy.config.trusted_ca_bundle }}
+      - name: reporting-operator-trusted-ca-bundle
+        valueFrom:
+          configMapKeyRef:
+            name: reporting-operator-trusted-ca-bundle
```
timflannagan1, line 453: Let's add a key to the `reporting-operator-config` ConfigMap that points to the `/etc/pki/ca-trust/extracted/pem/ca-bundle.crt`, and change the `name: reporting-operator-trusted-ca-bundle` to `name: reporting-operator-config`, and the `key` value is now the name of the variable that points to the `ca-bundle.crt`.

images/metering-ansible-operator/roles/meteringconfig/tasks/configure_networking.yml
timflannagan1:
```suggestion
  when:
  - cluster_proxy_config is defined
  - cluster_proxy_config.resources | length > 0
  - _cluster_proxy_config.status is defined
  - _cluster_proxy_config.status | length > 0

```

EmilyM1: hey just curious... why is one cluster_proxy_config and one _cluster_proxy_config? Thanks

PR #1115, opened 2020-03-03
images: Replace all instantiations of the k8s_facts Ansible module with k8s_info.

timflannagan1:
The `k8s_fact` command is going to be depreciated in the next couple of Ansible releases, so this is just being pro-active and switching to the now recommended `k8s_info` module.

From the `k8s_info` Ansible documentation [1]: 
"This module was called k8s_facts before Ansible 2.9. The usage did not change."

[1] https://docs.ansible.com/ansible/latest/modules/k8s_info_module.html


timflannagan1:
This can be sidelined for now as we're not using Ansible 2.9 anymore and the `k8s_info` was introduced in 2.9.

PR #1128, opened 2020-03-20
Remove the unused .helmignore files and update the .gitignore.

timflannagan1:


.gitignore
```diff
@@ -28,9 +22,7 @@ _tmp
 /client-gen
 /lister-gen
 /informer-gen
-/metering-custom.yaml
```
EmilyM1: I feel like we do still want this gitignored. It's a common name for the CR and the CR changes fairly often.

PR #1133, opened 2020-03-21
Migrate from dep to Go modules.

bentito:
WIP
`go build .\...` works
`make unit` works
`make verify` works (e2e-aws tests pass) but only with verify-codegen.sh & update-codegen.sh not run, otherwise if just update-codegen.sh run, fails on `go vet`.

Investigating why codegen is diff CI vs. local now.


bentito:
`Gopkg.lock` is split out into its own PR now: https://github.com/operator-framework/operator-metering/pull/1135

Makefile
```diff
@@ -95,8 +95,7 @@ fmt:
 	find . -name '*.go' -not -path "./vendor/*" | xargs gofmt -w
 
 # Update dependencies
-vendor: Gopkg.toml
-	dep ensure -v
+vendor:	go mod vendor
```
timflannagan1:
```suggestion
vendor:
    go mod tidy
    go mod vendor
    go mod verify
```

We may want to add the verbose option (`-v`) to all of those commands too.

timflannagan1:
/retitle *: Migrate from dep to Go modules.

Makefile
```diff
@@ -92,11 +92,12 @@ metering-ansible-operator-docker-build: $(METERING_ANSIBLE_OPERATOR_DOCKERFILE)
...
 # Runs gofmt on all files in project except vendored source
 fmt:
+	@echo path: $(shell pwd)
 	find . -name '*.go' -not -path "./vendor/*" | xargs gofmt -w
 
 # Update dependencies
-vendor: Gopkg.toml
-	dep ensure -v
+vendor:
+	go mod vendor
```
timflannagan1:
```suggestion
	go mod tidy
	go mod vendor
	go mod verify
```

manifests/deploy/openshift/olm/metering.catalogsourceconfig.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: operators.coreos.com/v1
+kind: CatalogSourceConfig
+metadata:
+  name: installed-redhat-metering-operators-openshift-metering
+  namespace: openshift-marketplace
+spec:
+  targetNamespace: openshift-metering
+  packages: metering-ocp
+
```
timflannagan1: We reverted this commit, so this needs to be rebased against the current master

Makefile
bentito: Not sure we need explicit calls to `go mod tidy` or `go mod verify` because eventually we call `go build` which runs both those with `go.mod` present, right?

manifests/deploy/openshift/olm/metering.catalogsourceconfig.yaml
bentito: Thank you for calling MCI... please hold for rebase.

Makefile
```diff
@@ -143,13 +144,7 @@ e2e-docker: metering-src-docker-build
 	docker cp metering-e2e-docker:/out bin/e2e-docker-test-output
 	docker rm metering-e2e-docker
 
-vet:
-	go vet $(GO_PKG)/cmd/... $(GO_PKG)/pkg/...
-
-verify: verify-codegen verify-olm-manifests verify-helm-templates fmt vet
-	@echo Checking for unstaged changes
-	# validates no unstaged changes exist in $(VERIFY_FILE_PATHS)
-	git diff --stat HEAD --ignore-submodules --exit-code -- $(VERIFY_FILE_PATHS)
```
timflannagan1: Hmm, I still think we want something similar to what `git diff` is doing here, even if it's only for validating go module changes, and more specifically checking if there are any changes made to go.mod, go.sum, vendor/ that are unstaged:
```make
...
    @git diff --exit-code -- go.sum go.mod vendor/
```

Makefile
timflannagan1: I'm also thinking we may want to build binaries using `go build -mod=vendor`.

hack/update-codegen.sh
```diff
@@ -19,23 +19,36 @@ set -o nounset
...
 set -x
+# Because go mod sux, we have to fake the vendor for generator in order to be able to build it...
+mv ${CODEGEN_PKG}/generate-groups.sh ${CODEGEN_PKG}/generate-groups.sh.orig
+sed 's/ go install/#go install/g' ${CODEGEN_PKG}/generate-groups.sh.orig > ${CODEGEN_PKG}/generate-groups.sh
+function cleanup {
+  mv ${CODEGEN_PKG}/generate-groups.sh.orig ${CODEGEN_PKG}/generate-groups.sh
+}
+trap cleanup EXIT
+
+go install ./${CODEGEN_PKG}/cmd/{defaulter-gen,client-gen,lister-gen,informer-gen,deepcopy-gen}
```
timflannagan1: The more I think about this, the less I like this approach as we're essentially running `go install ...` on this package for every CI run which is slow (and as we've seen, prone to verify errors). 

I haven't tested this yet, but I wonder what's stopping us from copying the generate-groups.sh script from `vendor/k8s.io/code-generator/generate-groups.sh` and simply sourcing our hack/ version of that script in update-codegen.sh as a workaround to the "no such file" error I see in `make verify` errors.

Makefile
```diff
@@ -143,13 +144,7 @@ e2e-docker: metering-src-docker-build
 	docker cp metering-e2e-docker:/out bin/e2e-docker-test-output
 	docker rm metering-e2e-docker
 
-vet:
-	go vet $(GO_PKG)/cmd/... $(GO_PKG)/pkg/...
```
timflannagan1: Can you remind why we're removing `make vet` as a verify check?

Makefile
bentito: Ok after reading more here:
https://github.com/golang/go/wiki/Modules#how-to-prepare-for-a-release (really good write-up of releasing with Go modules, I hadn't found before) I believe go mod tidy is needed to removed unneeded deps.
And after reading more here: https://github.com/golang/go/wiki/Modules#how-do-i-use-vendoring-with-modules-is-vendoring-going-away -- rationale about committing the go.sum file, I agree we should use it to verify what's on the disk for deps.

bentito: `go mod tidy` and `go mod verify` added to `vendor` make target in 63cc9b77

bentito: Since we're not overwriting `GOFLAGS`, are we then getting `-mod=vendor` as a flag on `go build` when it actually runs in CI?

timflannagan1: I assumed you we would update the Makefile `GO_BUILD_ARGS := -ldflags '-extldflags "-static"'` variable to use `-mod=vendor`.

hack/update-codegen.sh
bentito: Well above in the current `update-codegen.sh` we have:

```bash
mv ${CODEGEN_PKG}/generate-groups.sh ${CODEGEN_PKG}/generate-groups.sh.orig
sed 's/ go install/#go install/g' ${CODEGEN_PKG}/generate-groups.sh.orig > ${CODEGEN_PKG}/generate-groups.sh
```
That's taking the `vendor/k8s.io/code-generator/generate-groups.sh` disabling `go install`, running `go install` ourselves (installing the generator commands) and then calling generate-groups as needed. When you say we're `go install...` this package, the this is only the generator commands, so not too bad, right?

Makefile
bentito: My reasoning was that since we run "go test" elsewhere, all the code gets "go vet" run on it with a limited "high-confidence subset" of tests:

https://golang.org/pkg/cmd/go/internal/test/

Makefile
bentito: I think I am not getting something? How would there come to be module changes, or changes made to go.mod, go.sum or vendor that are unstaged? Like by running `go build` or `go mod tidy` as part of the CI process? I guess those commands could cause there to be diffs but why do they matter? The committed code will produce those changes when run through the CI process, why worry about essentially build artifacts?

timflannagan1: It's essentially a stop-gap to ensure that changes made to dependencies _get_ staged. In the case where you're adding a new dependency and you follow the intended workflow, i.e. you run `make vendor` and then commit those changes, then there's no issue as there's going to be no unstaged changes.

We do the same thing with `make metering-manifests`: if you make a change to a helm chart, we expect you to run that command so that the corresponding manifests also get updated. If you don't follow that workflow, then verify will complain about it.

Makefile
timflannagan1: This is fine reasoning, but if I'm making a quick, one-off change locally, I'd rather run `make verify` before deploying Metering with those updates than `go test ./...`.

bentito: Then it will be even faster to run `make verify` because it won't run `go vet` at that time?

Makefile
bentito: Okay, I put it back in 9c6beb3 Pretty sure it will cause a fail, but we can look at how to avoid the fails after I guess.

Makefile
bentito: okay, done in to be pushed commit, waiting on CI tests to see results of `git diff` addition first.

hack/update-codegen.sh
```diff
@@ -19,23 +19,36 @@ set -o nounset
...
+    "metering:v1" \
+    --go-header-file hack/boilerplate.go.txt
+    ${verify}
 
 # generate-groups doesn't do defaulters
 echo "Generating defaulters"
 ${GOPATH}/bin/defaulter-gen \
     --input-dirs "${SCRIPT_PACKAGE}/pkg/apis/metering/v1" \
     -O zz_generated.defaults \
-    --go-header-file ${SCRIPT_ROOT}/hack/boilerplate.go.txt
```
timflannagan1: "${SCRIPT_ROOT}/hack/boilerplate.go.txt` should still work?

hack/verify-codegen.sh
```diff
@@ -18,32 +18,5 @@ set -o errexit
...
-ret=0
-diff -Naupr "${DIFFROOT}" "${TMP_DIFFROOT}" || ret=$?
-cp -a "${TMP_DIFFROOT}"/* "${DIFFROOT}"
-if [[ $ret -eq 0 ]]
-then
-  echo "${DIFFROOT} up to date."
-else
-  echo "${DIFFROOT} is out of date. Please run hack/update-codegen.sh"
-  exit 1
-fi
```
timflannagan1: Can we check if the original verify-codegen.sh script works with the current version of `update-codegen.sh`?

Makefile
bentito: used Docker to run make update-codegen and commited results

hack/update-codegen.sh
bentito: Will try for restoring all of verify-codegen.sh as-was on master. Just want to see this as-is pass e2e first.

hack/verify-codegen.sh
bentito: yep

bentito:
man, if only I knew there was a cute Corgi at the end of this!

PR #1140, opened 2020-03-23
*: Add support for running upgrade tests in the end-to-end suite.

timflannagan1:
### Description
This would update the existing end-to-end suite to support running upgrade tests against a manual OLM installation that references an existing channel, to the most up-to-date payload, and run reporting tests against both installations.


test/deployframework/context.go
```diff
@@ -262,44 +261,137 @@ func (ctx *DeployerCtx) Setup(expectInstallErr bool) (*reportingframework.Report
 	return rf, nil
 }
 
+// Upgrade is a method that is reponsible for creating the necessary
```
bentito, line 264:
```suggestion
// Upgrade is a method that is responsible for creating the necessary
```

test/deployframework/framework.go
```diff
@@ -61,7 +64,7 @@ type DeployFramework struct {
 }
 
 // New is the constructor function that creates and returns a new DeployFramework object
-func New(logger logrus.FieldLogger, runLocal, runDevSetup bool, nsPrefix, repoDir, kubeconfig string) (*DeployFramework, error) {
+func New(logger logrus.FieldLogger, runLocal, runDevSetup bool, nsPrefix, repoDir, repoVersion, kubeconfig string) (*DeployFramework, error) {
```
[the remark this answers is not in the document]
timflannagan1: I think we should hold off on renaming this constructor as instantiating a new framework is done by calling `deployframework.New(...)` which is the correct usage (as deployframework essentially acts a namespace). If this was a method, instead of a function I would agree that it should be changed to something like `NewDeployerFramework`.

Makefile
```diff
@@ -197,14 +215,8 @@ build-reporting-operator: $(REPORTING_OPERATOR_BIN_DEPENDENCIES) $(GOFILES)
...
 
-metering-manifests:
-	export \
-		METERING_OPERATOR_IMAGE_REPO=$(METERING_OPERATOR_IMAGE_REPO) \
-		METERING_OPERATOR_IMAGE_TAG=$(METERING_OPERATOR_IMAGE_TAG); \
-	./hack/generate-metering-manifests.sh
-
 $(DEPLOY_METERING_BIN_OUT): $(GOFILES)
-	go build -o $(DEPLOY_METERING_BIN_OUT) $(GO_PKG)/cmd/deploy-metering
+	go build $(GO_BUILD_ARGS) -o $(DEPLOY_METERING_BIN_OUT) $(DEPLOY_METERING_PKG)
```
bentito, line 219: Couldn't this just be:
```suggestion
	go build $(GO_BUILD_ARGS) -o $(DEPLOY_METERING_BIN_OUT) ./...
```

test/deployframework/framework.go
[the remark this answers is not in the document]
bentito: ah okay

hack/create-upgrade-configmap.sh
```diff
@@ -0,0 +1,37 @@
...
+
+NAME=${NAME:-"metering-ocp"}
+NAMESPACE=${NAMESPACE:-"openshift-metering"}
+
+ROOT_DIR=$(dirname "${BASH_SOURCE}")/..
+MANIFESTS_DIR="${MANIFESTS_DIR:-${ROOT_DIR}/manifests/deploy/openshift/olm/bundle/}"
+MANIFEST_VERSION=${MANIFEST_VERSION:-$(basename $(find $MANIFESTS_DIR -type d | sort -r | head -n 1))}
+
+CRD=$(sed '/^#!.*$/d' $MANIFESTS_DIR/$MANIFEST_VERSION/*crd.yaml | grep -v -- "---" | indent apiVersion)
+CSV=$(sed '/^#!.*$/d' $MANIFESTS_DIR/$MANIFEST_VERSION/*version.yaml | sed -e "s,imagePullPolicy: IfNotPresent,imagePullPolicy: Always,"  | sed 's/namespace: placeholder/namespace: '$namespace'/' |grep -v -- "---" |  indent apiVersion)
```
bentito, line 21: Shouldn't `$namespace` be referring to the all caps version `${NAMESPACE}`

hack/create-upgrade-configmap.sh
```diff
@@ -0,0 +1,37 @@
...
+    echo "Overriding the metering-operator deployment images with ${IMAGE_OVERRIDE}"
+    CSV=$(echo "$CSV" | sed -e "s~containerImage:.*~containerImage: ${IMAGE_OVERRIDE}~" | indent apiVersion)
+    CSV=$(echo "$CSV" | sed -e "s~image:.*~image: ${IMAGE_OVERRIDE}\n~" | indent ApiVersion)
+fi
+
+cat <<EOF | sed 's/^  *$//' | oc apply -n $NAMESPACE -f -
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  name: $NAME
```
bentito, line 34: To match up with the rest of the script and improves shell convention, these should be in the form `${VAR}`, right?

hack/run-test-cleanup.sh
```diff
@@ -126,3 +126,40 @@ while read -r report; do
         fi
     fi
 done <<< "$REPORTS"
+
```
bentito, line 129: This script does no cleanup, it does test output capture, should be renamed, IMO.

test/deployframework/context.go
```diff
@@ -169,7 +178,6 @@ func (ctx *DeployerCtx) Setup(expectInstallErr bool) (*reportingframework.Report
 	if !installErr {
 		if ctx.RunTestLocal {
 			ctx.LocalCtx = ctx.NewLocalCtx()
-
```
bentito, line 172: 👍 vertical space reduction

test/deployframework/context.go
```diff
@@ -262,44 +264,140 @@ func (ctx *DeployerCtx) Setup(expectInstallErr bool) (*reportingframework.Report
 	return rf, nil
 }
 
+// Upgrade is a method that is responsible for creating the necessary
+// resources to upgrade an existing Metering install to use the most
+// up-to-date manifest files (CRDs, CSV, package.yaml).
+func (ctx *DeployerCtx) Upgrade(packageName, repoVersion string, purgeReports, purgeReportDataSources bool) (*reportingframework.ReportingFramework, error) {
```
bentito, line 276: I can't tell how/where this method is actually called from?

bentito: found it, nvm

test/e2e/main_test.go
```diff
@@ -90,6 +96,206 @@ type InstallTestCase struct {
...
+	require.NoError(t, err, "creating the test case output directory should produce no error")
+
+	// create a pre-upgrade test case directory
+	preUpgradeTestOutputDir := filepath.Join(testCaseOutputBaseDir, preUpgradeTestDirName)
+	err = os.Mkdir(preUpgradeTestOutputDir, 0777)
+	require.NoError(t, err, "creating the test case output directory should produce no error")
+
+	testFuncNamespace := fmt.Sprintf("%s-%s", namespacePrefix, strings.ToLower(testCaseName))
+	if len(testFuncNamespace) > kubeNamespaceCharLimit {
+		require.Fail(t, "The length of the test function namespace exceeded the kube namespace limit of %d characters", kubeNamespaceCharLimit)
```
bentito, line 238: How about just warn on this? Kube will just truncate and carry on, won't it?

test/testhelpers/assert.go
```diff
@@ -71,12 +71,34 @@ func AssertReportResultsEqual(t *testing.T, expected, actual []map[string]interf
 	}
 }
 
-func AssertErrorContainsErrorMsgs(t *testing.T, err error, errMsgArr []string) {
+func AssertErrorContainsErrorMsgs(t *testing.T, err error, errMsgArr []string) bool {
```
bentito, line 74: I'd kinda like to see less use of this type of assertion, not more. We're asserting on specific strings making the tests more brittle, if someone comes along and changes an error message, we've got a failing test and not really for a good reason. One thing to do might be to adopt error types. That way we can assert on the error type and not the specific message. Maybe not for this PR, but just to keep in mind.

test/testhelpers/helpers.go
```diff
@@ -16,10 +16,10 @@ import (
 func NewReport(name, namespace, testQueryName string, reportStart, reportEnd *time.Time, status metering.ReportStatus, schedule *metering.ReportSchedule, runImmediately bool) *metering.Report {
 	var start, end *meta.Time
 	if reportStart != nil {
-		start = &meta.Time{*reportStart}
+		start = &meta.Time{Time: *reportStart}
```
bentito, line 19: Was this bad Go style without the cast?

test/e2e/main_test.go
timflannagan1: Unfortunately not, as attempting to create the namespace in the `deploy` package will error if the total number of characters in the namespace, exceeds the namespace character limit.

test/testhelpers/helpers.go
timflannagan1: This is bad Go-style as we're trying to assign to the `meta.Time{..}` object w/o specifying one of the object's key, so `go vet ./...` was complaining about this.

hack/run-test-cleanup.sh
timflannagan1: Yeah I've been wondering why I named that script `run-test-cleanup.sh` for a while now, definitely something that should be renamed to properly reflect what it's doing.

test/deployframework/context.go
```diff
@@ -262,44 +264,140 @@ func (ctx *DeployerCtx) Setup(expectInstallErr bool) (*reportingframework.Report
...
 	}
 
+	err = ctx.MustGatherMeteringResources(cleanupScriptName)
+	if err != nil {
+		errArr = append(errArr, fmt.Sprintf("failed to successfully gather all of the metering resources: %v", err))
+	}
+
 	// Check if the user wants to run the E2E suite using the developer setup.
 	// If true, we skip the process of deleting and logging of the metering
 	// resources that were provisioned during the manual install.
```
bentito, line 404: Is this comment block still valid and needed (Teardown has its own comment block) given the changes below?

test/deployframework/context.go
```diff
@@ -311,3 +409,27 @@ func (ctx *DeployerCtx) Teardown() error {
 
 	return nil
 }
+
+// TODO: does this need to be a ctx method? Probably not.
+// Is there any state present in this method? Maybe the base directory?
```
bentito: Wha?

test/deployframework/helpers.go
```diff
@@ -289,3 +284,291 @@ func cleanupLocalCmds(logger logrus.FieldLogger, commands ...exec.Cmd) error {
...
+		}
+		logger.Infof("Created the %s CatalogSource", name)
+	} else if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// VerifyCatalogSourcePod is a deployframework helper function that checks the @namespace
```
bentito, line 318:
```suggestion
// A deployframework helper function that checks the @namespace
```

test/deployframework/helpers.go
```diff
@@ -289,3 +284,291 @@ func cleanupLocalCmds(logger logrus.FieldLogger, commands ...exec.Cmd) error {
...
+		return true, nil
+	})
+	if err != nil {
+		return fmt.Errorf("failed to wait for the %s catalogsource Pod to become ready: %v", packageName, err)
+	}
+
+	return nil
+}
+
+// CreateUpgradeConfigMap is a helper function responsible for creating a ConfigMap
```
bentito, line 353:
```suggestion
// A helper function responsible for creating a ConfigMap
```

test/deployframework/helpers.go
```diff
@@ -289,3 +284,291 @@ func cleanupLocalCmds(logger logrus.FieldLogger, commands ...exec.Cmd) error {
...
+	}
+
+	scanner := bufio.NewScanner(stderr)
+	scanner.Split(bufio.ScanWords)
+	for scanner.Scan() {
+		m := scanner.Text()
+		fmt.Println(m)
+	}
+
+	err = cmd.Wait()
```
bentito, line 408: This shouldn't really wait forever should it?

test/deployframework/context.go
timflannagan1: That needs to be updated.

test/deployframework/helpers.go
```diff
@@ -289,3 +284,291 @@ func cleanupLocalCmds(logger logrus.FieldLogger, commands ...exec.Cmd) error {
...
+		return true, nil
+	})
+	if err != nil {
+		return fmt.Errorf("failed to wait for the %s subscription to finish updating in the %s namespace: %v", name, namespace, err)
+	}
+	return nil
+}
+
+// WaitForMeteringOperatorDeployment is a helper function that will poll for the @name
+// deployment to report a single integer for both of the UpdatedReplicas and Replicas
```
bentito: actually it's checking for an integer typed value of 1 for both. The comment makes it seem like something else is going on.

test/deployframework/helpers.go
bentito: Go standard for external functions, got it.

test/e2e/main_test.go
bentito: ah

test/testhelpers/helpers.go
bentito: 👍

Makefile
timflannagan1: We're only interested in building the binary for the deploy-metering CLI for this Makefile target. Building the `cmd/deploy-metering/main.go` package will take a fraction of the time it will take to build the entire project root.

test/testhelpers/assert.go
timflannagan1: That makes sense to me, and I have updated the error messages in the `deploy` package before, which in turn would break a test's expected error messages. As a stop-gap, we decided to use a series of regexes instead of strings, which helps in the aforementioned case above, or in the case where the apiserver is returning a different error message than what we expected (namely due to differing kube versions.)

timflannagan1: I can add a `TODO: ...` comment that links back to your concerns so we can rethink how we handle error messages further down the line.

test/deployframework/helpers.go
timflannagan1: This shouldn't be a problem here and updating the implementation for this (and the other functions/methods in the testing suite) to adding and managing go routines isn't something I want to explore in this current PR. I can add a TODO comment so we can look more into adding a timeout function.

bentito: TODO sounds like way to go.

test/testhelpers/assert.go
bentito: yeah mention error types in the TODO -- I think that is the main logic to having error types in a testing regime, so you can assert on them and not error strings.

Makefile
bentito: ah, got it.

timflannagan1:
I'm going to pull out these commits better.
/hold

PR #1141, opened 2020-03-25
CatalogSourceConfig should just be CatalogSource

pruan-rht:


pruan-rht:
@timflannagan1 done...removed the `config` portion from the names

timflannagan1:
Alright, I give up - force merging.

PR #1149, opened 2020-03-25
Bentito go mod

timflannagan1:


pkg/util/workqueue/prometheus/prometheus.go
```diff
@@ -14,95 +14,218 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-// Source: https://github.com/kubernetes/kubernetes/blob/v1.13.3/pkg/util/workqueue/prometheus/prometheus.go
```
timflannagan1, line 17: We shouldn't remove this.

PR #1153, opened 2020-03-27
hack: Capture the deployment and statefulset description before tearing down a test installation.

timflannagan1:
The `hack/run-test-cleanup.sh` script is a way to generate a must-gather geared towards a Metering installation, so this would just expand that scripts capabilities to also dump the descriptions of the `deployment` and `statefulset` objects that get created during installation.

### DIFF
--- a/hack/run-test-cleanup.sh
+++ b/hack/run-test-cleanup.sh
@@ -2,8 +2,8 @@
 
 echo "Performing cleanup"
 echo "Storing pod descriptions and logs at $LOG_DIR"
-echo "Capturing pod descriptions"
 
+echo "Capturing pod descriptions"
 PODS="$(kubectl get pods --no-headers --namespace "$METERING_TEST_NAMESPACE" -o name | cut -d/ -f2)"
 while read -r pod; do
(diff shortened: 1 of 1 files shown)


hack/run-test-cleanup.sh
```diff
@@ -2,8 +2,8 @@
 
 echo "Performing cleanup"
 echo "Storing pod descriptions and logs at $LOG_DIR"
-echo "Capturing pod descriptions"
```
timflannagan1: In the case where any of these commands fail (e.g. there are no Pods in the testing namespace to read from), this script would log that error to stderr. I wonder if we may want to set `set +e` to eliminate the extra noise that this script can produce, especially in the case where we're testing invalid configurations and we expect that errors are going to occur.

bentito: /shrug

bentito: ¯\_(ツ)_/¯

bentito: ¯\\(ツ)/¯

PR #1156, opened 2020-04-01
Bug 1824817: Add a hive metastore database overrides dictionary.

timflannagan1:
By default, hive metastore creates a local, embedded derby database PVC that gets mounted to the Hive metastore Pod. In the case where a user wants to use a database like MySQL or PostgreSQL, they would need to specify this information in the MeteringConfig custom resource, and hive metastore would in turn use this dedicated SQL database for storing metastore data (and not create the derby PVC). In the current state of things, this PVC is still being created, despite a database configuration being provided in the user's MeteringConfig custom resource.

### DIFF
--- a/images/metering-ansible-operator/roles/meteringconfig/defaults/main.yml
+++ b/images/metering-ansible-operator/roles/meteringconfig/defaults/main.yml
@@ -270,13 +270,16 @@ _base_storage_overrides:
         hdfs:
           enabled: false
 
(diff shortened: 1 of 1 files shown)


timflannagan1:
We should look to consolidate these two fields in 4.6 rather than tracking these changes as a BZ.
/hold

timflannagan1:
Let's make this the consolidation PR that builds off the hive metastore database overrides dictionary that #1214 introduces.

PR #1158, opened 2020-04-01
*: Connect to the thanos-querier API instead of the cluster-wide Prometheus instance.

timflannagan1:
In 4.3, the monitoring team updated its architecture to introduce this Thanos Querier component, which provides a way for components like Metering to connect to a single, multi-tenant interface.
By default and when run on Openshift, Metering connects to the cluster-wide, prometheus-k8s service to retrieve the metrics scraped from the cluster. These changes update the metering-ansible-operator Ansible role to connect to the thanos-querier service API instead of the prometheus-k8s service.

In the case of running Metering on a vanilla Kubernetes cluster, nothing has changed - you're still required to configure/provide the Prometheus URL in the `MeteringConfig` custom resource.

### DIFF
--- a/Documentation/configuring-reporting-operator.md
+++ b/Documentation/configuring-reporting-operator.md
@@ -1,18 +1,20 @@
 # Configuring reporting-operator
 
 The reporting-operator is responsible for collecting data from Prometheus, storing the metrics in Presto, running report queries against Presto, and exposing their results via an HTTP API.
+
 Configuring the operator is primarily done within a `MeteringConfig` Custom Resources's `spec.reporting-operator.spec` section.
 
 ## Prometheus connection
 
-Depending on how you installed Metering, the default Prometheus URL varies.
-If you installed for Openshift then the default assumes Prometheus is available at `https://prometheus-k8s.openshift-monitoring.svc:9091/`.
-Otherwise it assumes that your Prometheus service is available at `http://prometheus-k8s.monitoring.svc:9090`.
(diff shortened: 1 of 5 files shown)


Documentation/configuring-reporting-operator.md
```diff
@@ -1,18 +1,20 @@
...
 Configuring the operator is primarily done within a `MeteringConfig` Custom Resources's `spec.reporting-operator.spec` section.
 
 ## Prometheus connection
 
-Depending on how you installed Metering, the default Prometheus URL varies.
-If you installed for Openshift then the default assumes Prometheus is available at `https://prometheus-k8s.openshift-monitoring.svc:9091/`.
-Otherwise it assumes that your Prometheus service is available at `http://prometheus-k8s.monitoring.svc:9090`.
-If you're not on Openshift and aren't using [kube-prometheus][kube-prometheus], then you will need to override the `reporting-operator.config.prometheus.url` configuration option.
+When deploying Metering on an Openshift cluster, the reporting-operator defaults to using the Thanos Querier interface in the openshift-monitoring namespace: `https://thanos-querier.openshift-monitoring.svc:9091/`.
+
```
bentito, line 10: Is there a particular min version OpenShift that this is true for?

bentito: Actually just read the PR summary, so 4.3, that's the min version of Openshift we really support, right? So I guess no need to mention the version?

bentito: master docs doc current release == 4.5, so all set.

PR #1164, opened 2020-04-15
*: Respect cluster-wide Proxy configuration.

timflannagan1:


pkg/aws/client.go
```diff
@@ -33,20 +40,95 @@ type ManifestRetriever interface {
...
+	logger log.FieldLogger
+	s3API  s3iface.S3API
+	bucket string
+	prefix string
 }
 
-func NewManifestRetriever(logger log.FieldLogger, region, bucket, prefix string) ManifestRetriever {
-	awsSession := session.Must(session.NewSession())
-	client := s3.New(awsSession, aws.NewConfig().WithRegion(region))
+func NewManifestRetriever(logger log.FieldLogger, region, bucket, prefix, caBundlePath string) (ManifestRetriever, error) {
```
bentito, line 49:
```suggestion
func proxyAwareManifestRetriever(logger log.FieldLogger, region, bucket, prefix, caBundlePath string) (ManifestRetriever, error) {
```

bentito: Or perhaps that's a bogus function name to suggest, but what's going on with `RetrieveManifests()` vs. `NewManifestRetreiver()` Is the New one getting new manifests or is it a new version of some other function?

timflannagan1: There's this `manifestRetriver` structure type that's contained in the `aws` package. The `NewManifestRetriever` is a function that spits out an initialized `manifestRetriever` object. The `RetrieveManifests(...)` is a _method_ on the `manifestRetriever` type that's responsible for retrieving these billing manifests from the path to the bucket that was provided.

We may want to investigate renaming these functions/methods for clarity. I don't have any great solutions off the top of my head, but renaming the `RetrieveManifests` method is probably the most straightforward.

charts/openshift-metering/templates/reporting-operator/reporting-operator-deployment.yaml
```diff
@@ -447,12 +462,12 @@ spec:
...
-         items:
-         - key: ca-bundle.crt
-           path: ca-bundle.crt
+      - name: reporting-operator-trusted-ca-bundle
+        configMap:
+          name: reporting-operator-trusted-ca-bundle
+          items:
+          - key: ca-bundle.crt
+            path: ca-bundle.crt
 {{- end }}
```
EmilyM1, line 471: What is this change?

pkg/aws/client.go
bentito: okay, sounds maybe out of scope for this PR then.

charts/openshift-metering/templates/reporting-operator/reporting-operator-deployment.yaml
bentito: I wondered too, but I think it just brings indentation to correct?

timflannagan1: It was just misaligned and my editor was complaining about it. I called that out in one of the commits.

PR #1169, opened 2020-04-18
Updating reference to latest kubeprometheus github

amalkurup89:
Motive 

Document review


timflannagan1:
OLM switched to using v1 CRDs which is breaking our upgrade-aws tests

PR #1170, opened 2020-04-20
charts,manifests: Remove the 'minimum' and 'maximum' schema properties.

timflannagan1:


timflannagan1:
The OLM BZ has been resolved, so this is no longer needed to convert v1beta1 -> v1 CRDs under-the-hood.

PR #1171, opened 2020-04-20
charts,manifests: Fix the operators capability console annotation.

timflannagan1:


timflannagan1:
It's expected that the upgrade test will fail due to the OLM crd regression

PR #1194, opened 2020-04-24
[WIP Hive Credentials for DB Secrets]

EmilyM1:


pkg/apis/metering/v1/metering_config.go
```diff
@@ -519,6 +519,7 @@ type HiveDBConfig struct {
 	Password                          string `json:"password,omitempty"`
 	URL                               string `json:"url,omitempty"`
 	Username                          string `json:"username,omitempty"`
+	Secret                            string `json:"secret,omitempty"`
```
timflannagan1, line 522:
```suggestion
	SecretName                            string `json:"secretName,omitempty"`
```

charts/openshift-metering/values.yaml
```diff
@@ -31,6 +31,14 @@ storage:
...
   hive: {}
 
+
+  config:
+    db:
+      url: ""
+      driver: ""
+      username: ""
+      password: ""
+      secret: ""
```
timflannagan1, line 41: No need to redefine this configuration - just add the `secretName` field to the existing `spec.hive.spec.config.db`.

EmilyM1: I found hive.spec.config.db at line 692

PR #1195, opened 2020-04-28
Add gitignore whitelist on test/e2e/

bentito:
Generalized to a whitelist on test/e2e. Subdirectories appear after running e2e tests and it would be good to automatically ignore them.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,3 +34,4 @@ _tmp
 bin/*
 coverage.out
 junit-metering.xml
+/test/e2e/HDFS-OLM-Upgrade/pre-upgrade/logs/
--- a/.gitignore
+++ b/.gitignore
@@ -34,4 +34,7 @@ _tmp
 bin/*
 coverage.out
 junit-metering.xml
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -34,3 +34,7 @@ _tmp
 bin/*
 coverage.out
 junit-metering.xml
+# this causes a whitelist on /test/e2e, add subdirs you want _in_ git
+/test/e2e/*
```
timflannagan1, line 38: I'm not sure this is what we want in the case of adding tests to the e2e suite. Is it possible to instead do something like `/test/e2e/**/*.log`, i.e. only filter out the `*.log` extension for all subdirectories in the test/e2e path?

timflannagan1: (I believe git supports the `/**/` notation)

bentito: I don't have my example any more, but I believe a testdata dir itself was auto-generated by a test, right? The thought with the whitelist was that anything in `test/e2e` would have to have an explicit inclusion to be in git. If it seems like too much, we could go with your suggestion instead to just avoid log files.

PR #1199, opened 2020-04-29
Bug 1829035: Stop loading the presto-prometheus connector.

timflannagan1:


bentito:
It's hard to know if there are any errors of omission of course, but seems like this edits will remove the calls that are causing a problem when built image has a missing Prometheus connector. 

Saving bookmark for this commit as it's going to be the easiest reference for putting all these back in place.

PR #1201, opened 2020-04-30
Bug 1829833: Replace all instances of the depreciated k8s_facts module with k8s_info.

timflannagan1:
When using the 2.9.6 Ansible version, the time it takes for Metering to finish its role execution jumps from 3-5 minutes to ~35-40 minutes. If we increase the verbosity of the ansible logs for the Metering deployment, in 2.9.6, we see that the `k8s_facts` tasks are making a file lookup call orders of magnitude more often, whereas before, that file lookup call was made once or twice.

A more concrete example:
```yaml

...
```

If we pin the Ansible version in both the be less than 2.9.6 and update to the [`k8s_info` module](https://docs.ansible.com/ansible/latest/modules/k8s_info_module.html), the performance issues are resolved and that file lookup call is only made once or twice.


sdodson:
Overrode bugzilla/valid-bug while we wait for the bot changeover to require 4.4.z here https://github.com/openshift/release/pull/8710

PR #1202, opened 2020-04-30
Makefile: Stop passing EXTRA_TEST_FLAGS to the e2e make target.

timflannagan1:
In the case where you only want to run the manual e2e tests and not the upgrade tests, you would need to specify `EXTRA_TEST_FLAGS="-run TestManualMeteringInstall"`.


timflannagan1:
We disable the ci/prow/metering-upgrade-aws test, but not on CI, so force-merging this and updating the CI configuration so we don't continue to waste CI resources.

PR #1203, opened 2020-05-01
Bug 1829072: Fix all of the k8s_event module instances.

timflannagan1:
Previously, all instances of the `k8s_event` module were using an invalid `k8s_event.name` configuration. This produced an error like the following:

```yaml
TASK [meteringconfig : Log Events for validating configurations] ***************
task path: /opt/ansible/roles/meteringconfig/tasks/validate.yml:91
Tuesday 28 April 2020  18:09:29 +0000 (0:00:00.365)       0:00:14.305 ********* 
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: MaxRetryError: HTTPConnectionPool(host='localhost', port=8888): Max retries exceeded with url: /api/v1/namespaces/openshift-metering/events/Validate Configuration (Caused by ProtocolError('Connection aborted.', InvalidURL("URL can't contain control characters. '/api/v1/namespaces/openshift-metering/events/Validate Configuration' (found at least ' ')",)))
fatal: [localhost]: FAILED! => {"changed": false, "msg": "HTTPConnectionPool(host='localhost', port=8888): Max retries exceeded with url: /api/v1/namespaces/openshift-metering/events/Validate Configuration (Caused by ProtocolError('Connection aborted.', InvalidURL(\"URL can't contain control characters. '/api/v1/namespaces/openshift-metering/events/Validate Configuration' 
```

This `k8s_event.name` field directly corresponds to the [traditional `metadata.name`](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names) for a k8s resource, which is how kubernetes refers to an object in a resource URL. This error is produced as the previous configuration(s) had invalid control characters being passed to the `k8s_event` module implementation.

On top of this, I changed the `k8s_event.reportingComponent` to be an optional field as it's not directly applicable to our use case in the metering-ansible-operator, and many of the core events often don't utilize this field.

Other changes that were made:
- Substituting the `source.component` to be the metering-ansible-operator, as "Metering Component" doesn't provide much value and it's difficult to interpret from that source what controller is responsible for firing off this event.
- Reorganizing the `k8s_event.reason`, `k8s_event.message`, and `k8s_event.state` to better reflect the current state of the metering-ansible-operator reconciliation effort.
- In the case of an error, i.e. a user-provided configuration was invalid, only the status field of the `MeteringConfig` custom resource was updated, and no event was fired off warning a user that they might need to take manually intervene.

### DIFF
--- a/images/metering-ansible-operator/roles/meteringconfig/tasks/configure_storage.yml
+++ b/images/metering-ansible-operator/roles/meteringconfig/tasks/configure_storage.yml
@@ -8,6 +8,22 @@
       message: "Configuring Hive storage"
       lastTransitionTime: "{{ now(utc=False).isoformat() + 'Z' }}"
 
+- name: Log event for configuring storage
+  k8s_event:
+    state: present
+    name: metering-operator-configuring-storage
+    namespace: "{{ meta.namespace }}"
+    message: Configuring storage for the metering-ansible-operator
+    reason: Started
+    type: Normal
+    source:
+      component: metering-ansible-operator
+    involvedObject:
+      apiVersion: metering.openshift.io
+      kind: MeteringConfig
+      name: "{{ meta.name }}"
+      namespace: "{{ meta.namespace }}"
+
 #
 # Configure S3 section
 #
@@ -225,21 +241,3 @@
       status: "True"
       message: "Finished configuring Hive storage"
       lastTransitionTime: "{{ now(utc=False).isoformat() + 'Z' }}"
-
-- name: Log Events for Configuring Storage
-  k8s_event:
-    state: present
-    name: Configuring Storage Event
-    namespace: "{{ meta.namespace }}"
-    message: Configure Storage
-    reason: Created
-    reportingComponent: Reporting components
-    type: Normal
-    source:
-      component: Metering components
-    involvedObject:
-      apiVersion: metering.openshift.io
-      kind: MeteringConfig
-      name: "{{ meta.name }}"
-      namespace: "{{ meta.namespace }}"
-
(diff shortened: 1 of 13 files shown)


timflannagan1:
Still need to verify some edge cases
/hold

images/metering-ansible-operator/roles/meteringconfig/tasks/configure_storage.yml
```diff
@@ -8,6 +8,22 @@
...
+- name: Log event for configuring storage
+  k8s_event:
+    state: present
+    name: metering-operator-configuring-storage
+    namespace: "{{ meta.namespace }}"
+    message: Configuring storage for the metering-ansible-operator
+    reason: Started
+    type: Normal
+    source:
+      component: metering-ansible-operator
```
EmilyM1, line 20: ex: only component so 1 line has changed here

timflannagan1:
@EmilyM1 I think you're missing the point of this change - It's possible I need to be more explicit in the PR's description. The problem stems from all of the instances of the `k8s_event.name` being invalid.

EmilyM1:
Oh I know see the name changes as well! I checked the k8 object name link. It does make sense.

images/metering-ansible-operator/roles/meteringconfig/tasks/validate.yml
```diff
@@ -87,21 +87,3 @@
       status: "True"
       message: "Finished the validation process"
       lastTransitionTime: "{{ now(utc=False).isoformat() + 'Z' }}"
-
-- name: Log Events for validating configurations
```
bentito, line 91: How come this section goes away without a replacement event?

timflannagan1: I didn't see much value in firing off an event in the current state of this task file, but I'll re-add it now.

timflannagan1:
We should probably file a BZ for the many-to-many errors when trying to query Prometheus for the pvc-related queries despite its reproducibility. 
/retest

PR #1208, opened 2020-05-04
Bug 1829836: Remove the ansible_failed_task variable from all block/rescue instances.

timflannagan1:
# Overview

With the template caching changes made in 2.9.6, having to evaluate the result of the `ansible_failed_task` expression was greatly slowing down the role and doesn't provide much value for debuggability purposes.

# Changed:

- Removed all `ansible_failed_task` references
- Normalized most of the status updates
- Stored the result of the `meteringconfig_spec` variable locally.

## Storing the result of the `meteringconfig_spec` variable locally

At that point in the role, we've finished overriding any variables that reference this `meteringconfig_spec` variable, which is a giant dictionary containing all of the overridden MeteringConfig default values, and we can safely store this result locally.

Before this change, we copied this variable to a tmp file that the `reconcile_*` task files reference to deploy metering-related resources:

```yaml
- name: Store MeteringConfig spec into values file
  copy: content="{{ meteringconfig_spec }}" dest=/tmp/metering-values.yaml

- include_tasks: "{{ item }}"
  loop:
    - reconcile_metering.yml
    - reconcile_monitoring.yml
    - reconcile_hdfs.yml
    - reconcile_hive.yml
    - reconcile_presto.yml
    - reconcile_reporting_operator.yml
    - reconcile_reporting.yml
```

The problem, however, is that we still use this `meteringconfig_spec` variable under-the-hood to determine whether or not to create a resource:

```yaml
- name: Deploy presto resources
  include_tasks: deploy_resources.yml
  vars:
    resources:
    ...
      - template_file: templates/presto/presto-auth-secrets.yaml
        apis: [ {kind: secret} ]
        prune_label_value: presto-auth-secrets
        create: "{{ meteringconfig_create_presto_auth_secrets }}"
```

Here, the `meteringconfig_create_presto_auth_secrets` variable is defined in the role's defaults/main.yml as the following:

```yaml
meteringconfig_create_presto_auth_secrets: "{{ _presto_spec.config.auth.enabled and _presto_spec.config.auth.createSecret | default(false) }}"
```

Where the value of this `_presto_spec` dictionary is pulled out of the result of `meteringconfig_spec.presto.spec`.

If we instead store the value of this variable locally before reconciling resources, role execution is greatly improved as we no longer need to template these variables (due to lazy evaluation) every time we're looping over this list of resource definitions.

### DIFF
--- a/images/metering-ansible-operator/roles/meteringconfig/tasks/validate_reporting_operator_tls.yml
+++ b/images/metering-ansible-operator/roles/meteringconfig/tasks/validate_reporting_operator_tls.yml
@@ -1,4 +1,7 @@
 ---
+#
+# Note: all of these tasks are wrapped in a block/rescue at the task file call site
+#
 
 #
 # Validate that the user-provided reporting-operator Presto fields were properly configured
@@ -10,17 +13,6 @@
       that:
         - meteringconfig_spec['reporting-operator'].spec.config.presto.tls.caCertificate != ""
       msg: "reporting-operator.spec.config.presto.tls.caCertificate cannot be empty if createSecret: true and secretName != ''"
-  rescue:
-  - include_tasks: update_meteringconfig_status.yml
-    vars:
-      end_play_after_updating_status: true
-      current_conditions:
-        type: "Invalid"
-        status: "True"
-        message: |
-          "Failed task name: {{ ansible_failed_task.name }}"
-          "Failed task message: {{ ansible_failed_result.msg }}"
-        lastTransitionTime: "{{ now(utc=False).isoformat() + 'Z' }}"
   when: meteringconfig_template_presto_tls_secret
 
 - name: Validate that the user-provided Presto client Auth fields are not empty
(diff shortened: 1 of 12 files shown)


images/metering-ansible-operator/roles/meteringconfig/tasks/validate_reporting_operator_tls.yml
```diff
@@ -1,4 +1,7 @@
 ---
+#
+# Note: all of these tasks are wrapped in a block/rescue at the task file call site
```
bentito, line 3: 👍

timflannagan1:
The e2e test flaked on an irrelevant test and passed on everything else.

Also from the e2e runs:
```yaml
Monday 04 May 2020  20:56:31 +0000 (0:00:00.024)       0:02:44.543 ************ 
=============================================================================== 
meteringconfig -------------------------------------------------------- 163.30s
gather_facts ------------------------------------------------------------ 1.11s
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ 
total ----------------------------------------------------------------- 164.41s
Playbook run took 0 days, 0 hours, 2 minutes, 44 seconds
```

After making the changes to how the `meteringconfig_spec` variable was stored, we can see that the role finished execution in 2:44m for the dynamic input test (shown above), and 3:00m for the static input test. This is close to a 50% reduction in the previous time it took to install metering in the testing suite.

PR #1209, opened 2020-05-05
Bug 1829833: Fix performance issues when deploying Metering on Ansible 2.9.6+ versions.

timflannagan1:
# Overview

In 2.9.6, changes were made to how the template caching mechanism works. In previous versions, all variables (i.e. jinja2 expressions) were being cached, instead of the intended behavior which is a single variable gets cached. This was obviously problematic, especially in the case where you're trying to generate a password a couple of times, which meant the first password result gets cached, and any subsequent calls to generating passwords would result in the value of the first, cached password being assigned to the corresponding variable.

## Changes
- Replaced all instances of the depreciated `k8s_facts` module with the `k8s_info` module.
- Stored the result of the `meteringconfig_default_values` variable locally.
- Stored the result of the `meteringconfig_spec` variable locally.
- Removed all of the `ansible_failed_task` block/rescue references.
- Fixed misspelled`creatSecret` error messages.
- Updated the metering-ansible-operator's Dockerfile to stop pinning against the 2.8 Ansible release.

### Replacing `k8s_facts` module with `k8s_info`

In 2.9, the `k8s_facts` module was deprecated in favor of the [`k8s_info` module](https://docs.ansible.com/ansible/latest/modules/k8s_info_module.html). The usage was the same so this essentially just silences the Ansible warning message.

### Storing the result of the `meteringconfig_default_values` variable locally

In the case of Metering, our Ansible role heavily relies on the notion of lazy evaluation - Ansible evaluates any variables at the last possible second. When this change to 2.9.6 was made, Metering saw a significant performance degradation, going from an average of 3-7 minutes to finish role execution, to 45+ minutes.

This was because we were relying on a buggy implementation of the templating caching mechanism. In our current implementation, we pass a variable to the name field of the module. This meant that this module needed to make a filesystem lookup call many, many times for a particular value stored in this meteringconfig_spec variable, which is a large value dictionary and essentially serves as the single source of truth that Metering references.

A more concrete example:
```yaml

...
```

That variable in turn also relies on the value of the `meteringconfig_default_values`, which is a dictionary containing the default helm chart values. Previously, the result of that expression was cached. Now, due to how lazy evaluation works, the `meteringconfig_default_values` needed to be re-evaluated every time it's used, causing the aforementioned performance issues. The workaround is to cache or finalize, for a lack of a better phrase, the resultant of this `meteringconfig_default_values` expression. This would help performance as we're no longer re-evaluating this variable every time we want to template something, and all changes made go through the meteringconfig_spec dictionary so there's no risk of creating this variable early in the role.

### Storing the result of the `meteringconfig` variable locally

At that point in the role, we've finished overriding any variables that reference this `meteringconfig_spec` variable, which is a giant dictionary containing all of the overridden MeteringConfig default values, and we can safely store this result locally.

Before this change, we copied this variable to a tmp file that the `reconcile_*` task files reference to deploy metering-related resources:

```yaml
- name: Store MeteringConfig spec into values file
  copy: content="{{ meteringconfig_spec }}" dest=/tmp/metering-values.yaml

- include_tasks: "{{ item }}"
  loop:
    - reconcile_metering.yml
    - reconcile_monitoring.yml
    - reconcile_hdfs.yml
    - reconcile_hive.yml
    - reconcile_presto.yml
    - reconcile_reporting_operator.yml
    - reconcile_reporting.yml
```

The problem, however, is that we still use this `meteringconfig_spec` variable under-the-hood to determine whether or not to create a resource:

```yaml
- name: Deploy presto resources
  include_tasks: deploy_resources.yml
  vars:
    resources:
    ...
      - template_file: templates/presto/presto-auth-secrets.yaml
        apis: [ {kind: secret} ]
        prune_label_value: presto-auth-secrets
        create: "{{ meteringconfig_create_presto_auth_secrets }}"
```

Here, the `meteringconfig_create_presto_auth_secrets` variable is defined in the role's defaults/main.yml as the following:

```yaml
meteringconfig_create_presto_auth_secrets: "{{ _presto_spec.config.auth.enabled and _presto_spec.config.auth.createSecret | default(false) }}"
```

Where the value of this `_presto_spec` dictionary is pulled out of the result of `meteringconfig_spec.presto.spec`.

If we instead store the value of this variable locally before reconciling resources, role execution is greatly improved as we no longer need to template these variables (due to lazy evaluation) every time we're looping over this list of resource definitions.

### Removing all of the ansible_failed_task block/rescue references.

Like the other items mentioned above, this change has to do with the template caching changes. If a task wrapped in a block/rescue fails (e.g. a configuration failed our validation), then we attempt to update the `MeteringConfig` custom resource with this information. The problem, however, is that most of the failed tasks that get dropped to the rescue portion of the block use lazily evaluated variables, which means that it has to re-evaluate the contents of the task again and significantly slows down the role execution. That said, we use the `ansible_failed_task` solely for debuggability purposes, and that variable provides pretty low value in helping users or developers debug what went wrong with their Metering installation.

As long as we still have access to the `ansible_failed_result`, which is essentially the [registered](https://docs.ansible.com/ansible/latest/user_guide/playbooks_variables.html#registering-variables) version of that failed task (and does not get affected by the template cache changes), we should have all the information we need to be able to produce better, more actionable errors.

### DIFF
--- a/Dockerfile.metering-ansible-operator
+++ b/Dockerfile.metering-ansible-operator
@@ -3,8 +3,7 @@ FROM quay.io/openshift/origin-metering-helm:latest as helm
 # final image needs kubectl, so we copy `oc` from cli image to use as kubectl.
 FROM openshift/origin-cli:latest as cli
 # the base image is the ansible-operator's origin images
-# TODO: stop using the latest tag for base images
-FROM quay.io/openshift/origin-ansible-operator:latest
+FROM quay.io/openshift/origin-ansible-operator:4.4
 
 USER root
 RUN set -x; INSTALL_PKGS="curl bash ca-certificates less which inotify-tools openssl" \
@@ -31,10 +30,7 @@ RUN ln -f -s /tini /usr/bin/tini
 # 3. cryptography is used by the openssl_* modules for TLS/authentication purposes
 # TODO: the ansible-operator base image uses setuptools >= 45.2.0 which needs python 3.5 so this is a temporary fix
 RUN pip install --no-cache-dir --upgrade openshift botocore boto3 cryptography netaddr "setuptools<=45.1.0"
-# In order to use the 'ansible_failed_result' and 'ansible_failed_task' variables in a block/rescue,
-# we need to ensure that the 2.8 version is being used while this is fixed in 2.9 upstream.
-# TODO: revert this change once the issue mentioned above is resolved.
-RUN pip install --upgrade ansible==2.8
+RUN pip install --upgrade ansible~=2.9
 
 ENV HOME /opt/ansible
 ENV HELM_CHART_PATH ${HOME}/charts/openshift-metering
--- a/images/metering-ansible-operator/roles/meteringconfig/tasks/configure_hive_tls.yml
+++ b/images/metering-ansible-operator/roles/meteringconfig/tasks/configure_hive_tls.yml
@@ -4,7 +4,21 @@
 # Validate user-provided Hive TLS configuration when top-level spec.tls.enabled is set to false
 #
 - name: Validate the user-provided Hive TLS configuration
-  include_tasks: validate_hive_tls.yml
+  block:
+  - include_tasks: validate_hive_tls.yml
+  rescue:
+  - include_tasks: update_meteringconfig_status.yml
+    vars:
+      current_conditions:
+        type: "Invalid"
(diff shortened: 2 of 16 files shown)


sdodson:
Upstream bug testing is blocked on other unrelated problems. Overriding bugzilla flags.

PR #1213, opened 2020-05-12
Make metering CI Docker build more like production

bentito:


Dockerfile.metering-ansible-operator
```diff
@@ -1,34 +1,27 @@
 # need the helm-cli from the helm image
 FROM quay.io/openshift/origin-metering-helm:latest as helm
 # final image needs kubectl, so we copy `oc` from cli image to use as kubectl.
-FROM openshift/origin-cli:latest as cli
+FROM registry.svc.ci.openshift.org/ocp/4.6:cli as cli
```
bentito, line 4: As compared to `FROM openshift/ose-metering-helm:latest as helm` in `Dockerfile.rhel`, is hopefully identical image.

Dockerfile.metering-ansible-operator
```diff
@@ -1,34 +1,27 @@
...
 # the base image is the ansible-operator's origin images
-FROM quay.io/openshift/origin-ansible-operator:4.6
+FROM registry.svc.ci.openshift.org/ocp/4.6:ansible-operator
```
bentito, line 6: As compared to: `FROM openshift/ose-ansible-operator:4.6` for base in Dockerfile.rhel

Dockerfile.metering-ansible-operator
```diff
@@ -1,34 +1,27 @@
...
-    && yum install --setopt=skip_missing_names_on_install=False -y \
-        $INSTALL_PKGS  \
-    && yum clean all \
-    && rm -rf /var/cache/yum
-
-ENV TINI_VERSION v0.18.0
-ADD https://github.com/krallin/tini/releases/download/${TINI_VERSION}/tini /tini
-RUN chmod +x /tini
+RUN set -x;
+ARG INSTALL_PKGS="less python2-botocore python2-boto3 openssl tini"
```
bentito, line 10: Move to yum install vs. pip install (pip cannot be used in OSBS/ART build) the following:
`botocore`,`boto3`
Note the following that previously were pip installed, but are now neither yum, nor pip, installed:
- `ansible` does not need to be installed, it's present from the base image.
- `openshift` does not need to be installed. 
- `cryptography` does not to be pip installed, yum install of `openssl` is sufficient.

Dockerfile.metering-ansible-operator
```diff
@@ -1,34 +1,27 @@
...
+
+# the boto core packages will only install if the following repo are added w/out removing base repos
+RUN curl http://base-4-4.ocp.svc > /etc/yum.repos.d/base-4-4.repo
```
bentito, line 13: Using the proxied yum repos via these links promises to make same packages available as on ART build.

We must add this repo to make the boto packages available.

Dockerfile.metering-ansible-operator
```diff
@@ -1,34 +1,27 @@
...
-RUN set -x; INSTALL_PKGS="curl bash ca-certificates less which inotify-tools openssl" \
-    && yum clean all && rm -rf /var/cache/yum/* \
-    && yum install -y https://dl.fedoraproject.org/pub/epel/epel-release-latest-8.noarch.rpm  \
-    && yum install --setopt=skip_missing_names_on_install=False -y \
-        $INSTALL_PKGS  \
-    && yum clean all \
-    && rm -rf /var/cache/yum
-
-ENV TINI_VERSION v0.18.0
-ADD https://github.com/krallin/tini/releases/download/${TINI_VERSION}/tini /tini
```
bentito, line 18: `tini` is provided in the base packages so no longer needs custom munging

Dockerfile.metering-ansible-operator
```diff
@@ -1,34 +1,27 @@
...
 RUN ln -f -s /usr/bin/oc /usr/bin/kubectl
-RUN ln -f -s /tini /usr/bin/tini
 
-# List of python packages:
-# 1. botocore and boto3 are used by the aws-related modules (aws_s3)
-# 2. netaddr is needed to use the ipv4/ipv6 jinja filter
-# 3. cryptography is used by the openssl_* modules for TLS/authentication purposes
-RUN pip install --no-cache-dir --upgrade ansible~=2.9 openshift botocore boto3 cryptography netaddr
+# netaddr is needed to use the ipv4/ipv6 jinja filter
+RUN pip install --no-cache-dir --upgrade netaddr
```
bentito, line 24: Was unable to find from any combo of base packages a replacement that would install and allow to get rid of the `netaddr` pip install.

Dockerfile.metering-ansible-operator
timflannagan1: Are there ways to run this image locally?

Dockerfile.metering-ansible-operator
timflannagan1: I think we still need to rethink deleting these packages and instead pin a particular version for them - that way, we don't run into the problem where the package we get from the base image does not differ between CI and ART again.

Dockerfile.metering-ansible-operator
bparees: are you just setting this for local dev/building?  during CI builds this value should be getting replaced (if it's not, we should look into why not)

Dockerfile.metering-ansible-operator
bparees: same here: are you just setting this for local dev/building? during CI builds this value should be getting replaced (if it's not, we should look into why not)

Dockerfile.metering-ansible-operator
bparees: there's no existing base image you can use that already has this repo?  can we change the ansible-operator image you're building on top of to properly include it?

Dockerfile.metering-ansible-operator
bparees: i assume the OSBS builds aren't doing a pip install (it would fail since it has to reach out remotely), so how are the osbs builds handling this?

Dockerfile.metering-ansible-operator
bentito: Are you referring to the URL host portion: `registry.svc.ci.openshift.org`? Either way, no this is as intended for CI builds, and currently it's working, so if it's working by taking advantage of a bug, let me know.

Dockerfile.metering-ansible-operator
bparees: so this image is being substituted by this logic:

https://github.com/openshift/release/blob/master/ci-operator/config/kube-reporting/metering-operator/kube-reporting-metering-operator-master.yaml#L21

https://github.com/openshift/release/blob/master/ci-operator/config/kube-reporting/metering-operator/kube-reporting-metering-operator-master.yaml#L2-L6

so changing this FROM line doesn't affect what CI is going to do.  It may make it clearer what image is being used (which is fine) but i wouldn't expect it to change the behavior/content.

(last FROM statement in the dockerfile is substituted)

Dockerfile.metering-ansible-operator
bentito: For `ansible` we're pinning it presumably as the base image has semver specification and that's where it's coming from: `FROM registry.svc.ci.openshift.org/ocp/4.6:ansible-operator`.

For `openshift` it literally does not seem to be needed, as not installing it does not seem to harm anything. I think same logic as above for anisble might apply, because if some other package is grabbing it as a dependency, we only have two sources and they both have version explicitly specified (4.7:ansible-operator base image repos, or base-4.4.ocp.svc proxy to repos).

For `cryptography`, Ansible seems to actually need the openssl CLI client to create a random seed. Installing openssl takes care of this need and the python module does not seem needed. We could pin the openssl package to a point in time?

Dockerfile.metering-ansible-operator
bparees: this will "work" but it would probably be better to do the substitution in the release config so you don't have to change the tag here every time you branch.

example:
https://github.com/openshift/release/blob/master/ci-operator/config/openshift/origin/openshift-origin-master.yaml#L50

ensures this gets substituted:
https://github.com/openshift/origin/blob/master/images/hyperkube/Dockerfile.rhel#L1

with the image pointed to by the "bin" tag within the CI process.

(well it should, actually it's broken since the substitution is looking for 1.12 and the dockerfile contains 1.13)

Dockerfile.metering-ansible-operator
bentito: @timflannagan1 well I'm sure you can grab the artifact and run that, but as far as building no, it's the major slowness of this work, the proxies are only reachable from the CI build.

@bparees We could switch off of ansible-operator as a base, there might be a lot to additionally do at that point. There might be a better ansible-operator image, like ansible-operator-aws, or something, but how to inspect the available images? How would we change  ansible-operator to give us what we want, would we fork ansible-operator and then maintain that fork?

Dockerfile.metering-ansible-operator
bentito: Ah, right @timflannagan1 mentioned that substitution early on and I forgot. But how does that work? I know from experience that is the URL is wrong (registry.svc.ci.openshift.org/ocp/4.6) the build will fail.

Dockerfile.metering-ansible-operator
bentito: OSBS: `yum install python-netaddr` but I have so far been unable to locate a public version of that package that is installable with all non-conflicting dependencies.

Dockerfile.metering-ansible-operator
bparees: actually i think the best way to do this is to do an intermediary image, again as part of your CI release config, see:

https://github.com/openshift/release/blob/master/ci-operator/config/openshift-priv/images/openshift-priv-images-release-4.4.yaml#L39-L45

as an example.

then use the tag that step produces, as your base image in the job config/substitution logic.

that way your dockerfile doesn't need to contain CI specific logic and you can more easily collapse on a single dockerfile for CI and OSBS

Dockerfile.metering-ansible-operator
bparees: so there are two substitutions that take place:

1) the last FROM in your dockerfile gets substituted by whatever this `from` references/resolves to:
https://github.com/openshift/release/blob/master/ci-operator/config/kube-reporting/metering-operator/kube-reporting-metering-operator-master.yaml#L21

2) earlier FROMs get substituted by this, if defined:
https://github.com/openshift/release/blob/master/ci-operator/config/openshift/origin/openshift-origin-master.yaml#L50

where in that example the "bin" tag (whatever pullspec it resolves to) is used to substitute anywhere in your dockerfile that `registry.svc.ci.openshift.org/ocp/builder:golang-1.12` appears.  you can reference any tag (one defined by your release config, or one that you know will exist because some other component builds it).

bparees: @bentito you might find this a useful background: https://docs.google.com/presentation/d/1IVxsmB66fxRSo6q_C09DLkm7efth00puM7BCtjJJ3Po/edit#slide=id.g6bd699cf61_2_156

bparees: full presentation here https://bluejeans.com/playback/s/ONXUWzK42mTUbtWR8IUbi8FEdf2Su0vOXdonC9AnSNAZFhppbZPxj1vdrynwigH0

Dockerfile.metering-ansible-operator
bparees: you should be able to get it mirrored into the CI rpm mirror

bentito: Maybe, but it's not as simple as asking for another package, idt. With the current base and curl'd in base-4.4 the following happens on trying to `yum install python-netaddr` (which is available):
```  
Problem: conflicting requests
  - package python-netaddr-0.7.5-7.el7.noarch requires /usr/bin/python, but none of the providers can be installed
  - package python-netaddr-0.7.5-9.el7.noarch requires /usr/bin/python, but none of the providers can be installed
  - package python-2.7.5-16.el7.x86_64 is filtered out by modular filtering
  - package python-2.7.5-38.el7_2.x86_64 is filtered out by modular filtering
  - package python-2.7.5-18.el7_1.1.x86_64 is filtered out by modular filtering
  - package python-2.7.5-34.el7.x86_64 is filtered out by modular filtering
  - package python-2.7.5-39.el7_2.x86_64 is filtered out by modular filtering
  - package python-2.7.5-48.el7.x86_64 is filtered out by modular filtering
  - package python-2.7.5-58.el7.x86_64 is filtered out by modular filtering
  - package python-2.7.5-68.el7.x86_64 is filtered out by modular filtering
  - package python-2.7.5-69.el7_5.x86_64 is filtered out by modular filtering
  - package python-2.7.5-76.el7.x86_64 is filtered out by modular filtering
  - package python-2.7.5-77.el7_6.x86_64 is filtered out by modular filtering
  - package python-2.7.5-80.el7_6.x86_64 is filtered out by modular filtering
  - package python-2.7.5-79.el7_6.x86_64 is filtered out by modular filtering
  - package python-2.7.5-86.el7.x86_64 is filtered out by modular filtering
  - package python-2.7.5-88.el7.x86_64 is filtered out by modular filtering 
```
I tried all available permutations I think. I wasn't able to tell what was causing the modular filtering or how to avoid/turn it off.

bparees: ok.  ART might be able to help you sort that out but maybe as a followup.

Dockerfile.metering-ansible-operator
timflannagan1: We're already substituting these references to use the corresponding ocp imagestream: https://github.com/openshift/release/blob/master/ci-operator/config/kube-reporting/metering-operator/kube-reporting-metering-operator-master.yaml#L2-L9, so this would be nice for local dev purposes.

Dockerfile.metering-ansible-operator
bentito: Actually since .../ocp/4.4:base tag exists, committed a change that makes it a layer and copies from it, curl'd repo is eliminated.

Dockerfile.metering-ansible-operator
bentito: yep, now adding a layer for the existing (and needed for the boto packages) 4.4/base and COPYing from it.

Dockerfile.metering-ansible-operator
bentito: yep as a follow-on or next time we need to touch the Dockerfile

Dockerfile.metering-ansible-operator
```diff
@@ -1,34 +1,30 @@
 # need the helm-cli from the helm image
 FROM quay.io/openshift/origin-metering-helm:latest as helm
```
timflannagan1, line 2: While we're at it, we may as well change this to use the corresponding registry.svc.ci.openshift.org OCP imagestream.

Dockerfile.metering-ansible-operator
bentito: yeah for local dev, not quite sure the one, but not the other: 
```
Step 2/26 : FROM registry.svc.ci.openshift.org/ocp/4.6:cli as cli
 ---> 072597ea8205
Step 3/26 : FROM registry.svc.ci.openshift.org/ocp/4.4:base as base-4-4
Get https://registry.svc.ci.openshift.org/v2/ocp/4.4/manifests/base: unauthorized: authentication required
```

PR #1215, opened 2020-05-13
hack,charts,manifests: Update the manual OLM metering manifests.

timflannagan1:
This updates our manual olm manifests for deploying Metering. The CatalogSourceConfig has been a depreciated object for a while now, so that's been replaced with using a ConfigMap-based CatalogSource implementation. Note: there's currently no ConfigMap that gets templated, but there is a `./hack/create-upgrade-configmap.sh` script that achieves the same functionality that can be renamed/generalized in down the line.


hack/olm-install.sh
```diff
@@ -12,37 +12,41 @@ if [ "$METERING_NAMESPACE" != "openshift-metering" ]; then
...
         > "$TMPDIR/metering.subscription.yaml"
 
-        export OLM_MANIFESTS_DIR="$TMPDIR"
+    export OLM_MANIFESTS_DIR="$TMPDIR"
+    export NAMESPACE=$METERING_NAMESPACE
 fi
 
-msg "Installing Metering Catalog Source Config"
+msg "Creating the Metering ConfigMap"
+"$ROOT_DIR/hack/create-upgrade-configmap.sh" \
```
pweil-, line 36: trailing `\` not necessary here?

pkg/deploy/uninstall.go
```diff
@@ -74,7 +74,6 @@ func (deploy *Deployer) uninstallMeteringCSV() error {
 	// and hope that the user is re-running the olm-uninstall command.
 	sub, err := deploy.olmV1Alpha1Client.Subscriptions(deploy.config.Namespace).Get(deploy.config.SubscriptionName, metav1.GetOptions{})
 	if apierrors.IsNotFound(err) {
-		deploy.logger.Warnf("The metering subscription does not exist")
```
bentito, line 77: Why are we still bothering with the conditional?

pweil-: Not found is usually an acceptable condition during an uninstall since the intent is to remove the object.

bentito: Aj, so the `IsNotFound` is trapping out the uninstall case. Should add a comment there to that effect then maybe.

pweil-:
```go
// in the case where the subscription resource does not already exist, exit early
// and hope that the user is re-running the olm-uninstall command.
```

maybe not the clearest, but I think that's the intent 👍

timflannagan1:
Overriding valid-bz label - this is just for QE testing.

PR #1218, opened 2020-05-14
Bug 1837043: Mount the cluster's certificate authority bundle in the auth-proxy sidecar container.

timflannagan1:
This CA bundle already gets mounted in the reporting-operator container when the cluster-wide proxy is enabled, but not in the sidecar container.


timflannagan1:
Want to reproduce the user's cluster/`MeteringConfig` setup before merging.
/hold

PR #1219, opened 2020-05-18
pkg/deploy: Add the empty openshift.io/node-selector annotation to the metering-operator namespace.

timflannagan1:
In the case where a cluster administrator has configured the `defaultNodeSelector` field for the cluster's `Scheduler` object, which is responsible for determining the placement of Pods in a cluster, there may be conflicts when a user also specifies a `nodeSelector` configuration for the metering-operator's operands in the `MeteringConfig` object. Instead, if we always set the `'openshift.io/node-selector'` annotation to be empty, meaning that we don't care where the Pods in this namespace get scheduled, the `defaultNodeSelector` won't be injected by the Scheduler and the user-provided `nodeSelector` configuration will work as intended.

Note: these changes are only geared towards manually deploying Metering manually (e.g. deploy everything ourselves) and deploying Metering through OLM manually. In order to install Metering through OperatorHub, you would still need to create/label/annotate the namespace yourself as a pre-requisite step to configuring Metering.

### DIFF
--- a/pkg/deploy/install.go
+++ b/pkg/deploy/install.go
@@ -26,9 +26,32 @@ func (deploy *Deployer) installNamespace() error {
 			deploy.logger.Infof("Labeling the %s namespace with '%s=%s'", deploy.config.Namespace, key, val)
 		}
 
+		/*
+			In the case where the platform is set to Openshift (the default value),
+			we need to make a few modifications to the namespace metadata.
+
+			The 'openshift.io/cluster-monitoring' labels tells the cluster-monitoring
+			operator to scrape Prometheus metrics for the installed Metering namespace.
+
+			The 'openshift.io/node-selector' annotation is a way to control where Pods
+			get scheduled in a specific namespace. If this annotation is set to an empty
+			label, that means that Pods for this namespace can be scheduled on any nodes.
+
+			In the case where a cluster administrator has configured a value for the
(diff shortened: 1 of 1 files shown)


pkg/deploy/install.go
```diff
@@ -29,6 +29,10 @@ func (deploy *Deployer) installNamespace() error {
 		if deploy.config.Platform == "openshift" {
 			labels["openshift.io/cluster-monitoring"] = "true"
 			deploy.logger.Infof("Labeling the %s namespace with 'openshift.io/cluster-monitoring=true'", deploy.config.Namespace)
+			namespaceObjectMeta.Annotations = map[string]string{
```
bentito, line 32: I think a comment along the lines of the commit message would be helpful to explain what's going on.

timflannagan1: Updated in the last force-push. I tried to add some context to both of the namespace metadata modifications we make.

bentito: image

timflannagan1:
Let me file a BZ for that upgrade e2e failure
/hold

timflannagan1:
Filed a BZ for the upgrade failure flake. Going to override that test and manually add the valid-bz label as this Go library is only used internally + testing purposes.
/hold cancel

PR #1221, opened 2020-05-21
Bug 1841883: Change aggregate function SUM to AVG

bentito:
Sum was resulting in values that were off by <correctValue> = <reportedValue>/timeprecision


charts/openshift-metering/templates/openshift-reporting/report-queries/persistentvolumeclaim-usage.yaml
```diff
@@ -165,7 +165,7 @@ spec:
       timestamp '{| default .Report.ReportingEnd .Report.Inputs.ReportingEnd | prestoTimestamp |}' AS period_end,
       namespace,
       persistentvolumeclaim,
-      sum(persistentvolumeclaim_usage_bytes)
+      add(persistentvolumeclaim_usage_bytes)
```
timflannagan1, line 168:
```suggestion
      avg(persistentvolumeclaim_usage_bytes)
```

bentito:
See this gist for some notes about the changes in values in the reports due to the move from SUM to AVG in this PR: https://gist.github.com/bentito/f81d1aa47b6311724331e1a312a63e55

charts/openshift-metering/templates/openshift-reporting/report-queries/persistentvolumeclaim-usage.yaml
bentito: fixed

test/e2e/testdata/reports/namespace-persistentvolumeclaim-usage.json
```diff
@@ -1 +1 @@
...
\ No newline at end of file
+[{"namespace":"metering-chancez","period_end":"2019-03-18T16:05:00Z","period_start":"2019-03-18T16:00:00Z","persistentvolumeclaim_usage_bytes":41467084.8},{"namespace":"metering-ci2-openshift-continuous-upgrade-master","period_end":"2019-03-18T16:05:00Z","period_start":"2019-03-18T16:00:00Z","persistentvolumeclaim_usage_bytes":873716394.6666666},{"namespace":"metering-tschuy","period_end":"2019-03-18T16:05:00Z","period_start":"2019-03-18T16:00:00Z","persistentvolumeclaim_usage_bytes":239100450.13333333},{"namespace":"openshift-monitoring","period_end":"2019-03-18T16:05:00Z","period_start":"2019-03-18T16:00:00Z","persistentvolumeclaim_usage_bytes":2256873062.4},{"namespace":"telemeter-tschuy","period_end":"2019-03-18T16:05:00Z","period_start":"2019-03-18T16:00:00Z","persistentvolumeclaim_usage_bytes":266761011.2}]
```
timflannagan1, line 1: Is there a reason we cannot pretty-print this Report json as well?

test/e2e/testdata/reports/persistentvolumeclaim-usage.json
```diff
@@ -1 +1,121 @@
...
\ No newline at end of file
+[
```
timflannagan1, line 1: I took a quick look at the commits, but it's difficult to tell what's been changed vs. what's just been pretty-printed.

test/e2e/testdata/reports/namespace-persistentvolumeclaim-usage.json
bentito: Done in 24017f48

test/e2e/testdata/reports/persistentvolumeclaim-usage.json
bentito: This image should help, it shows the report file which had real value changes. I soft-wrapped the version from master:
image

bentito: image

bentito: image

bentito:
We're thinking we should test this against less static data (per @timflannagan1 ) that data should be available, so when we have a test against it, this can merge.

charts/openshift-metering/templates/openshift-reporting/report-queries/persistentvolumeclaim-usage.yaml
```diff
@@ -205,7 +205,7 @@ spec:
       timestamp '{| default .Report.ReportingStart .Report.Inputs.ReportingStart| prestoTimestamp |}' AS period_start,
       timestamp '{| default .Report.ReportingEnd .Report.Inputs.ReportingEnd | prestoTimestamp |}' AS period_end,
       namespace,
-      sum(persistentvolumeclaim_usage_bytes)
+      avg(persistentvolumeclaim_usage_bytes)
```
timflannagan1, line 208: Just picking this back up now, but I realize that using sum here is probably the most reasonable way to calculate usage per-PVC-per-namespace, but rather the value of the `persistentvolumeclaim_usage_bytes` is likely way off in the way we calculate and expose that column. We likely should be summing the averaged value for an individual PVC here instead of summing all of the metric samples we've gathered between a datetime and timestamp range, which would help explain why we were seeing orders-of-magnitude results when running this ReportQuery custom resource.

timflannagan1:
I'm going to tackle fixing these quer(ies) in a separate PR.

/close

PR #1222, opened 2020-05-22
Bug 1829446: Disable the presto-prometheus by default for OCP installations.

timflannagan1:


images/metering-ansible-operator/roles/meteringconfig/defaults/main.yml
```diff
@@ -394,7 +394,9 @@ _ocp_enabled_overrides:
       config:
         connectors:
           prometheus:
-            enabled: true
+            # Re-enable by default for OCP installs once we resolved
+            # https://bugzilla.redhat.com/show_bug.cgi?id=1829446
+            enabled: false
```
bentito, line 399: :-(

bentito: but we'll get there. So many moving parts.

PR #1224, opened 2020-05-26
images: Finalize the meteringconfig_spec_overrides dictionary before running any top-level tasks.

timflannagan1:
This is another example of a static dictionary that we reference throughout the role where storing its result into a fact will shave off some overall role execution time, as Ansible no longer has to evaluate it every time its referenced.


timflannagan1:
Let's wait for the release-4.6 branch to be opened up before merging.
/hold

PR #1225, opened 2020-05-26
[release-4.4] Bug 1847111: Add support for upgrading between Metering CSV versions.

timflannagan1:
# Overview

In the 4.5 release cycle, support was added to support upgrading between Metering CSV versions. We were asked to backport upgrade support to any release that made sense, which in this case would be 4.3/4.4 as 4.2 is marked as EOL once 4.5 GA is released.

# Changes

## charts,manifests[meteringoperator.v4.4.0.clusterserviceversion.yaml]

The [`olm.skipRange`](https://docs.openshift.com/container-platform/4.4/operators/understanding_olm/olm-understanding-olm.html#olm-upgrades-replacing-multiple_olm-understanding-olm) CSV annotation adds the ability to upgrade between Metering CSV versions in a single `package.yaml`, which ART populates.

## charts,manifests[art.yaml]

This file governs the substitution rules that ART is configured to read from. In this case, we're substituting the second x.y.z version from <4.4.0 reference, with the latest 4.4.z z-stream release.

### DIFF
--- a/charts/metering-ansible-operator/templates/olm/art.yaml
+++ b/charts/metering-ansible-operator/templates/olm/art.yaml
@@ -9,6 +9,8 @@ updates:
     # replace spec.version value
     - search: 'version: "{{ .Values.olm.csv.version }}"'
       replace: 'version: {FULL_VER}'
+    - search: 'olm.skipRange: ">=4.2.0 <{MAJOR}.{MINOR}.0"'
+      replace: 'olm.skipRange: ">=4.2.0 <{FULL_VER}"'
   - file: "metering.package.yaml"
     update_list:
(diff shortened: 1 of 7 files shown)


timflannagan1:
Manually adding the lgtm label as the bot didn't pick up the `/lgtm` comment when it ran out of github tokens yesterday.

timflannagan1:
It looks like the release-4.4 branch is complaining about a missing dependency in our build root Dockerfile for CI, so pushing the workaround we have in master.

timflannagan1:
Dropped the Dockerfile.src updates and instead moved that to its own PR as changes to that Dockerfile don't propagate to CI builds.

PR #1227, opened 2020-06-01
charts,manifests: Fix MeteringConfig CRD apiserver validation errors.

timflannagan1:
The MeteringConfig CRD was failing numerous apiserver validation checks after we installed the CRD ourselves, so these changes are necessary for translating the CRD from v1beta1 to v1 down the line.


timflannagan1:
Closing in favor of #1253

PR #1229, opened 2020-06-01
Bug 1842978: Migrate to obtaining the service-serving CA bundle from a ConfigMap.

timflannagan1:
## Overview

In 4.1, the `/var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt` field was depreciated, and in 4.5 it was removed entirely. Now, we need to migrate to reconciling an empty ConfigMap with the `service.beta.openshift.io/inject-cabundle="true"` annotation, which results in the service CA certificate bundle being automatically injected through one of the cluster's controllers. We can then mount that certificate into any of the Metering operand Pods that require access to that CA bundle.

### Changes

- charts/openshift-metering/templates/metering/metering-service-ca-crt.yaml adds an empty ConfigMap with the `service.beta.openshift.io/inject-cabundle="true"` annotation
- images/metering-ansible-operator/roles/meteringconfig/tasks/reconcile_metering.yml adds another reconciliation item to the reconcile_metering task file, and we always create this file.
- charts/openshift-metering/templates/monitoring/reporting-operator-service-monitor.yaml replaces the static/hardcoded /var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt path with the ConfigMap that the Prometheus operator manages.
- charts/openshift-metering/*: updates the operand deployments/statefulsets/configmap scripts to mount the ConfigMap data into the applicable containers, which can then be consumed as a `volumeMount`.

### DIFF
--- a/charts/openshift-metering/templates/hadoop/hadoop-scripts.yaml
+++ b/charts/openshift-metering/templates/hadoop/hadoop-scripts.yaml
@@ -43,9 +43,9 @@ data:
       importCert /s3-compatible-ca/ca-bundle.crt changeit $JAVA_HOME/lib/security/cacerts
     fi
     # always add the openshift service-ca.crt if it exists
-    if [ -a /var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt ]; then
-      echo "Adding /var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt to $JAVA_HOME/lib/security/cacerts"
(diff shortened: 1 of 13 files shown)


timflannagan1:
Need to create BZs for both 4.5 and 4.6.
/hold

timflannagan1:
This passed the ci/prow/metering-e2e-aws tests that we care about so I'm going to force-merge so we can get 4.5 builds back on track quicker.

PR #1238, opened 2020-06-04
Documentation/dev: Add documentation for the deploy-metering CLI utility.

timflannagan1:
This utility has been in use for a while but there was no available documentation that described the various sub-commands and their options.

### DIFF
--- a/Documentation/dev/deploy-metering.md
+++ b/Documentation/dev/deploy-metering.md
@@ -1,0 +1,165 @@
+# Overview
+
+The `deploy-metering` utility aims to emulate the install/uninstall logic that's currently done in the `hack/install.sh` and `hack/uninstall.sh` shell scripts.
+
+When interacting with the `deploy-metering` binary, you have the choice of using the command flags available, or their respective environment variable.
+
+In the case where both the flag, and the environment variable are specified, the [flag value takes precedence](#example-of-precedence) over the value stored in the environment variable.
+
+## Building the binary
+
+```bash
+make bin/deploy-metering
+```
+
+### Prerequisites
+
+With the addition of the deploy-metering CLI, we can assign a default (typically an empty value) value for most of the configuration options.
+
+This means the only required flags/environment variables that needs to be set, are the namespace in which the Metering stack will be managed and the path to the `MeteringConfig` custom resource file.
+
+Before deploying metering, ensure that either of the following options are set:
+
+```bash
+export METERING_NAMESPACE="metering"
+export METERING_CR_FILE=<path to custom MeteringConfig custom resource>
+```
+
+Alternatively, run the CLI using the corresponding flags:
+
+```bash
+./bin/deploy-metering install --namespace metering --meteringconfig <path to custom MeteringConfig custom resource>
+```
+
+It's important to note that when the path to the `MeteringConfig` custom resource is unspecified, the package uses the default `MeteringConfig` manifest resource, which contains an invalid storage specification.
+
+If you do not set that environment variable or flag, the installation will succeed, but it will fail the validation check in the metering-ansible-operator Ansible role due to an invalid user-provided storage configuration.
+
+### General Configuration Options
+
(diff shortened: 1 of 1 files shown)


Documentation/dev/deploy-metering.md
```diff
@@ -0,0 +1,165 @@
...
+
+```bash
+./bin/deploy-metering install --namespace metering --meteringconfig <path to custom MeteringConfig custom resource>
+```
+
+It's important to note that when the `MeteringConfig` custom resource is unspecified, the package uses the default `MeteringConfig` manifest resource, which contains an invalid storage specification.
+
+If you do not set that environment variable or flag, the installation will succeed, but it will fail the validation check in the metering-ansible-operator Role due to an invalid storage specification.
+
+### General Configuration Options
```
EmilyM1, line 38: Could add that KUBECONFIG is also an env var that needs to be set. Not sure how much basic Kubernetes understanding we are catering to here.

Documentation/dev/deploy-metering.md
```diff
@@ -0,0 +1,165 @@
...
+```
+
+#### Override the default images install
+
+```bash
+export METERING_NAMESPACE="metering"
+export METERING_OPERATOR_IMAGE_REPO="quay.io/tflannag/origin-metering-ansible-operator"
+export METERING_OPERATOR_IMAGE_TAG="latest"
+./bin/deploy-metering install
+```
```
EmilyM1, line 105: Let's make this more generic throughout.
export METERING_OPERATOR_IMAGE_REPO="quay.io/REPLACE-WITH-USERNAME/origin-metering-ansible-operator"
export METERING_OPERATOR_IMAGE_TAG="REPLACE-WITH-IMAGE-TAG"
or something less literal

Documentation/dev/deploy-metering.md
pruan-rht: @EmilyM1 that's a `oc` thing, so it doesn't apply here.

Documentation/dev/deploy-metering.md
```diff
@@ -0,0 +1,165 @@
...
+### Prerequisites
+
+With the addition of the deploy-metering CLI, we can assign a default (typically an empty value) value for most of the configuration options.
+
+This means the only required flags/environment variables that needs to be set, are the namespace in which the Metering stack will be managed and the path to the `MeteringConfig` custom resource file.
+
+Before deploying metering, ensure that either of the following options are set:
+
+```bash
+export METERING_NAMESPACE="metering"
```
pruan-rht, line 24: @timflannagan1 OLM install via OperatorHub defaults to `openshift-metering`.  Perhaps change the value to match

Documentation/dev/deploy-metering.md
timflannagan1: @pruan-rht I think this may be a valid point, but let's hold off for now on listing the more granule requirements as pre-requisites until we flesh this initial documentation more.

Documentation/dev/deploy-metering.md
timflannagan1: My intent was to make it platform agnostic - once we start adding a better OLM installation path to this utility (right now it only installs via the redhat-operators catalogsource), we may want to start documenting a more restrictive namespace recommendation, at least for Openshift.

Documentation/dev/deploy-metering.md
timflannagan1: This is also why I threw this in the dev documentation for now, until there's a better avenue with interacting with this tool to expose more generally.

Documentation/dev/deploy-metering.md
```diff
@@ -0,0 +1,165 @@
...
+```
+
+#### Specify manifest location explicitly install
+
+```bash
+$ pwd
+/home/tflannag/go/src/github.com/operator-framework/operator-metering
+$ cd ../
+$ pwd
+/home/tflannag/go/src/github.com/operator-framework
```
pruan-rht, line 114: @timflannagan1 looks like an outdated output...it's pointing to the old github name

PR #1244, opened 2020-06-10
test/e2e: Replace the MeteringConfig Golang objects with decoded YAML manifests.

timflannagan1:
This would add a set of MeteringConfig custom resource YAML manifests in the test/e2e/testdata directory. We can decode those manifests into Go object which is what the rest of the e2e deployframework suite currently expects. This way, we reduce the size of the main_test.go significantly, and it's much easier to follow the configuration listed in the YAML manifest, as compared to a giant Go object.

### DIFF
--- a/test/e2e/main_test.go
+++ b/test/e2e/main_test.go
@@ -7,18 +7,15 @@ import (
 	"path/filepath"
 	"strings"
 	"testing"
-	"time"
 
 	"github.com/sirupsen/logrus"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
-	v1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/api/resource"
-	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
+	metering "github.com/kube-reporting/metering-operator/pkg/apis/metering/v1"
+	"k8s.io/apimachinery/pkg/util/yaml"
 	_ "k8s.io/client-go/plugin/pkg/client/auth/oidc"
 
-	metering "github.com/kube-reporting/metering-operator/pkg/apis/metering/v1"
 	"github.com/kube-reporting/metering-operator/test/deployframework"
 	"github.com/kube-reporting/metering-operator/test/reportingframework"
 	"github.com/kube-reporting/metering-operator/test/testhelpers"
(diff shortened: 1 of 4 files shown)


test/e2e/main_test.go
```diff
@@ -98,23 +96,24 @@ type InstallTestCase struct {
...
+		Name                       string
+		MeteringOperatorImageRepo  string
+		MeteringOperatorImageTag   string
+		Skip                       bool
+		PurgeReports               bool
+		PurgeReportDataSources     bool
+		ExpectInstallErr           bool
+		ExpectInstallErrMsg        []string
+		InstallSubTest             InstallTestCase
+		MeteringConfigManifestPath string
```
bentito, line 108:
```suggestion
		MeteringConfigManifestFilename string
```

test/e2e/main_test.go
```diff
@@ -221,7 +169,7 @@ func testManualOLMUpgradeInstall(
 	purgeReports,
 	purgeReportDataSources bool,
 	testInstallFunction InstallTestCase,
-	testMeteringConfigSpec metering.MeteringConfigSpec,
+	testMeteringConfigManifestPath string,
```
bentito, line 172: Does this need the `test` prefix? 

Also, this func name should be capitalized (not changing in this PR, but might as well fix)

timflannagan1: This is technically a test helper function to the top-level `TestManualMeteringInstall` function, so we don't want the function to be exported here. Something I can do in another commit or a follow-up is organized these functions better so we have a better idea of what's being run and why.

bentito: 👍

test/e2e/main_test.go
```diff
@@ -98,23 +96,24 @@ type InstallTestCase struct {
...
+		MeteringOperatorImageRepo      string
+		MeteringOperatorImageTag       string
+		Skip                           bool
+		PurgeReports                   bool
+		PurgeReportDataSources         bool
+		ExpectInstallErr               bool
+		ExpectInstallErrMsg            []string
+		InstallSubTest                 InstallTestCase
+		MeteringConfigManifestFilename string
 	}{
```
EmilyM1, line 109: Nit: indents on the values

test/e2e/main_test.go
```diff
@@ -298,14 +255,14 @@ func testManualOLMUpgradeInstall(
...
-		MeteringConfigSpec        metering.MeteringConfigSpec
+		Name                           string
+		MeteringOperatorImageRepo      string
+		MeteringOperatorImageTag       string
+		Skip                           bool
+		ExpectInstallErr               bool
+		ExpectInstallErrMsg            []string
+		InstallSubTest                 InstallTestCase
+		MeteringConfigManifestFilename string
 	}{
```
EmilyM1, line 266: same

test/e2e/main_test.go
timflannagan1: All of that is automatically controlled by `gofmt`. It's part of the `make verify` check as well.

PR #1245, opened 2020-06-11
Documentation,images: Generalize and update the configuring_hive_metastore task file.

timflannagan1:
We need to avoid using `set_fact` on the `meteringconfig_spec` dictionary before we're finished processing any non-reconciliation tasks. If we added an additional task file that we include after the `configure_hive_metastore.yml` task file, override a variable in the `meteringconfig_spec` dictionary, we won't see that propagate to the rest of the role as we ran `set_fact` on this dictionary in a previous task file.

The main change here is consolidating the need to set `hive.spec.metastore.storage.create` to false when a non-default database configuration has been specified. Also, we add MeteringConfig status updates and some light validation centered around asserting that the user didn't provide a username/password/secretName configuration in the `MeteringConfig` custom resource. This is important to check as the username and password fields will override what gets pulled out in the `secretName` object, as the `meteringconfig_spec_overrides` (e.g. what the user has provided as a configuration) has precedence over the hive metastore overrides dictionary.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 - [#1214](https://github.com/kube-reporting/metering-operator/pull/1214) Add initial support for configuring the Hive Metastore database to reference a secret containing the base64 encrypted username and password credentials.
 - [#1224](https://github.com/kube-reporting/metering-operator/pull/1224) Improve performance of the metering-ansible-operator by "finalizing" the `meteringconfig_spec_overrides` dictionary.
 - [#1226](https://github.com/kube-reporting/metering-operator/pull/1226) Re-generate assets and YAML manifests to point to the 4.6 images.
+- [#1245](https://github.com/kube-reporting/metering-operator/pull/1245) Consolidate Hive Metastore configurable fields. You no longer need to specify `spec.hive.spec.metastore.storage.create: false` when using a non-default database for Hive Metastore.
 
 ### Bug Fixes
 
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 - [#1214](https://github.com/kube-reporting/metering-operator/pull/1214) Add initial support for configuring the Hive Metastore database to reference a secret containing the base64 encrypted username and password credentials.
 - [#1224](https://github.com/kube-reporting/metering-operator/pull/1224) Improve performance of the metering-ansible-operator by "finalizing" the `meteringconfig_spec_overrides` dictionary.
 - [#1226](https://github.com/kube-reporting/metering-operator/pull/1226) Re-generate assets and YAML manifests to point to the 4.6 images.
+- [#1245](https://github.com/kube-reporting/metering-operator/pull/1245) Consolidate Hive Metastore configurable fields. You no longer need to specify `spec.hive.spec.metastore.storage.create: false` when using a non-default database for Hive Metastore.
```
bentito, line 8: image

timflannagan1:
Let's get the e2e update merged before this just in case
/hold

EmilyM1:
Testing in cluster
/hold

EmilyM1:
Config error in the pod logs.

EmilyM1:
/hold 
Will retest in cluster

EmilyM1:
Verified mysql, reports and data population in select tables

PR #1247, opened 2020-06-11
[release-4.3] Add support for upgrading between Metering CSV versions.

timflannagan1:
In the 4.5 release cycle, support was added to support upgrading between Metering CSV versions. We were asked to backport upgrade support to any release that made sense, which in this case would be 4.3/4.4 as 4.2 is marked as EOL once 4.5 GA is released.

The [`olm.skipRange`](https://docs.openshift.com/container-platform/4.4/operators/understanding_olm/olm-understanding-olm.html#olm-upgrades-replacing-multiple_olm-understanding-olm) CSV annotation adds the ability to upgrade between Metering CSV versions in a single `package.yaml`, which ART populates.

This file governs the substitution rules that ART is configured to read from. In this case, we're substituting the second x.y.z version from <4.3.0 reference, with the latest 4.4.z z-stream release.

### DIFF
--- a/charts/metering-ansible-operator/templates/olm/art.yaml
+++ b/charts/metering-ansible-operator/templates/olm/art.yaml
@@ -9,6 +9,8 @@ updates:
     # replace spec.version value
     - search: 'version: "{{ .Values.olm.csv.version }}"'
       replace: 'version: {FULL_VER}'
+    - search: 'olm.skipRange: ">=4.2.0 <{MAJOR}.{MINOR}.0"'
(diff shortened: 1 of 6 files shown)


timflannagan1:
Need to get clarification as to whether we need to backport to a release that's currently in the maintenance phase. 
/hold

timflannagan1:
Unfortunately, this release branch appears to be a relatively poor place. We migrated this project's repository from the operator-framework organization to this current one ~4.4 timeline, which results in some import errors that are referencing the old GOPATH location.

Going to try and get images squared away in another PR first, then the import errors, before we can get CI green again for 4.3.

PR #1251, opened 2020-06-11
[WIP] shared storage to test for CI 

EmilyM1:


charts/openshift-metering/templates/reporting-operator/reporting-operator-deployment.yaml
```diff
@@ -390,6 +390,8 @@ spec:
           name: reporting-operator-trusted-ca-bundle
           readOnly: true
 {{- end }}
+        - mountPath: /var/nfs
+          name: nfs-storage
```
timflannagan1, line 394: We don't need to mount anything in either of the containers in the reporting-operator Pod.

test/e2e/main_test.go
```diff
@@ -409,3 +409,6 @@ func testManualMeteringInstall(
 	err = deployerCtx.Teardown(deployerCtx.Deployer.Uninstall)
 	assert.NoError(t, err, "capturing logs and uninstalling metering should produce no error")
 }
+
+func TestSharedStorage(t *testing.T) {
+	testInstallConfigs := []struct {
```
timflannagan1, line 414: We would most likely add this type of test as an additional test table index in the top-level `TestManualMeteringInstall` test function.

PR #1253, opened 2020-06-14
Bump the metering-related CRDs to use apiextensions.k8s.io/v1.

timflannagan1:
This updated all of the CRDs from apiextensions.k8s.io v1beta1 to v1.

A couple of things to note:
- CRDs went GA in 1.16 and the v1beta1 version was marked as deprecated
- Need to determine the type of some of the more dynamic fields (`x-kubernetes-preserve-unknown-fields: true` should be sufficient for the time being)
- Need to determine whether or not we should still specify the `v1alpha1` version


go.mod
```diff
@@ -13,7 +13,6 @@ require (
 	github.com/go-chi/chi v3.3.2+incompatible
 	github.com/gogo/protobuf v1.3.1 // indirect
 	github.com/golang/mock v1.4.3
-	github.com/google/gofuzz v1.1.0 // indirect
```
bentito, line 16: These are all just results of running `go mod vendor`?

timflannagan1:
I'm going to bump the k8s dependencies to 1.18 in a separate PR as there's a ton of changes made to client-go and we already made some changes to the module dependencies in this PR too.
/hold

PR #1255, opened 2020-06-15
Bug 1847151: Fix the olm.skipRange CSV annotation being rendered incorrectly.

timflannagan1:
In the art.yaml, which defines the substitution rules for ART, we search for this olm.skipRange annotation in the Metering CSV file:

```yaml
    - search: 'olm.skipRange: ">=4.3.0 <{MAJOR}.{MINOR}.0"'
```

What's problematic here is that we're searching for strings that start with double-quotes, but helm renders many of the chart's CSV annotations as single quotes, instead of double-quotes, so the search for substitution rule fails and the end result is ART cannot properly substitute the 4.6 version.


timflannagan1:
Test failed due to many-to-many PVC-related metrics bug that's already being tracked in BZ.

PR #1256, opened 2020-06-15
[release-4.5] Bug 1847154: Fix the olm.skipRange CSV annotation being rendered incorrectly.

timflannagan1:
In the art.yaml, which defines the substitution rules for ART, we search for this olm.skipRange annotation in the Metering CSV file:

```yaml
    - search: 'olm.skipRange: ">=4.2.0 <{MAJOR}.{MINOR}.0"'
```

What's problematic here is that we're searching for strings that start with double-quotes, but helm renders all of the chart's annotations as single quotes, instead of quote, so the search for that string fails and the end result is ART cannot properly substitute the 4.5 version.

### DIFF
--- a/charts/metering-ansible-operator/templates/olm/clusterserviceversion.yaml
+++ b/charts/metering-ansible-operator/templates/olm/clusterserviceversion.yaml
@@ -5,6 +5,7 @@ metadata:
   namespace: placeholder
 {{- if .Values.olm.csv.annotations }}
   annotations:
+    olm.skipRange: {{ .Values.olm.csv.skipRange | quote }}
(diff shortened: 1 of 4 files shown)


timflannagan1:
Verified fixing this ends up correctly substituted in distgit:

```git
+  name: metering-operator.4.6.0-202006160202
   namespace: placeholder
   annotations:
+    olm.skipRange: ">=4.3.0 <4.6.0-202006160202"
```

timflannagan1:
The cherry-pick bot is broken temporarily for Metering (due to the repositories name change). Here is a link to the master/4.6 PR: https://github.com/kube-reporting/metering-operator/pull/1255

PR #1263, opened 2020-06-18
Get faq via static Linux binary

bentito:
This will need to be force merged to take effect.

Slight mod to simplify for Dockerfile usage from the instructions for static Linux binary given here:
https://github.com/jzelinskie/faq

### DIFF
--- a/Dockerfile.src
+++ b/Dockerfile.src
@@ -6,12 +6,14 @@ FROM openshift/origin-cli:latest as cli
 
 FROM openshift/origin-release:golang-1.13
 
-# our copy of faq and jq
-#COPY hack/faq.repo /etc/yum.repos.d/ecnahc515-faq-epel-7.repo
+# go get faq via static Linux binary approach
+ARG LATEST_RELEASE=0.0.6
+RUN curl -Lo /usr/local/bin/faq https://github.com/jzelinskie/faq/releases/download/$LATEST_RELEASE/faq-linux-amd64
+RUN chmod +x /usr/local/bin/faq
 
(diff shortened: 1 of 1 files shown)


Dockerfile.src
```diff
@@ -6,12 +6,14 @@ FROM openshift/origin-cli:latest as cli
 
 FROM openshift/origin-release:golang-1.13
 
-# our copy of faq and jq
-#COPY hack/faq.repo /etc/yum.repos.d/ecnahc515-faq-epel-7.repo
+# go get faq via static Linux binary approach
+ARG LATEST_RELEASE=0.0.6
+RUN curl -Lo /usr/local/bin/faq https://github.com/jzelinskie/faq/releases/download/$LATEST_RELEASE/faq-linux-amd64
+RUN chmod +x /usr/local/bin/faq
```
timflannagan1, line 12: I think another approach we could take is grabbing that binary from jimmy's quay account. Not sure if that's any better though.

timflannagan1: https://quay.io/repository/jzelinskie/faq?tag=latest&tab=tags

bentito: Yep, but likely ART build will require some other fix which we'll want to match here, so... punting for now.

timflannagan1:
Looks like we only addressed this in 4.6+:

https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_release/10638/rehearse-10638-periodic-ci-kube-reporting-metering-operator-release-4.5-metering-periodic-aws/1290299279805517824

PR #1265, opened 2020-06-19
Add t.Parallel() to tests & subtests and adapt

bentito:
WIP
Before looking at the e2e logs, etc. I strongly recommend checking whether my Go code is trash or not. Still learning here.

### DIFF
--- a/test/e2e/main_test.go
+++ b/test/e2e/main_test.go
@@ -15,8 +15,6 @@ import (
 )
 
 var (
-	df *deployframework.DeployFramework
-
 	kubeConfig    string
 	logLevel      string
 	runTestsLocal bool
@@ -38,6 +36,8 @@ var (
 	postUpgradeTestDirName          = "post-upgrade"
 	gatherTestArtifactsScript       = "gather-test-install-artifacts.sh"
 	testMeteringConfigManifestsPath = "/test/e2e/testdata/meteringconfigs/"
+
+	logger logrus.FieldLogger
 )
 
 type InstallTestCase struct {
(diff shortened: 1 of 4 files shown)


test/e2e/main_test.go
```diff
@@ -161,14 +156,21 @@ func TestManualMeteringInstall(t *testing.T) {
...
 
 		t.Run(testCase.Name, func(t *testing.T) {
+			t.Parallel()
+			var err error
+			var df *deployframework.DeployFramework
+
+			if df, err = deployframework.New(logger, runTestsLocal, runDevSetup, namespacePrefix, repoPath, repoVersion,
+				kubeConfig); err != nil {
+				logger.Fatalf("Failed to create a new deploy framework: %v", err)
+			}
```
timflannagan1, line 173: Is there a reason you needed to do this? We should be aiming towards parallelizing each deployment context instead of instantiating a new deploy framework per test case.

bentito: Well it seemed like the easiest way to get separation between the namespaces. Otherwise the tests fail with context errors on the api calls: mismatched namespace to call.

test/e2e/main_test.go
```diff
@@ -73,17 +73,12 @@ func TestMain(m *testing.M) {
 	flag.StringVar(&testOutputPath, "test-output-path", "", "The absolute/relative path that you want to store test logs within.")
 	flag.Parse()
 
-	logger := testhelpers.SetupLogger(logLevel)
+	logger = testhelpers.SetupLogger(logLevel)
```
EmilyM1, line 77: Looks good so far. I think that because this is in a function the short assign would be an option too

bentito: I think if I create and assign it here then it wouldn't be accessible in the other places it's used?

test/e2e/main_test.go
timflannagan1: So I think we may want to re-think that approach as the `deployframework` structure is intended to encapsulate all metering installations. We may be able to get away with adding `t.Parallel()` to `TestManualMeteringInstall`, and again at the top of the helper function that top-level test calls, and go from there. I imagine the actual deployment of metering shouldn't be a problem, at least at the surface, but I imagine running the teardown method + cleanup script + post-install tests would need further refinement.

bentito:
replaced by #1289

PR #1268, opened 2020-06-21
images: Fix conditional when checking if the cluster-wide proxy has been configured.

timflannagan1:
The first conditional check was configured to check if the `cluster_proxy_config` variable was defined. This is problematic check as that variable is the registered output of querying for the OCP cluster-wide proxy object, and in the case of Ansible, a registered variable will always be defined even if a task has been skipped. If we, instead, check if the `.resources` dictionary key is defined instead of the top-level object, that's sufficient as a baseline check.

### DIFF
--- a/images/metering-ansible-operator/roles/meteringconfig/tasks/configure_networking.yml
+++ b/images/metering-ansible-operator/roles/meteringconfig/tasks/configure_networking.yml
@@ -71,7 +71,7 @@
     _ocp_enabled_global_proxy_https_proxy: "{{ _cluster_proxy_config.status.httpsProxy }}"
   no_log: true
   when:
-  - cluster_proxy_config is defined
+  - cluster_proxy.config.resources is defined
   - cluster_proxy_config.resources | length > 0
   - _cluster_proxy_config.status is defined
   - _cluster_proxy_config.status | length > 0


images/metering-ansible-operator/roles/meteringconfig/tasks/configure_networking.yml
```diff
@@ -71,7 +71,7 @@
     _ocp_enabled_global_proxy_https_proxy: "{{ _cluster_proxy_config.status.httpsProxy }}"
   no_log: true
   when:
-  - cluster_proxy_config is defined
+  - cluster_proxy.config.resources is defined
```
bentito, line 74: It's not redundant to check if `defined` and `length > 0`

EmilyM1:
Where/why is the resources key being added?

images/metering-ansible-operator/roles/meteringconfig/tasks/configure_networking.yml
timflannagan1: At a glance it's redundant - I know this type of pattern, i.e. check if the dictionary key exists and check if non-empty, was required in previous Ansible versions. I don't think there's been any updates to the `length` jinja2 filter that handles the failure case where that dictionary key isn't defined.

bentito: ah, okay.

PR #1270, opened 2020-06-23
pkg/deploy: Cleanup conditional structure in the top-level deploy methods.

timflannagan1:
Eliminates most of the redundant if/else if/else structures, and updates the handling of non-nil errors.

### DIFF
--- a/pkg/deploy/install.go
+++ b/pkg/deploy/install.go
@@ -14,6 +14,10 @@ import (
 )
 
 func (deploy *Deployer) installNamespace() error {
+	// TODO: we can further cleanup this method by separating the
+	// creation of the namespace object and the label/annotation
+	// handling by always treating the latter as an update:
+	// https://github.com/kube-reporting/metering-operator/pull/1270#discussion_r444436226
 	namespace, err := deploy.client.CoreV1().Namespaces().Get(context.TODO(), deploy.config.Namespace, metav1.GetOptions{})
 	if apierrors.IsNotFound(err) {
 		namespaceObjectMeta := metav1.ObjectMeta{
@@ -30,14 +34,11 @@ func (deploy *Deployer) installNamespace() error {
 		/*
 			In the case where the platform is set to Openshift (the default value),
 			we need to make a few modifications to the namespace metadata.
-
 			The 'openshift.io/cluster-monitoring' labels tells the cluster-monitoring
 			operator to scrape Prometheus metrics for the installed Metering namespace.
-
 			The 'openshift.io/node-selector' annotation is a way to control where Pods
 			get scheduled in a specific namespace. If this annotation is set to an empty
 			label, that means that Pods for this namespace can be scheduled on any nodes.
-
 			In the case where a cluster administrator has configured a value for the
 			defaultNodeSelector field in the cluster's Scheduler object, we need to set
 			this namespace annotation in order to avoid a collision with what the user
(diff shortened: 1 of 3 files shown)


pkg/deploy/install.go
```diff
@@ -65,39 +65,43 @@ func (deploy *Deployer) installNamespace() error {
...
-		return err
+
+		_, err := deploy.client.CoreV1().Namespaces().Update(context.TODO(), namespace, metav1.UpdateOptions{})
+		if err != nil {
+			return fmt.Errorf("failed to add the 'openshift.io/cluster-monitoring' label to the %s namespace: %v", deploy.config.Namespace, err)
+		}
+		return nil
 	}
 
+	// TODO: handle updating the namespace for non-openshift installations
```
pweil-, line 102: You can eliminate a few more of the redundancies and this TODO if you think about this method in two pieces: the ns itself and the labels/annotations.  Ie. you always treat the labels/annotations as an update.  

What you get: the updates for existing namespaces, removal of the duplicate code to annotate and label in the openshift deployment case, ensuring the `ExtraNamespaceLabels` are applied uniformly (looks like they're only applied here in the creation case)

What you lose: logging the add vs update but that could be done with some additional logic if it really is meaningful.  I guess the same for the labels/annotations cased for the openshift deployment type.  Does it harm anything to keep it on there in all cases? (Not looking for an answer, just something to consider)

So something like this: (note, I didn't try to write this in an IDE or check it for formatting, it may not even work.  Demo purposes only 😄 )


```go
func (deploy *Deployer) installNamespace() error {
	namespace, err := deploy.client.CoreV1().Namespaces().Get(context.TODO(), deploy.config.Namespace, metav1.GetOptions{})
	if err != nil && (!apierrors.IsNotFound(err) || !apierrors.IsAlreadyExists(err)) {
		return err
	}

	if apierrors.IsAlreadyExists(err) {
		deploy.logger.Infof("The %s namespace already exists", deploy.config.Namespace)
	}

	// not found - create the ns
		namespaceObjectMeta := metav1.ObjectMeta{
			Name: deploy.config.Namespace,
		}

		namespaceObj := &v1.Namespace{
			ObjectMeta: namespaceObjectMeta,
		}

		namespace, err = deploy.client.CoreV1().Namespaces().Create(context.TODO(), namespaceObj, metav1.CreateOptions{})
		if err != nil {
			return err
		}
		deploy.logger.Infof("Created the %s namespace", deploy.config.Namespace)
	}

	// add extra namespace labels and annotations
	labels := namespace.ObjectMeta.Labels
	if labels == nil {
		labels = make(map[string]string)
	}

	annotations := namespace.ObjectMeta.Annotations
	if annotations == nil {
		annotations = make(map[string]string)
	}

	for key, val := range deploy.config.ExtraNamespaceLabels {
		labels[key] = val
		deploy.logger.Infof("Labeling the %s namespace with '%s=%s'", deploy.config.Namespace, key, val)
	}

	/*
		In the case where the platform is set to Openshift (the default value),
		we need to make a few modifications to the namespace metadata.

		The 'openshift.io/cluster-monitoring' labels tells the cluster-monitoring
		operator to scrape Prometheus metrics for the installed Metering namespace.

		The 'openshift.io/node-selector' annotation is a way to control where Pods
		get scheduled in a specific namespace. If this annotation is set to an empty
		label, that means that Pods for this namespace can be scheduled on any nodes.

		In the case where a cluster administrator has configured a value for the
		defaultNodeSelector field in the cluster's Scheduler object, we need to set
		this namespace annotation in order to avoid a collision with what the user
		has supplied in their MeteringConfig custom resource. This implies that whenever
		a cluster has been configured to schedule Pods using a default node selector,
		those changes must also be propogated to the MeteringConfig custom resource, else
		the Pods in Metering namespace will be scheduled on any available node.
	*/
	if deploy.config.Platform == "openshift" {
		labels["openshift.io/cluster-monitoring"] = "true"
		deploy.logger.Infof("Added the 'openshift.io/cluster-monitoring' label to the %s namespace", deploy.config.Namespace)

		annotations["openshift.io/node-selector"] = ""
		deploy.logger.Infof("Added the empty 'openshift.io/node-selector' annotation to the %s namespace", deploy.config.Namespace)
	}

	namespace.ObjectMeta.Labels = labels
	namespace.ObjectMeta.Annotations = annotations

	_, err = deploy.client.CoreV1().Namespaces().Update(context.TODO(), namespace, metav1.UpdateOptions{})
	if err != nil {
		return fmt.Errorf("failed to update labels and annotations on the %s namespace: %v", deploy.config.Namespace, err)
	}
	return nil
}
```

pkg/deploy/install.go
```diff
@@ -15,88 +15,84 @@ import (
...
+	*/
+	if deploy.config.Platform == "openshift" {
+		labels["openshift.io/cluster-monitoring"] = "true"
+		deploy.logger.Infof("Added the 'openshift.io/cluster-monitoring' label to the %s namespace", deploy.config.Namespace)
+
+		annotations["openshift.io/node-selector"] = ""
+		deploy.logger.Infof("Added the empty 'openshift.io/node-selector' annotation to the %s namespace", deploy.config.Namespace)
+	}
+
+	// in order to avoid updating a stale namespace resourceVersion, grab
```
pweil-, line 83: If the object has changed since creation/fetch and the update on line 92 then you're risking overwriting labels or annotations that could've been added with lines 89 & 90.  You'd have to put this into a control loop if this is really an issue - and you're still running the risk of resource version change between 85 & 92 anyway.

Probably better to just let it fail out then try to program around it without a guaranteed update loop.  

(also shouldn't ignore the `NotFound` on line 86 if this is really necessary - means someone deleted it from under us)

timflannagan1: I think the problem is that we're always going to referencing an out-of-date resource version when the namespace doesn't already exist. In the current state of this utility, it's only used for dev/testing, so overwriting any labels or annotations shouldn't be too much of a concern. I'm going to revert these changes to the namespace installation handling, and just link a TODO comment for what you were suggesting.

pweil-: > always going to referencing an out-of-date resource version when the namespace doesn't already exist

I'm more pointing out that the comment here indicates some level of safety but that's not correct.  You could re-fetch after the create, if necessary, to get the latest resource version but the resource version conflict issue still exists.  The current code has it, this new code has it.    The issue is a "normal" problem of a system where there are multiple produces of updates and no transaction locks (like a `select for update` type mechanism).  

The only true way to program around the issue is to loop until successful by continuing to update a refreshed fetch of the object until you don't have a resource version conflict.

pweil-: and to be clear, I'm not suggesting you do a guaranteed update loop, reverting, or committing the suggesting - only providing some clarity on why this, in and of itself, probably isn't what you were shooting for.  :-)

timflannagan1: I think this is all good feedback, and it's something that I wasn't considering when I initially developed this tooling. I would prefer to get these types of changes merged, especially in the case where we have PRs that conflict with other changes that get made to this Go package, but we should revisit this once I get the OLM installation work wrapped up.

timflannagan1: Besides the control loop suggestions, I think there can be further refinement in the actual organization of the various functions and methods, and emulate something similar to the how the openshift client-go or api libraries handle the creation and updates of resources.

PR #1272, opened 2020-06-23
Dockerfile.src: Get the faq dependency via a static Linux binary.

timflannagan1:
Our CI is configured to use this Dockerfile as the build root - it essentially just adds things like `operator-courier`, `faq`, `helm`, etc. that get used throughout our testing configuration.


timflannagan1:
We most likely need to force-merge this so our 4.4 CI builds are working again. This is low risk as this is purely for our CI tests and doesn't affect our downstream RHEL builds.

timflannagan1:
:(

PR #1273, opened 2020-06-23
cmd,pkg/deploy: Support passing a list of configurable OLM flagsets/environment variables.

timflannagan1:
This functionality was mainly only used for upgrade testing in our e2e suite, so the creating and deletion of OLM resources like a Subscription were configured to use several constants/hardcoded values, like using the redhat-operators CatalogSource and the `metering-ocp` package name.

Now, you can specify the subscription name, the name, and namespace of an existing CatalogSource resource, etc. This will be mainly used for updating the e2e suite to deploy Metering using OLM, instead of manually creating the requisite objects, so this work is intended to be iterative.

### DIFF
--- a/pkg/deploy/deploy.go
+++ b/pkg/deploy/deploy.go
@@ -75,6 +75,9 @@ type Config struct {
 	Repo                     string
 	Tag                      string
 	Channel                  string
+	PackageName              string
+	CatalogSourceName        string
+	CatalogSourceNamespace   string
 	SubscriptionName         string
 	ExtraNamespaceLabels     map[string]string
 	OperatorResources        *OperatorResources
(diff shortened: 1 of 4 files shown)


timflannagan1:
Looks like this is having difficulties with uninstalling the metering subscription resource.
/hold

timflannagan1:
Accidently replaced .Get with .Delete
/hold cancel

pkg/deploy/deploy.go
```diff
@@ -240,6 +249,27 @@ func (deploy *Deployer) UninstallOLM() error {
...
+	if deploy.config.DeleteCRB {
+		err = deploy.uninstallMeteringRole()
+		if err != nil {
+			return fmt.Errorf("failed to delete the metering role: %v", err)
+		}
+
+		err = deploy.uninstallMeteringRoleBinding()
+		if err != nil {
+			return fmt.Errorf("failed to delete the metering role binding: %v", err)
+		}
```
timflannagan1, line 261: I realize this is probably not needed as OLM will create these resources for us, and mark the either the Subscription or CSV resources as the owner reference, so deleting those resources will implicitly delete these roles/rolebindings.

PR #1274, opened 2020-06-24
hack,pkg,test: Update the e2e suite to support deploying Metering using OLM.

timflannagan1:
## Overview

In order to deploy the most up-to-date operator bundle, we need to be able to create a CatalogSource custom resource that references an existing image.

We have a couple of different options for doing this:
- Create ConfigMap containing the `package.yaml`, CRDs and CSV manifests.
- Build and push an image containing the manifest bundle and an initialized registry server to an image registry.
- Build and push an image containing the manifest bundle and serve the registry server locally.

In the case of CI, we need to take the third approach as they expose a series of environment variables pointing to the newly built metering/reporting operator images.

In the `Dockerfile.registry`, we copy over the current openshift manifest bundle to a known destination path in the container and copy over a script responsible for manipulating the CSV manifest with the images that get built by CI or provided by devs through the `$METERING_OPERATOR_IMAGE_*`/`$REPORTING_OPERATOR_IMAGE_*` variables.

### Registry Manifests

In the manifest directory, we have a registry deployment, which is responsible for calling that CSV manipulation script we copied over in the Dockerfile.registry in an `initContainer`, and serving that updated manifest bundle in a local registry server. We also have a registry service, which is responsible for exposing this local registry and can be consumed by a CatalogSource custom resource using the ClusterIP of that service and the grpc port we defined.

This mainly mirrors the setup that cluster-logging and the cluster-resource-override-admission operators have already configured.

The only main differences with those implementations are that we instantiate and configure the CatalogSource and registry manifests in Go, instead of manipulating and applying those manifests using a client.

### Changes made to E2E

This updates the `deployframework` object, creating another method for instantiating a local registry and a `CatalogSource` custom resource that points to that local registry. From there, we can update the installation closure used to create and manage an installation of metering, from the manual installation method to the existing OLM installation method.

In order to create this ` CatalogSource`, we need to first decode the YAML registry manifests into Go objects, which we can then further manipulate and manage ourselves before creating those resources using the core clientsets. For the registry deployment, there are a couple of things we need to update: the default registry image, which we currently don't have a default (e.g. quay.io/openshift/metering-registry) to point to the `--registry-image` Go test flag that was provided (which is configured to read from the `$IMAGE_METERING_ANSIBLE_OPERATOR_REGISTRY` is present). In the case where there's a non-empty `$METERING_OPERATOR_IMAGE_*` or `$REPORTING_OPERATOR_IMAGE_*` environment variable(s), we also update the values in the deployment to point to those overridden images. 

A quick note on the `$METERING_OPERATOR_IMAGE_*` and `$REPORTING_OPERATOR_IMAGE_*` environment variables:
CI exposes a `$IMAGE_FORMAT` environment variable, which we can further process by doing some substitution. We don't need to do this manipulation as we already have a hack/lib/test.sh script (which is sourced in the hack/e2e.sh script) that pulls out the repository and tag for the metering-ansible-operator and metering-reporting-operator images.


test/deployframework/framework.go
```diff
@@ -196,3 +210,61 @@ func (df *DeployFramework) NewDeployerConfig(
 		MeteringConfig:           meteringConfig,
 	}, nil
 }
+
+// CreateCatalogSource is a deployframework method responsible
+// for instantiating a new CatalogSource that can be used
+// throughout individual Metering installations.
+func (df *DeployFramework) CreateCatalogSource(registryImage, meteringOperatorImage, reportingOperatorImage string) (string, string, error) {
```
timflannagan1, line 217: For whatever reason, this is still a bit awkward to me returning the name and namespace of the CatalogSource object we create. We may just want to add these fields to the deployframework object, but I'm still on the fence as to whether or not those fields are state that object needs to manage.

olm_deploy/Dockerfile.registry
```diff
@@ -0,0 +1,17 @@
...
+# Need to remove the art.yaml and image-references files from the list of OLM
+# manifests as those are invalid in a manifest bundle
+RUN chmod -R g+w /manifests && rm /manifests/art.yaml /manifests/*/image-references
+COPY olm_deploy/scripts/ /scripts/
+
+COPY --from=builder /bin/initializer /usr/bin/initializer
+COPY --from=builder /bin/registry-server /usr/bin/registry-server
+COPY --from=builder /bin/grpc_health_probe /usr/bin/grpc_health_probe
+
+WORKDIR /bundle
```
timflannagan1, line 17: Something we may want to consider further down the line is providing a better developer workflow, as in take a list of arguments for the various operator and operand images and do the CSV manipulation inside of the Dockerfile and instantiate and serve that registry bundle from Dockerfile.

In this case, we may need to extend the e2e suite or the deploy-metering utility to create a CatalogSource that points to an external image registry, instead of the current implementation which references a local registry service.

olm_deploy/manifests/deployment.yaml
```diff
@@ -0,0 +1,77 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: metering-ansible-operator-registry
```
timflannagan1, line 4: I realize we may want to change this to also include the `df.NamespacePrefix` field as a prefix to the `metadata.name` to support multi-tenancy.

olm_deploy/manifests/service.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: metering-ansible-operator-registry
```
timflannagan1, line 4: Same thing as above, probably need to add a namespace prefix here to distinguish between multiple instantiations.

test/deployframework/framework.go
```diff
@@ -196,3 +211,100 @@ func (df *DeployFramework) NewDeployerConfig(
...
+	}
+
+	serviceManifestPath := filepath.Join(df.RepoDir, olmManifestsDir, registryServiceManifestName)
+	addr, err := CreateRegistryService(df.Logger, df.Client, serviceManifestPath, registryDeployNamespace)
+	if err != nil {
+		return "", "", fmt.Errorf("failed to create the registry service manifest in the %s namespace: %v", err, registryDeployNamespace)
+	}
+	if addr == "" {
+		return "", "", fmt.Errorf("the registry service spec.ClusterIP returned is empty")
+	}
```
timflannagan1, line 232: We should probably reverse the creation of resources here as it's traditionally good practice to instantiate the service object before the deployment/pod that service is responsible for routing traffic.

test/deployframework/helpers.go
```diff
@@ -574,11 +576,108 @@ func DeleteAllTestReports(logger logrus.FieldLogger, client meteringclient.Meter
...
+		blindly indexing into this the environment array.
+	*/
+	if meteringOperatorImage != "" {
+		logger.Infof("Overriding the default metering-ansible-operator image with %s", meteringOperatorImage)
+		deployment.Spec.Template.Spec.InitContainers[0].Env[0].Value = meteringOperatorImage
+	}
+	if reportingOperatorImage != "" {
+		logger.Infof("Overriding the default metering-reporting-operator image with %s", reportingOperatorImage)
+		deployment.Spec.Template.Spec.InitContainers[0].Env[1].Value = reportingOperatorImage
+	}
```
timflannagan1, line 636: Building on top of what the above comment suggests, we may be able to just get away with creating a constant variable from the environment variable initContainer array, e.g. `meteringOperatorEnvVarIndex = 0`, to help provide some more context.

test/e2e/metering_manual_install_tests.go
```diff
@@ -52,14 +56,23 @@ func testManualMeteringInstall(
...
-
 	deployerCtx.Logger.Infof("DeployerCtx: %+v", deployerCtx)
-	rf, err := deployerCtx.Setup(deployerCtx.Deployer.Install, expectInstallErr)
+
+	installFunc := deployerCtx.Deployer.InstallOLM
+	uninstallFunc := deployerCtx.Deployer.UninstallOLM
+	if installationMethod != "olm" {
+		installFunc = deployerCtx.Deployer.Install
+		uninstallFunc = deployerCtx.Deployer.Uninstall
+	}
```
timflannagan1, line 74: This is something we start addressing in the near future, with the intent to provide more flexibility to the e2e suite. At the moment, we pass a closure to the `deployerCtx.Setup` method that expects a `func() error` type signature, but this doesn't account for cases where we expect an error may occur when attempting to create an invalid resource, e.g. a MeteringConfig CR missing a storage specification, or we need to do some pre-processing before installing Metering.

As a short-term workaround, we may be able to just pass a global flag, or a per-test-table-indice installation method to help alleviate this. The cleanest solution is most likely making the registry image a required e2e field, which would also eliminate the need to do this kind of check.

timflannagan1: And just to tack on to that last comment and provide more context on the `... MeteringConfig CR missing a storage specification`, we shouldn't need to instantiate a deploy-metering context to test out our CRD schema validation as that's a waste of time and pings the api-server more than necessary. We could, instead, update the `deployerCtx.Setup` method to support the different tests that we may want to perform. In that case, we could instead, pass a closure that attempts to create the invalid `MeteringConfig` custom resource, and performs a check to whether or not an error was returned.

olm_deploy/manifests/deployment.yaml
bentito: ah, yeah, I think that is what I am hitting on with my parallelization work. I see errors like:
```
            	            	failed to install metering: failed to create the metering resources: failed to create the metering role: the namespace of the provided object does not match the namespace sent on the request
            	Test:       	TestManualMeteringInstall/ValidHDFS-ReportStaticInputData
```

timflannagan1: In that case, it's most likely we're passing the `deploy.config.Namespace` as the namespace parameter to the Create method, which is different than what's in the `metadata.namespace` of the resource we're also passing as we're decoding a YAML manifest to a Go object, and modifying the resource from there.

olm_deploy/Dockerfile.registry
bentito: Is the naming of this file something required by OLM, or I don't understand the naming?

timflannagan1: The naming is arbitrary - we would just reference this Dockerfile in the release repository as part of our configuration, and other places in the repository such as hack scripts and Makefile targets.

olm_deploy/manifests/service.yaml
timflannagan1: Done - added the namespace prefix that gets passed to the e2e suite (typically the value is the `$METERING_NAMESPACE` environment variable)

test/deployframework/framework.go
timflannagan1: Done

hack/e2e.sh
```diff
@@ -30,6 +30,8 @@ METERING_SHORT_TESTS="${METERING_SHORT_TESTS:=false}"
 METERING_REPO_PATH="${METERING_REPO_PATH:=${ROOT_DIR}}"
 METERING_REPO_VERSION="$(basename $(find $MANIFESTS_BASE_DIR -type d | sort -r | head -n 1))"
 METERING_RUN_TESTS_LOCALLY="${METERING_RUN_TESTS_LOCALLY:=false}"
+METERING_OLM_SUBSCRIPTION_CHANNEL="${METERING_OLM_SUBSCRIPTION_CHANNEL:="4.6"}"
+IMAGE_METERING_ANSIBLE_OPERATOR_REGISTRY="${IMAGE_METERING_ANSIBLE_OPERATOR_REGISTRY:=""}"
```
timflannagan1, line 34: This is the name of the image that CI should be exposing once we update the release repository to start building the Dockerfile.registry. Else, you would need to manually override this variable when running the `make e2e` target:

```bash
$ ./hack/push-olm-manifests.sh quay.io/tflannag/metering-registry:latest
$ make e2e IMAGE_METERING_ANSIBLE_OPERATOR_REGISTRY=quay.io/tflannag/metering-registry:latest
```

timflannagan1:
Welp that was obvious in hindsight, going to make that a non-required field for now, but hardcode the installation closure to be using the OLM method instead of the manual method:

```
time="06-25-2020 18:43:22" level=info msg="Setting the log level to debug" app=deploy
time="06-25-2020 18:43:22" level=fatal msg="Error: the --registry-image is a required field and the value is empty" app=deploy
FAIL	github.com/kube-reporting/metering-operator/test/e2e	0.026s
FAIL
```

hack/e2e.sh
timflannagan1: Made some more changes - renamed that variable and exported it in the `hack/lib/tests.sh` so you would only need to run `make e2e` with the `export METERING_ANSIBLE_OPERATOR_IMAGE_REGISTRY` specified locally (unless this is running in CI where the value we pull out of `$IMAGE_FORMAT` would take precedence.)

hack/e2e.sh
```diff
@@ -30,6 +30,8 @@ METERING_SHORT_TESTS="${METERING_SHORT_TESTS:=false}"
 METERING_REPO_PATH="${METERING_REPO_PATH:=${ROOT_DIR}}"
 METERING_REPO_VERSION="$(basename $(find $MANIFESTS_BASE_DIR -type d | sort -r | head -n 1))"
 METERING_RUN_TESTS_LOCALLY="${METERING_RUN_TESTS_LOCALLY:=false}"
+METERING_OLM_SUBSCRIPTION_CHANNEL="${METERING_OLM_SUBSCRIPTION_CHANNEL:="4.6"}"
+METERING_ANSIBLE_OPERATOR_IMAGE_REGISTRY="${METERING_ANSIBLE_OPERATOR_IMAGE_REGISTRY:=""}"
```
timflannagan1, line 34: Something that should be noted is that we don't delete the registry resources that get created (deployment, service, CatalogSource) if you kill the `make e2e` process, but we do defer the deletion of those resources as long as the testing suite wasn't run in developer mode.

PR #1276, opened 2020-06-25
test/e2e: Add a wrapper function around the top-level TestMain function.

timflannagan1:
This is needed in the case where you want to defer any test cleanup as os.Exit doesn't respect that call in TestMain.
If we instead create a wrapper function around TestMain and call that wrapper function in TestMain, then we can start
cleaning up any testing resources in the e2e suite, instead of doing all of that work as part of the `trap cleanup EXIT` call in hack/e2e.sh.

Link to article that explains that behavior:
- http://blog.englund.nu/golang,/testing/2017/03/12/using-defer-in-testmain.html

### DIFF
--- a/test/e2e/main_test.go
+++ b/test/e2e/main_test.go
@@ -56,6 +56,21 @@ func init() {
 }
 
 func TestMain(m *testing.M) {
+	os.Exit(testMainWrapper(m))
+}
+
+// testMainWrapper is a wrapper function around the
+// top-level TestMain function and this pattern is
+// needed as os.Exit() doesn't respect any defer calls
+// that may occur during the TestMain workflow. If we
+// we instead doing the heavy-lifting in this function,
+// and then return an integer code that os.Exit can correctly
+// interpret, then the defer call will work.
+//
+// See the following references for more information:
+// - https://golang.org/pkg/os/#Exit
+// - http://blog.englund.nu/golang,/testing/2017/03/12/using-defer-in-testmain.html
+func testMainWrapper(m *testing.M) int {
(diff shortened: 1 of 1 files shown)


test/e2e/main_test.go
```diff
@@ -56,6 +56,20 @@ func init() {
...
+// testMainWrapper is a wrapper function around the
+// top-level TestMain function and this pattern is
+// needed as os.Exit() doesn't respect any defer calls
+// that may occur during the TestMain workflow. If we
+// we instead doing the heavy-lifting in this function,
+// and then return an integer code that os.Exit can correctly
+// interpret, then the defer call will work.
+//
+// Link to article that explains this behavior:
+// http://blog.englund.nu/golang,/testing/2017/03/12/using-defer-in-testmain.html
```
pweil-, line 71: > * http://blog.englund.nu/golang,/testing/2017/03/12/using-defer-in-testmain.html

This is a normal part of the language and [defined in the spec](https://golang.org/pkg/os/#Exit).  I'd link there if you want folks to have official reference.

For reviewers: Because the function never reaches the end of its body the [defer semantics](https://golang.org/ref/spec#Defer_statements) do not apply.  [Simple example](https://play.golang.org/p/_AVhjzsIHgR).  For those with a Java background, think of this like throwing `System.exit()` somewhere in a program.  😄

timflannagan1: Yeah, that's a good call on linking to the official documentation there. I'll add that now as the tests just flaked and I would need to re-run anyways.

timflannagan1:
more flakes
/retest

PR #1281, opened 2020-06-26
olm_deploy: Provide the Openshift OLM bundle as the default value for the MANIFEST_BUNDLE build argument.

timflannagan1:


timflannagan1:
Used the wrong syntax, updated to add the `=` between the argument name and default value:
```
ARG MANIFEST_BUNDLE=manifests/deploy/openshift/olm/bundle
```

timflannagan1:
ci/prow/images passed and this is blocking other tests, so forcing this in now.

PR #1282, opened 2020-06-29
hack,test/e2e: Wait for the metering-ocp packagemanifest to be ready before installing Metering.

timflannagan1:


hack/e2e.sh
```diff
@@ -14,8 +14,13 @@ function cleanup() {
...
         echo "Removing namespaces with the 'name=${METERING_NAMESPACE}-metering-testing-ns' label"
         kubectl delete ns -l "name=${METERING_NAMESPACE}-metering-testing-ns" --wait=false || true
+
+        # Note: the `openshift-marketplace` namespace is hardcoded for now until we have the need
+        # to create the registry-related resources in another namespace (e.g. testing upstream manifests).
+        echo "Removing the local registry resources with the 'name=${METERING_NAMESPACE}-metering-testing-ns' label"
+        kubectl -n openshift-marketplace delete catsrc,deployment,service -l "name=${METERING_NAMESPACE}-metering-testing-ns" --wait=false || true
     else
-        echo "Skipping the namespace deletion"
+        echo "Skipping the deletion of the metering testing namespaces and the local registry resources"
```
bentito, line 23: Is this the same idea? I think the idea on devs not wanting the namespace deleted was to be able to examine state after tests run, does that apply to the OLM namespace too?

timflannagan1: This will only clean up those OLM-related resources if you're running a non-dev Makefile target to deploy the e2e tests.

timflannagan1: But in general, it's nice to have as we don't support making any updates to the existing registry deployment or service objects, we just GET and CREATE them.

bentito: image

PR #1287, opened 2020-06-29
[release-4.3] Dockerfile.src: Get the faq dependency via a static Linux binary.

timflannagan1:


timflannagan1:
This only affects CI builds, and these changes need to be force-merged.

PR #1288, opened 2020-06-30
test: Add support for running multiple post-install Metering tests.

timflannagan1:
## Overview of Changes

This continues the trend of follow-ups to the recent e2e testing changes and aims towards adding some initial support for the parallelization of tests and running multiple post-installation tests.

### Parallelization Changes

I added a `t.Parallel` call at the `t.Run(...)` call site for running the top-level `TestManualMeteringInstall` which spawns off several Metering installations at once. I was having some difficulties with any further parallelization attempts, namely trying to run several post-install tests at the same time. I need to further investigate how to block a non-top-level test from returning early.

### General Changes

- Updated all of the context logging (e.g. the logs that get produced from an individual Metering installations test run) to log to a file, instead of stdout.
- Created a few helper functions (creating the context logger, decoding the MeteringConfig YAML manifests to Go) to clean-up some of the test functions.
- Updated the manual installation test helper function to support running multiple post-install tests against a single installation. This namely involved adding support for passing an array of `InstallTestCase` types and we iterate over that list and call the `InstallTestCase.TestFunc` during the `t.Run` call.


timflannagan1:
Link to the test output: https://storage.googleapis.com/origin-ci-test/pr-logs/pull/kube-reporting_metering-operator/1288/pull-ci-kube-reporting-metering-operator-master-metering-e2e-aws/1277971104706596864/artifacts/metering-e2e-aws/metering_test_output/e2e/tests/e2e-tests.log

test/deployframework/context.go
```diff
@@ -120,8 +121,31 @@ func (df *DeployFramework) NewDeployerCtx(
...
+	logger.SetLevel(logrus.DebugLevel)
+	logger.SetFormatter(&logrus.TextFormatter{
+		FullTimestamp:   true,
+		TimestampFormat: "01-02-2006 15:04:05",
+		ForceColors:     true,
+	})
+
+	outLogger := logger.WithFields(logrus.Fields{
+		"context": namespace,
+	})
```
timflannagan1, line 145: I need to clean this up a bit, but the general idea is that for an individual Metering installation, instead of logging everything to stdout using the same logger as the `deployframework` structure, we create another instance logger and re-direct it's output to a file.

timflannagan1:
That test run took ~15m to complete, but that a decent improvement from the ~30-35m average we were seeing before these types of changes.

test/e2e/metering_manual_install_tests.go
```diff
@@ -59,23 +60,50 @@ func testManualMeteringInstall(
 		catalogSourceNamespace,
 		subscriptionChannel,
 		testCaseOutputBaseDir,
-		testInstallFunction.ExtraEnvVars,
+		envVars,
 		mc.Spec,
 	)
 	require.NoError(t, err, "creating a new deployer context should produce no error")
-	deployerCtx.Logger.Infof("DeployerCtx: %+v", deployerCtx)
+	defer deployerCtx.LoggerOutFile.Close()
```
timflannagan1, line 67: We need to defer closing the file descriptor in this test helper function to avoid logging to a closed file during the post-install test runs.

bentito:
So in parallel (pun intended) I was doing this bit where I lifted the context up out of the sub-test itself. I think it allows the sub-tests to be run in parallel as well.

https://github.com/kube-reporting/metering-operator/pull/1289

PR #1290, opened 2020-07-01
test: Add a post-install test for checking if the service-serving CA bundle has been properly mounted.

timflannagan1:
We're now responsible for ensuring the contents of this service-serving CA bundle are distributed to our operands instead of that CA bundle being automatically injected by a controller in the cluster. As part of that change, we create a ConfigMap with the `service.beta.openshift.io/inject-cabundle="true"` annotation, and we mount the service-ca.crt in the `/var/run/configmaps/service-ca-bundle` path in the various operand containers.

This adds a post-install check that verifies that container is present and this type of test can be further iterated on to ensure that all operands Pods (and their corresponding containers) have this CA bundle present, and that CA bundle matches what we expect.


timflannagan1:
That was strange
/retest

test/e2e/service_serving_ca_bundle_exists_test.go
```diff
@@ -0,0 +1,71 @@
...
+	require.NoError(t, err, "failed to execute the REST request")
+
+	var stdoutBuf, stderrBuf bytes.Buffer
+	err = exec.Stream(remotecommand.StreamOptions{
+		Stdin:  nil,
+		Stdout: &stdoutBuf,
+		Stderr: &stderrBuf,
+		Tty:    false,
+	})
+	assert.NoError(t, err, "failed to stream the exec results")
```
timflannagan1, line 66: We should probably generalize this as a helper function of sorts that takes something like a `*rest.Config`, `corev1.PodExecOptions` and returns a stdout/stderr buffer and error that was encountered.

test/e2e/service_serving_ca_bundle_exists_test.go
```diff
@@ -0,0 +1,71 @@
...
+
+	t.Logf("Stdout: %v", stdoutBuf.String())
+	t.Logf("Stderr: %v", stderrBuf.String())
+	assert.Contains(t, stdoutBuf.String(), "service-ca.crt", "expected that the service-ca.crt would be present in the reporting-operator container")
```
timflannagan1, line 70: We should beef up these checks as well to ensure that `stderr` is empty, at least in the current form where we always expect this kind of check to succeed, and the contents of that `service-ca.crt` bundle matches what we expect.

test/e2e/service_serving_ca_bundle_exists_test.go
```diff
@@ -0,0 +1,71 @@
...
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/client-go/tools/remotecommand"
+
+	"github.com/kube-reporting/metering-operator/test/reportingframework"
+)
+
+func testReportOperatorServiceCABundleExists(t *testing.T, rf *reportingframework.ReportingFramework) {
```
timflannagan1, line 20: Going forward, we'll most like want to iterate over the operand containers, exec into each container, and ensure the service-ca.crt exists.

test/e2e/presto_prometheus_connector_up_test.go
```diff
@@ -2,18 +2,17 @@ package e2e
...
 	"strings"
 
+	"github.com/kube-reporting/metering-operator/test/reportingframework"
+	"github.com/stretchr/testify/require"
+
 	"testing"
 )
 
 func testPrometheusConnectorWorks(t *testing.T, rf *reportingframework.ReportingFramework) {
-	t.Helper()
```
bentito, line 15: Was this just leftover junk that got missed in a previous revision?

timflannagan1: Yeah `t.Helper()` will silence any of these test helper logs, and we moved all of the post-install logging into its own file, so I think removing this and capturing the logs from this function is best.

test/e2e/service_serving_ca_bundle_exists_test.go
```diff
@@ -0,0 +1,71 @@
...
+	// Attempt to grab the reporting-operator Pod in the rf.Namespace.
+	// Because this Pod is spun up using a Deployment instead of a Statefulset
+	// we need to list any Pods matching the app=reporting-operator label as we
+	// don't know the name ahead of time.
+	podList, err := rf.KubeClient.CoreV1().Pods(rf.Namespace).List(context.Background(), metav1.ListOptions{
```
bentito, line 25: Is waiting for all pods to have spun up implicit here somehow? Like if we're at this point in the code than that's a given?

timflannagan1: We wait for all Pods to become "ready" before running any of the post-install tests, so yes that's the assumption.

test/e2e/service_serving_ca_bundle_exists_test.go
```diff
@@ -0,0 +1,71 @@
...
+		LabelSelector: "app=reporting-operator",
+	})
+	require.NoError(t, err, "failed to list the reporting-operator Pod in the %s namespace", rf.Namespace)
+	require.Len(t, podList.Items, 1, "expected the list of app=reporting-operator Pods in the %s namespace to match a length of 1", rf.Namespace)
+
+	// We expect there's only going to be one reporting-operator
+	// in a single namespace, so hardcode the reporting-operator Pod
+	// object to be the first item in the list returned.
+	pod := podList.Items[0]
```
bentito, line 30:
```suggestion
  const ThereCanBeOnlyOne = 0
	pod := podList.Items[0]
```

test/e2e/service_serving_ca_bundle_exists_test.go
bentito: I was trying to find where the pod waiting happens

timflannagan1: https://github.com/kube-reporting/metering-operator/blob/master/test/deployframework/context.go#L221-L224

test/e2e/service_serving_ca_bundle_exists_test.go
pweil-: is it more meaningful to test the expectation in the comment than actually have a constant.  If there should only be one maybe we should fail if there is more than one thing returned.

timflannagan1: Before assigning to this variable, we already require that the list of Pods returned by that label selector is a length of one:
```
	require.Len(t, podList.Items, 1, "expected the list of app=reporting-operator Pods in the %s namespace to match a length of 1", rf.Namespace)
```

pweil-: ah, I see now.  Thank you.

PR #1292, opened 2020-07-04
test: Support running post-install tests in parallel.

timflannagan1:


timflannagan1:
Welp, I realize this was a bit short-sighted in hindsight as we essentially block running sub-tests from another context while another context's sub-tests get to run, due to the blocking nature of the `t.Run("group"), ...`, which is problematic in the case of the `testReportingProducesData` as that sub-tests takes the longest. We'll see how this implementation fairs against the average e2e test run we've been seeing.

PR #1296, opened 2020-07-08
test/e2e: Add support for running post-install tests against a Metering installation using s3.

timflannagan1:


test/e2e/metering_manual_install_tests.go
```diff
@@ -107,3 +125,72 @@ func DecodeMeteringConfigManifest(basePath, manifestPath, manifestFilename strin
...
+	_, err = ctx.Client.CoreV1().Secrets(ctx.Namespace).Get(context.Background(), secretName, metav1.GetOptions{})
+	if apierrors.IsNotFound(err) {
+		s, err := ctx.Client.CoreV1().Secrets("kube-system").Get(context.Background(), secretName, metav1.GetOptions{})
+		if err != nil {
+			return err
+		}
+
+		key, ok := s.Data["aws_access_key_id"]
+		if !ok {
+			return fmt.Errorf("failed to grab the AWS access key ID from the secret data")
```
pruan-rht, line 170: s/grab/retrieve/

test/e2e/metering_manual_install_tests.go
```diff
@@ -107,3 +125,72 @@ func DecodeMeteringConfigManifest(basePath, manifestPath, manifestFilename strin
...
+		}
+		access, ok := s.Data["aws_secret_access_key"]
+		if !ok {
+			return fmt.Errorf("failed to grab the AWS secret access key from the secret data")
```
pruan-rht, line 174: s/grab/retrieve/

test/e2e/metering_manual_install_tests.go
```diff
@@ -107,3 +125,72 @@ func DecodeMeteringConfigManifest(basePath, manifestPath, manifestFilename strin
...
+		}
+
+		secret := &corev1.Secret{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      secretName,
+				Namespace: ctx.Namespace,
+			},
+			Data: map[string][]byte{
+				"aws-access-key-id":     key,
```
pruan-rht, line 183: @timflannagan1 the official AWS cred format is `aws_access_key_id`, should we follow suit here?

timflannagan1: Unfortunately, that's not what we expect to see on the helm-chart level, which is why some translation is needed when decoding the kube-system aws-creds secret (which uses the `aws_access_key_id` data key name): https://github.com/kube-reporting/metering-operator/blob/master/charts/openshift-metering/templates/reporting-operator/reporting-operator-aws-credentials-secrets.yaml#L12

timflannagan1: That said, I should call out the need to do this translation somewhere in a comment.

test/e2e/metering_manual_install_tests.go
```diff
@@ -113,3 +125,72 @@ func DecodeMeteringConfigManifest(basePath, manifestPath, manifestFilename strin
...
+	// The default ctx.TargetPodsCount value assumes that HDFS
+	// is being used a storage backend, so we need to decrement
+	// that value to ensure we're not going to poll forever waiting
+	// for Pods that will not be created by the metering-ansible-operator
+	ctx.TargetPodsCount = nonHDFSTargetPodCount
+
+	// Before we can create the AWS credentials secret in the ctx.Namespace, we need to ensure
+	// that namespace has been created before attempting to query for a resource that does not exist.
+	_, err := ctx.Client.CoreV1().Namespaces().Get(context.Background(), ctx.Namespace, metav1.GetOptions{})
+	if apierrors.IsNotFound(err) {
```
pweil-, line 139: do you need to handle if the error is non-nil and NOT `IsNotFound`? Same on line 162 (which, if I'm reading correctly, could overwrite this err object since it is shadowing the var).

timflannagan1: Oh yeah, good catch - looks like we just ignore non-IsNotFound errors.

PR #1297, opened 2020-07-08
test/e2e: Add support for running any pre-installation tests.

timflannagan1:
In order to add more flexibility to the e2e suite, we need to support creating any pre-installation resources before punting the current test context to the deploy-metering Go package. In order to do this we can create a function type that takes a deployframework.DeployerCtx object, with the assumption that when we run the preInstallFunc, that state has been initialized. This is useful in the case where we want to create things like an NFS provisioner, or create a secret containing the AWS credentials that will be used throughout the testing suite.

### DIFF
--- a/test/e2e/main_test.go
+++ b/test/e2e/main_test.go
@@ -47,12 +47,6 @@ var (
 	testMeteringConfigManifestsPath = "/test/e2e/testdata/meteringconfigs/"
 )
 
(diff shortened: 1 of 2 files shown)


timflannagan1:
In the future, we may just want to inflate the deploy metering package to support passing additional installation functions, as we do with extra environment variables, or pass additional teardown functions.

PR #1298, opened 2020-07-08
Additions for Report expiration

bentito:
Less WIP:

- Does have any required manifest changes;
- Does not have validation on the retention value provided;
- Does have consideration of `spec.input` dependencies (please check that seem complete and correct way to go about doing them though)

Note: Do we think the new `report.Spec.Expiration` on the Report CRD should just be a duration to be added to Report create timestamp?

### DIFF
--- a/pkg/apis/metering/v1/report.go
+++ b/pkg/apis/metering/v1/report.go
@@ -34,6 +34,9 @@ type ReportSpec struct {
 	// Schedule configures when the report runs.
 	Schedule *ReportSchedule `json:"schedule,omitempty"`
 
+	// Expiration of the report. Report will be deleted once Report creation time + this retention period is reached, if no deps found
+	Expiration *meta.Duration `json:"expiration,omitempty"`
+
 	// ReportingStart specifies the time this Report should start from
 	// instead of the current time.
 	// This is intended for allowing a Report to start from the past
--- a/pkg/operator/operator.go
+++ b/pkg/operator/operator.go
@@ -22,6 +22,7 @@ import (
 	log "github.com/sirupsen/logrus"
 	"github.com/taozle/go-hive-driver"
 	v1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/clock"
 	"k8s.io/apimachinery/pkg/util/wait"
@@ -843,6 +844,44 @@ func (op *Reporting) newPrometheusConn(promConfig promapi.Config) (prom.API, err
 		return nil, fmt.Errorf("can't connect to prometheus: %v", err)
 	}
 	return prom.NewAPI(client), nil
+}
+
+func (op *Reporting) handleExpiredReport(report *metering.Report, now time.Time) error {
+	// check if the Report is being deleted already
+	if report.DeletionTimestamp != nil {
+		op.logger.Warnf("report was already marked for deletion")
+		return nil
+	}
+	// check if the Report is past its retention time
+	if reportExpired := isReportExpired(op.logger, report, now); !reportExpired {
+		op.logger.Debugf("the %s Report in the %s namespace has not yet reached the expiration date", report.Name, report.Namespace)
+		return nil
+	}
+	// check if the Report is used by any other Report or ReportQuery, if not delete it
+	if reportIsNotInput := isReportNotUsedAsInput(report, op); reportIsNotInput {
+		newDeletionTimestamp := metav1.Now()
+		report.SetDeletionTimestamp(&newDeletionTimestamp)
+		err := op.meteringClient.MeteringV1().Reports(report.Namespace).
+			Delete(context.TODO(), report.Name, metav1.DeleteOptions{})
+		if apierrors.IsNotFound(err) {
+			op.logger.Infof("report: %s, not deleted because it was not found at time delete attempted", report.Name)
+			return nil
+		}
+		if err != nil {
+			return fmt.Errorf("failed to delete the expired report: %s. err: %s", report.Name, err)
+		}
+		op.logger.Infof("deleted Report: %s in the Namespace: %s because it reached the expiration time",
+			report.Name, report.Namespace)
+		op.eventRecorder.Event(report, v1.EventTypeNormal, "ExpiredReportHasBeenDeleted",
+			"Deleted the %s Report as the configured expiration date has passed")
+		return nil
+	}
+	// if here, warn about dependency before return
+	op.logger.Warnf("report: %s, would be deleted because expired, but is depended on", report.Name)
+	op.eventRecorder.Event(report, v1.EventTypeWarning, "ExpiredReportHasDependencies",
+		"Skipping the deletion of the %s Report as other resources are dependent on it, "+
+			"despite reaching the desired expiration date.")
+	return nil
 }
 
 type DependencyResolver interface {
--- a/pkg/operator/reports.go
+++ b/pkg/operator/reports.go
@@ -91,7 +91,13 @@ func (op *Reporting) syncReport(logger log.FieldLogger, key string) error {
 		}
 		return err
 	}
+
 	sr := report.DeepCopy()
+
+	err = op.handleExpiredReport(sr, time.Now())
+	if err != nil {
+		return err
+	}
 
 	if report.DeletionTimestamp != nil {
 		_, err = op.removeReportFinalizer(sr)
@@ -347,6 +353,11 @@ func getReportPeriod(now time.Time, logger log.FieldLogger, report *metering.Rep
 func (op *Reporting) runReport(logger log.FieldLogger, report *metering.Report) error {
 	// check if the report was previously finished; store result in bool
 	if reportFinished := isReportFinished(logger, report); reportFinished {
+		return nil
+	}
+
+	// check if report is expired (CreationTime + expiration). If true, stop processing and exit early.
+	if reportExpired := isReportExpired(logger, report, time.Now()); reportExpired {
 		return nil
 	}
 
@@ -742,6 +753,60 @@ func (op *Reporting) runReport(logger log.FieldLogger, report *metering.Report) 
 	return nil
 }
 
+// We check first for Reports depending on this Report, return if any, and then check ReportQueries depending on Report
+func isReportNotUsedAsInput(report *metering.Report, op *Reporting) bool {
+	// Consider Reports referencing this report in the namespace of this report
+	reports, err := op.reportLister.Reports(report.Namespace).List(labels.Everything())
+	if err != nil {
+		op.logger.Errorf("unable to determine list of reports that might be input for report: %s", report.Name)
+		return false
+	}
+	for _, report := range reports {
+		// If Report has no spec.inputs it can't be depending on this report.
+		if report.Spec.Inputs == nil {
+			continue
+		}
+		if does, depReport := op.reportsDependOnThisReport(report, reports); does {
+			op.logger.Infof(
+				"report %s exists that uses report %s as input, "+
+					"will not delete though retention period has expired", depReport, report.Name)
+			return false
+		}
+	}
+	// Consider ReportQueries referencing this report in the namespace of this report
+	reportQueries, err := op.reportQueryLister.ReportQueries(report.Namespace).List(labels.Everything())
+	if err != nil {
+		op.logger.Errorf("unable to determine list of ReportQueries that might be input for report: %s", report.Name)
+		return false
+	}
+	for _, reportQuery := range reportQueries {
+		// If ReportQuery has no spec.inputs it can't be depending on this report.
+		if reportQuery.Spec.Inputs == nil {
+			continue
+		}
+
+		if does, depQuery := op.reportQueriesDependOnThisReport(report, reportQueries); does {
+			op.logger.Infof("ReportQuery, %s exists that uses report %s as input, " +
+				"will not delete though retention period has expired", depQuery, report.Name)
+			return false
+		}
+	}
+	return true
+}
+
+func isReportExpired(logger log.FieldLogger, report *metering.Report, currentTime time.Time) bool {
+	if report.Spec.Expiration != nil {
+		currentTimeNano := currentTime.UTC().Truncate(time.Millisecond).UnixNano()
+		reportCreationNano := report.ObjectMeta.CreationTimestamp.UTC().Truncate(time.Millisecond).UnixNano()
+		retentionAmountNano := report.Spec.Expiration.Nanoseconds()
+		if currentTimeNano > (reportCreationNano + retentionAmountNano) {
+			return true
+		}
+		logger.Infof("curr: %d, creat: %d, retain: %d", currentTimeNano, reportCreationNano, retentionAmountNano)
+	}
+	return false
+}
+
 func getRunOnceReportPeriod(report *metering.Report) (*reportPeriod, error) {
 	if report.Spec.ReportingEnd == nil || report.Spec.ReportingStart == nil {
 		return nil, fmt.Errorf("run-once reports must have both ReportingEnd and ReportingStart")
@@ -867,6 +932,47 @@ func (op *Reporting) queueDependentReportsForReport(report *metering.Report) err
 		}
 	}
 	return nil
+}
+
+// for each report in the namespace, find ones that depend on the report passed into the function
+func (op *Reporting) reportsDependOnThisReport(report *metering.Report, reports []*metering.Report) (bool, string) {
+	var depReportName = ""
+	for _, otherReport := range reports {
+		deps, err := op.getReportDependencies(otherReport)
+		if err != nil {
+			op.logger.Errorf("was unable to get dependencies for report: %s, while looking for dependent reports", report.Name)
+			return true, depReportName
+		}
+		// If a Report has a dependency on the passed in report, we're done
+		for _, dep := range deps.Reports {
+			if dep.Name == report.Name {
+				depReportName = dep.Name
+				return true, depReportName
+			}
+		}
+	}
+	return false, depReportName
+}
+
+// Look for all ReportQueries in the namespace
+func (op *Reporting) reportQueriesDependOnThisReport(report *metering.Report, reportQueries []*metering.ReportQuery) (bool, string) {
+	var depQueryName = ""
+	// for each report in the namespace, find queries that depend on the report passed into the function.
+	for _, reportQuery := range reportQueries {
+		deps, err := op.getQueryDependencies(reportQuery.Namespace, reportQuery.Name, nil)
+		if err != nil {
+			op.logger.Errorf("was unable to get dependencies for report: %s, while looking for dependent queries")
+			return true, ""
+		}
+		// If this ReportQuery has a dependency on the passed in report, we're done
+		for _, dep := range deps.Reports {
+			if dep.Name == report.Name {
+				depQueryName = dep.Name
+				return true, depQueryName
+			}
+		}
+	}
+	return false, depQueryName
 }
 
 // queueDependentReportQueriesForReport will queue all
--- a/pkg/operator/reports.go
+++ b/pkg/operator/reports.go
@@ -936,7 +936,7 @@ func (op *Reporting) queueDependentReportsForReport(report *metering.Report) err
 
 // for each report in the namespace, find ones that depend on the report passed into the function
 func (op *Reporting) reportsDependOnThisReport(report *metering.Report, reports []*metering.Report) (bool, string) {
-	var depReportName = ""
+	var depReportName string
 	for _, otherReport := range reports {
 		deps, err := op.getReportDependencies(otherReport)
 		if err != nil {
@@ -946,17 +946,16 @@ func (op *Reporting) reportsDependOnThisReport(report *metering.Report, reports 
 		// If a Report has a dependency on the passed in report, we're done
 		for _, dep := range deps.Reports {
 			if dep.Name == report.Name {
-				depReportName = dep.Name
-				return true, depReportName
+				return true, dep.Name
 			}
 		}
 	}
 	return false, depReportName
 }
 
-// Look for all ReportQueries in the namespace
+// for each ReportQuery in the namespace, find ones that depend on the report passed into the function
 func (op *Reporting) reportQueriesDependOnThisReport(report *metering.Report, reportQueries []*metering.ReportQuery) (bool, string) {
-	var depQueryName = ""
+	var depQueryName string
 	// for each report in the namespace, find queries that depend on the report passed into the function.
 	for _, reportQuery := range reportQueries {
 		deps, err := op.getQueryDependencies(reportQuery.Namespace, reportQuery.Name, nil)
@@ -967,8 +966,7 @@ func (op *Reporting) reportQueriesDependOnThisReport(report *metering.Report, re
 		// If this ReportQuery has a dependency on the passed in report, we're done
 		for _, dep := range deps.Reports {
 			if dep.Name == report.Name {
-				depQueryName = dep.Name
-				return true, depQueryName
+				return true, dep.Name
 			}
 		}
 	}
--- a/charts/metering-ansible-operator/templates/crds/report.crd.yaml
+++ b/charts/metering-ansible-operator/templates/crds/report.crd.yaml
@@ -52,6 +52,9 @@ spec:
               reportingEnd:
                 type: string
                 format: date-time
+              expiration:
+                type: string
+                format: duration
               runImmediately:
                 type: boolean
               overwriteExistingData:
--- a/manifests/deploy/openshift/metering-ansible-operator/report.crd.yaml
+++ b/manifests/deploy/openshift/metering-ansible-operator/report.crd.yaml
@@ -52,6 +52,9 @@ spec:
               reportingEnd:
                 type: string
                 format: date-time
+              expiration:
+                type: string
+                format: duration
               runImmediately:
                 type: boolean
               overwriteExistingData:
--- a/manifests/deploy/openshift/olm/bundle/4.6/report.crd.yaml
+++ b/manifests/deploy/openshift/olm/bundle/4.6/report.crd.yaml
@@ -52,6 +52,9 @@ spec:
               reportingEnd:
                 type: string
                 format: date-time
+              expiration:
+                type: string
+                format: duration
               runImmediately:
                 type: boolean
               overwriteExistingData:
--- a/manifests/deploy/upstream/metering-ansible-operator/report.crd.yaml
+++ b/manifests/deploy/upstream/metering-ansible-operator/report.crd.yaml
@@ -52,6 +52,9 @@ spec:
               reportingEnd:
                 type: string
                 format: date-time
+              expiration:
+                type: string
+                format: duration
               runImmediately:
                 type: boolean
               overwriteExistingData:
--- a/manifests/deploy/upstream/olm/bundle/4.6/report.crd.yaml
+++ b/manifests/deploy/upstream/olm/bundle/4.6/report.crd.yaml
@@ -52,6 +52,9 @@ spec:
               reportingEnd:
                 type: string
                 format: date-time
+              expiration:
+                type: string
+                format: duration
               runImmediately:
                 type: boolean
               overwriteExistingData:
--- a/pkg/apis/metering/v1/zz_generated.deepcopy.go
+++ b/pkg/apis/metering/v1/zz_generated.deepcopy.go
@@ -3057,6 +3057,11 @@ func (in *ReportSpec) DeepCopyInto(out *ReportSpec) {
 		*out = new(ReportSchedule)
 		(*in).DeepCopyInto(*out)
 	}
+	if in.Expiration != nil {
+		in, out := &in.Expiration, &out.Expiration
+		*out = new(metav1.Duration)
+		**out = **in
+	}
 	if in.ReportingStart != nil {
 		in, out := &in.ReportingStart, &out.ReportingStart
 		*out = (*in).DeepCopy()
--- a/pkg/operator/http_test.go
+++ b/pkg/operator/http_test.go
@@ -118,7 +118,7 @@ func TestAPIV1ReportsGet(t *testing.T) {
 	}{
 		"report-finished-no-results": {
 			reportName: testReportName,
-			report:     testhelpers.NewReport(testReportName, namespace, testQueryName, reportStart, reportEnd, metering.ReportStatus{}, nil, false),
+			report:     testhelpers.NewReport(testReportName, namespace, testQueryName, reportStart, reportEnd, metering.ReportStatus{}, nil, false, nil),
 			query: testhelpers.NewReportQuery(testQueryName, namespace, []metering.ReportQueryColumn{
 				{
 					Name: "timestamp",
@@ -146,7 +146,7 @@ func TestAPIV1ReportsGet(t *testing.T) {
 		},
 		"report-finished-with-results": {
 			reportName: testReportName,
-			report:     testhelpers.NewReport(testReportName, namespace, testQueryName, reportStart, reportEnd, metering.ReportStatus{}, nil, false),
+			report:     testhelpers.NewReport(testReportName, namespace, testQueryName, reportStart, reportEnd, metering.ReportStatus{}, nil, false, nil),
 			query: testhelpers.NewReportQuery(testQueryName, namespace, []metering.ReportQueryColumn{
 				{
 					Name: "timestamp",
@@ -187,7 +187,7 @@ func TestAPIV1ReportsGet(t *testing.T) {
 		},
 		"report-finished-db-errored": {
 			reportName: testReportName,
-			report:     testhelpers.NewReport(testReportName, namespace, testQueryName, reportStart, reportEnd, metering.ReportStatus{}, nil, false),
+			report:     testhelpers.NewReport(testReportName, namespace, testQueryName, reportStart, reportEnd, metering.ReportStatus{}, nil, false, nil),
 			query: testhelpers.NewReportQuery(testQueryName, namespace, []metering.ReportQueryColumn{
 				{
 					Name: "timestamp",
@@ -232,7 +232,7 @@ func TestAPIV1ReportsGet(t *testing.T) {
 		},
 		"mismatched-results-schema-to-table-schema": {
 			reportName: testReportName,
-			report:     testhelpers.NewReport(testReportName, namespace, testQueryName, reportStart, reportEnd, metering.ReportStatus{}, nil, false),
+			report:     testhelpers.NewReport(testReportName, namespace, testQueryName, reportStart, reportEnd, metering.ReportStatus{}, nil, false, nil),
 			query: testhelpers.NewReportQuery(testQueryName, namespace, []metering.ReportQueryColumn{
 				{
 					Name: "timestamp",
@@ -394,7 +394,7 @@ func TestAPIV2ReportsFull(t *testing.T) {
 	}{
 		"report-finished-with-results": {
 			reportName: testReportName,
-			report:     testhelpers.NewReport(testReportName, namespace, testQueryName, reportStart, reportEnd, metering.ReportStatus{}, nil, false),
+			report:     testhelpers.NewReport(testReportName, namespace, testQueryName, reportStart, reportEnd, metering.ReportStatus{}, nil, false, nil),
 			apiPath:    apiReportV2URLFull(namespace, testReportName) + testFormat,
 			query: testhelpers.NewReportQuery(testQueryName, namespace, []metering.ReportQueryColumn{
 				{
@@ -444,7 +444,7 @@ func TestAPIV2ReportsFull(t *testing.T) {
 		},
 		"report-finished-no-results": {
 			reportName: testReportName,
-			report:     testhelpers.NewReport(testReportName, namespace, testQueryName, reportStart, reportEnd, metering.ReportStatus{}, nil, false),
+			report:     testhelpers.NewReport(testReportName, namespace, testQueryName, reportStart, reportEnd, metering.ReportStatus{}, nil, false, nil),
 			apiPath:    apiReportV2URLFull(namespace, testReportName) + testFormat,
 			query: testhelpers.NewReportQuery(testQueryName, namespace, []metering.ReportQueryColumn{
 				{
@@ -475,7 +475,7 @@ func TestAPIV2ReportsFull(t *testing.T) {
 		},
 		"report-finished-db-errored": {
 			reportName: testReportName,
-			report:     testhelpers.NewReport(testReportName, namespace, testQueryName, reportStart, reportEnd, metering.ReportStatus{}, nil, false),
+			report:     testhelpers.NewReport(testReportName, namespace, testQueryName, reportStart, reportEnd, metering.ReportStatus{}, nil, false, nil),
 			apiPath:    apiReportV2URLFull(namespace, testReportName) + testFormat,
 			query: testhelpers.NewReportQuery(testQueryName, namespace, []metering.ReportQueryColumn{
 				{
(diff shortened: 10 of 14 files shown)


pkg/operator/reports.go
```diff
@@ -734,6 +740,65 @@ func (op *Reporting) runReport(logger log.FieldLogger, report *metering.Report)
 	return nil
 }
 
+func deleteReport(logger log.FieldLogger, report *metering.Report, op *Reporting) {
+	if reportNotUsedAsInput(logger, report, op) {
+		logger.Infof("Deleted report after retention period: %s, past reportingEnd: %s",
+			report.Spec.Expiration, report.Spec.ReportingEnd)
+	}
+}
```
timflannagan1, line 748: We probably want to return at least an error here, and use the metering clientset to delete a list of Report custom resources that exist within a namespace.

pkg/operator/reports.go
```diff
@@ -734,6 +740,65 @@ func (op *Reporting) runReport(logger log.FieldLogger, report *metering.Report)
...
+
+func reportNotUsedAsInput(logger log.FieldLogger, report *metering.Report, op *Reporting) bool {
```
timflannagan1, line 750: Same here, I would assume we would return a boolean, declaring whether or not the `@report` has input dependencies and any error that may occur during the process of determining the value of that return type.

pkg/operator/reports.go
```diff
@@ -734,6 +740,65 @@ func (op *Reporting) runReport(logger log.FieldLogger, report *metering.Report)
...
+	// Consider Reports referencing this report in the namespace of this report
+	reports, err := op.reportLister.Reports(report.Namespace).List(labels.Everything())
+	if err != nil {
+		logger.Errorf("Unable to determine list of reports that might be input for report: %s", report.Name)
+		return false
+	}
+	for _, report := range reports {
```
timflannagan1, line 757: If the length of this list is zero, can we exit early?

pkg/operator/reports.go
```diff
@@ -859,6 +924,58 @@ func (op *Reporting) queueDependentReportsForReport(report *metering.Report) err
...
+	}
+	// true means no other reports depend on this report
+	return true
+}
+
+func (op *Reporting) doReportQueriesDependOnThisReport(reportQuery *metering.ReportQuery) bool {
+	// Look for all ReportQueries in the namespace
+	reportQueries, err := op.reportQueryLister.ReportQueries(reportQuery.Namespace).List(labels.Everything())
+	if err != nil {
+		log.Errorf("Was unable to list reports for report: %s, while looking for dependent queries")
```
timflannagan1, line 957: A couple of things here:
- Probably want to use the initialized `op.Logger` (or whatever the name of that field may be).
- You specified a format specification but didn't provide the corresponding variable.

pkg/apis/metering/v1/report.go
```diff
@@ -34,6 +34,9 @@ type ReportSpec struct {
 	// Schedule configures when the report runs.
 	Schedule *ReportSchedule `json:"schedule,omitempty"`
 
+	// Expiration of the report. Report will be deleted once this date has passed
+	Expiration *meta.Duration `json:"expiration,omitempty"`
+
```
timflannagan1, line 39: Anytime we add any fields to any of the CRD APIs, we need to mirror those changes to the corresponding CRD manifest (`charts/metering-ansible-operator/templates/crds/...`) and run `make metering-manifests`.

pkg/operator/reports.go
bentito: yeah didn't actually put in the delete action yet 🍅

bentito: okay, delete action in place now: 
in 099029d

bentito: Errors about problems in this new check-if-report-expired process are logged, but was thinking we don't want to stop the calling func, `runReport` from doing its thing just because we're unable to delete the report?

pkg/operator/reports.go
bentito: same answer as above

pkg/operator/reports.go
```diff
@@ -734,6 +740,73 @@ func (op *Reporting) runReport(logger log.FieldLogger, report *metering.Report)
 	return nil
 }
 
+func deleteReport(logger log.FieldLogger, report *metering.Report, op *Reporting) {
```
timflannagan1, line 743: Is this supposed to be a method of the `Reporting` structure? I see `op.deleteReport` referenced throughout, but I don't see that defined anywhere.

pkg/operator/reports.go
```diff
@@ -344,7 +344,13 @@ func getReportPeriod(now time.Time, logger log.FieldLogger, report *metering.Rep
 // according the report's schedule. If the next scheduled reporting period
 // hasn't elapsed, runReport will requeue the resource for a time when
 // the period has elapsed.
+// Report expiration is checked before running, report is deleted if expired.
 func (op *Reporting) runReport(logger log.FieldLogger, report *metering.Report) error {
+	// check if the report is expired
+	if reportExpired := isReportExpired(logger, report, time.Now()); reportExpired {
+		deleteReport(logger, report, op)
+		return nil
+	}
```
timflannagan1, line 353: I need to double-check how we currently process Report objects, but I would expect this kind of functionality to live higher up like in `syncReport` or `handleReport`.

pkg/operator/reports.go
bentito: I was modeling this off of the func and usage for `isReportFinished()` which seems to pass around the reference to `logger` that `runReport()` has itself been passed. Is this a bad model to follow?

bentito: I was modeling this off of the func and usage for `isReportFinished()` which seems to pass around the reference to `logger` that `runReport()` has itself been passed. Is this a bad model to follow?

bentito: however just changed them all to op.logger, which does seem to log properly, so...

pkg/apis/metering/v1/report.go
bentito: done in 
0c8e8d6

pkg/operator/reports.go
bentito: this PR only references this new `deleteReport()`, a function on the default class for the file `reports.go`, not sure how to refer to it. It's at the same location as for instance: `getReportPeriod()` function. Should it be moved to be a method of the Reporting type?

pkg/operator/reports.go
bentito: ok, let me know, I can lift it up as needed, if it's a bad place for it logically.

pkg/operator/reports.go
```diff
@@ -344,7 +344,15 @@ func getReportPeriod(now time.Time, logger log.FieldLogger, report *metering.Rep
...
 // the period has elapsed.
+// Report expiration is checked before running, report is deleted if expired.
 func (op *Reporting) runReport(logger log.FieldLogger, report *metering.Report) error {
+	// check if the report is expired
+	if reportExpired := isReportExpired(logger, report, time.Now()); reportExpired {
+		deleteReport(report, op)
+		return nil
+	} else {
+		op.logger.Infof("not deleting report: %s, it's not expired", report.Name)
+	}
```
timflannagan1, line 355: You can deflate this to only using an `if` statement as the `else` clause is redundant.

pkg/operator/reports.go
```diff
@@ -734,6 +742,73 @@ func (op *Reporting) runReport(logger log.FieldLogger, report *metering.Report)
 	return nil
 }
 
+func deleteReport(report *metering.Report, op *Reporting) {
```
timflannagan1, line 745: If you're passing `op* Reporting`, I would investigate making this a method of that type instead of a function that takes that type as an argument.

pkg/operator/reports.go
```diff
@@ -344,7 +344,15 @@ func getReportPeriod(now time.Time, logger log.FieldLogger, report *metering.Rep
 // according the report's schedule. If the next scheduled reporting period
 // hasn't elapsed, runReport will requeue the resource for a time when
 // the period has elapsed.
+// Report expiration is checked before running, report is deleted if expired.
 func (op *Reporting) runReport(logger log.FieldLogger, report *metering.Report) error {
+	// check if the report is expired
+	if reportExpired := isReportExpired(logger, report, time.Now()); reportExpired {
+		deleteReport(report, op)
+		return nil
+	} else {
```
timflannagan1, line 353: Hmm, I wonder if we should instead just set the Report's `metadata.DeletionTimestamp` instead, and leaving the deletion of this object to the updateFunc for the Report informer.

pkg/operator/reports.go
```diff
@@ -734,6 +742,73 @@ func (op *Reporting) runReport(logger log.FieldLogger, report *metering.Report)
...
+		if err != nil {
+			op.logger.WithError(err).Errorf("on deleting report: %s, for expiration, failed to list reports for namespace: %s",
+				report.Name, report.Namespace)
+		}
+		for _, report := range reports {
+			op.deleteReport(report)
+		}
+		op.logger.Infof("deleted report after retention period: %s, past reportingEnd: %s",
+			report.Spec.Expiration, report.Spec.ReportingEnd)
+	}
```
timflannagan1, line 757: I need to double-check how we're passing the `@report` object to this function, but I worry doing something like this will invalidate the cache and we should avoid deleting any object we're currently processing in this part of the reconciliation process.

timflannagan1: And just to clarify, marking the object as a deletion candidate is probably fine, with the intention that a higher-level function, like this object's updateFunc will take care of managing the deletion of that resource, but deleting the object outright may cause problems.

timflannagan1: Actually, @bparees (or @pweil-) do either of you have an opinion about this kind of pattern about deleting objects we're currently processing in the worker queue? It looks like we're passing a deep copy that the worker grabbed from the cache, so we might be all set in terms of invalidating the cache, but I would imagine we should just mark the object as a deletion candidate, quit early, and delete that object when it's re-queued.

bparees: i'm not positive i understand the scenario, but if this is a typical event/controller loop, i would expect you to not manipulate the informer cache directly, i would expect you to make a client api call to delete the resource itself(by name/namespace), which in turn will trigger an event back to your informer cache that will remove the entry from the cache (and let your controller loop receive an event that the resource is being deleted and handle that event as desired).

pweil-: The only thing I'd add to @bparees response is that I would argue that calling `deleteReport` from `runReport` is an unintended side effect and probably confusing.  Let `runReport` (and other methods) provide safety checks for bad input but if `runReport` actually runs, it only runs a report.  Otherwise it returns `nil` or errors but does not take further action.  Easier to test, easier to grok.

Let the orchestrator be responsible for state management and which method to call (`runReport`, `deleteReport`, `rescheduleReport`, etc) and let those method guard against bad input and execute.

pkg/operator/reports.go
bentito: yep moving to using op.deleteReport and removing the one that was added

pkg/operator/reports.go
bentito: yes, was for debug

pkg/operator/reports.go
bentito: I don't think that's how custom resources get deleted. But then again calling `op.deleteReport` isn't deleting them AND doesn't seem to be setting the `DeletionTimestamp`, so what do I know.

pkg/operator/reports.go
bentito: Moved to `syncReport()` and now uses `op.deleteReport()`

pkg/operator/reports.go
bentito: moved and using the op method now

pkg/operator/reports.go
```diff
@@ -734,6 +752,58 @@ func (op *Reporting) runReport(logger log.FieldLogger, report *metering.Report)
 	return nil
 }
 
+func isReportNotUsedAsInput(report *metering.Report, op *Reporting) bool {
```
bentito, line 755: I think we'll need to test this via an added e2e test.

pkg/operator/reports.go
```diff
@@ -91,8 +91,26 @@ func (op *Reporting) syncReport(logger log.FieldLogger, key string) error {
 		}
 		return err
 	}
+
 	sr := report.DeepCopy()
 
+	// check if the report is expired
+	if report.DeletionTimestamp == nil {
+		if reportExpired := isReportExpired(logger, report, time.Now()); reportExpired {
```
timflannagan1, line 99: I think we may need to check if the report has reached expiration in `runReport` too, like we check if a report has already finished:

```go
func (op *Reporting) runReport(logger log.FieldLogger, report *metering.Report) error {
	// check if report has reached its expiration date. If true, stop processing
        // processing the resource, and exit earliy.
	if reportExpired := isReportExpired(logger, report, time.Now()); reportExpired {
		return nil
	}
        ...
```

pkg/operator/reports.go
```diff
@@ -91,8 +91,26 @@ func (op *Reporting) syncReport(logger log.FieldLogger, key string) error {
...
+			if reportIsNotInput := isReportNotUsedAsInput(report, op); reportIsNotInput {
+				op.deleteReport(sr)
+				newDeletionTimestamp := metav1.Now()
+				report.SetDeletionTimestamp(&newDeletionTimestamp)
+				op.meteringClient.MeteringV1().Reports(report.Namespace).
+					Delete(context.TODO(), report.Name, metav1.DeleteOptions{})
```
timflannagan1, line 105: Hmm, this seems strange to me looking at the other `delete<resource>` methods we expose. I wonder if it makes sense to encapsulate this logic in another reporting method that's responsible for making the client delete call. Then, the `op.deleteReport` method call would call that method instead of `syncReport`.

pkg/operator/reports.go
bentito: Done in 4a78406b

pkg/operator/reports.go
bentito: This PR doesn't touch `op.deleteReport` so far, but what I noticed in practice is that `op.deleteReport` only seems to do work against the queue, making it so report is no longer enqueued if you call it. But it left the report installed in the cluster. Are you saying we should move the metering client delete (that actually remove the Report from the cluster into `op.deleteReport` such that that method become comprehensive? I think it makes sense, less error prone for the function user, just clarifying.

pkg/operator/reports.go
timflannagan1, line 99: I would just deflate this into a single conditional. Easier to read imo.

```suggestion
        // check if the report we're currently processing in the queue has reached the
        // configuring rention period.
	if report.DeletionTimestamp == nil && isReportExpired(logger, report, time.Now() {
```

timflannagan1: I guess you would lose the `op.logger.Warnf(...)` call, but there are probably better ways to structure this anyways.

pkg/operator/reports.go
bentito: Moved to `syncReport` and used existing `op.deleteReport` instead of creating a new deleteReport. The existing `op.deleteReport` deletes the report from the queues and then, yep, as Ben said, it was needed to call the client delete to delete the Report resource.

pkg/operator/reports.go
bentito: There are better ways to structure, but next pass...

pkg/operator/reports.go
bentito: MC-905 addresses this.

pkg/operator/reports.go
timflannagan1, line 105: Hmm, this looks like it's deleting the actual object we grabbed from the cache instead of the deep copy that was initialized on line 95.

pkg/operator/reports.go
timflannagan1, line 99: I think I would like to see a single call site for handling an expired Report. WDYT about doing something like the following:
```go
err := op.handleExpiredReport(logger, sr, time.Now())
if err != nil {
    return err
}
```
and that `op.handleExpiredReport` would essentially encapsulate the work you're doing here:
```
func (op *Reporting) handleExpiredReport(logger logrus.FieldLogger, report *metering.Report, processTime time.Time) error {
    if report.DeletionTimestamp != nil {
       return fmt.Errorf("report was already marked for deletion")
    }
    // handle case where the report has expired, log message and exit early
    if reportHasExpired := isReportExpired(logger, report, processTime); !reportExpired {
        logger.Infof("the %s Report in the %s namespace has not reached the expiration data", report.Name, report.Namespace)
        return nil
    }
    // we know the report is considered "expired" by now
    if reportCanBeDeleted := op.isReportNotUsedAsInput(report); reportCanBeDeleted {
        // do report deletion work here
        ...
```

timflannagan1: Also if `isReportNotUsedAsInput` passes a reporting structure, I would expect that function to migrate to a method on the reporting type or be restructured such that it uses dependency injection to get the fields in the reporting structure you care about.

timflannagan1: After looking at some of the reporting-operator codebase, it looks like we have support for resolving the dependencies for a Report: https://github.com/kube-reporting/metering-operator/blob/1f5c0686730008b8bafcbc602514eb3120c132cd/pkg/operator/reports.go#L295-L301

That method on that interface would then return the following type:
```go
type DependencyResolutionResult struct {
	Dependencies *ReportQueryDependencies
	InputValues  map[string]interface{}
}
```

pkg/operator/reports.go
```diff
@@ -859,6 +934,57 @@ func (op *Reporting) queueDependentReportsForReport(report *metering.Report) err
 	return nil
 }
 
+func (op *Reporting) doReportsDependOnThisReport(report *metering.Report) bool {
+	// Look for all reports in the namespace
+	reports, err := op.reportLister.Reports(report.Namespace).List(labels.Everything())
+	if err != nil {
+		op.logger.Errorf("was unable to list reports for report: %s, while looking for dependent reports", report.Name)
+		return false
+	}
```
timflannagan1, line 943: We already made the call to grab the list of reports at the call site. Can we just pass the list of reports we grab there as a parameter to this method?

pkg/operator/reports.go
```diff
@@ -859,6 +934,57 @@ func (op *Reporting) queueDependentReportsForReport(report *metering.Report) err
...
+		if err != nil {
+			op.logger.Errorf("was unable to get dependencies for report: %s, while looking for dependent reports", report.Name)
+			return false
+		}
+		// If a Report has a dependency on the passed in report, we're done
+		for _, dep := range deps.Reports {
+			if dep.Name == report.Name {
+				break
+			}
+		}
```
timflannagan1, line 957: Just trying to get some clarity in the situation where we're creating something like a monthly roll-up Report, which depends on a weekly scheduled Report as an input to that object, and the weekly scheduled Report has a configured retention period: if I'm reading this correctly, we essentially don't mark that object as "expired" as there's another Report (the monthly scheduled one) marked as a dependent. What happens in the case where the outer report doesn't have a retention period and the user hasn't explicitly deleted this object? It seems like the inner report will just keep getting marked as un-expired and re-processed in the queue.

timflannagan1: And I'm sure the immediate answer could be to make the documentation explicit enough to ensure that both the inner and outer report have a configured retention period but just wanted to see if I was reading that situation correctly.

timflannagan1: Maybe this could be a validation check too, e.g. if the outer report object depends on a report input, ensure that the outer report has a configured retention period if the report dependency does also?

pkg/operator/reports.go
```diff
@@ -734,6 +757,58 @@ func (op *Reporting) runReport(logger log.FieldLogger, report *metering.Report)
...
+	for _, report := range reports {
+		// If Report has no spec.inputs it can't be depending on this report.
+		if report.Spec.Inputs == nil {
+			continue
+		}
+
+		if op.doReportsDependOnThisReport(report) {
+			op.logger.Infof("reports exist that use report %s as input, will not delete though retention period has expired", report.Name)
+			return false
+		}
```
timflannagan1, line 776: Can we make this clear in the description of this function that we exit early if we detect that the Report we're currently processing has a dependency on another Report, we stop processing further and exit early? I missed that the first time around and assumed we would just then process the ReportQueries next, even if we detected a report dependency in the previous block.

pkg/operator/reports.go
bentito: `Reports().Delete` just takes the Name, so I don't think it matters?

pkg/operator/reports.go
bentito: yep

pkg/operator/reports.go
bentito: Not sure about the inner/outer terminology, but yes, it's easy to get in a situation where a Report won't be going away because something depends on it. We should document this.

On validation, probably. A Report added with an expiration period set which is depended on by a Report without one: I can't imagine a situation where this would be desired.

The only downside is the validation requires scanning the dependency tree of the Reports. From that scan it may be hard to clearly tell the user why the validation is failing, for instance, if it's failing validation because of two or more dependent Reports. But I suppose just knowing about the problem with any feedback is better than just wondering why the Report never goes away?

pkg/operator/reports.go
bentito: adding function comment

pkg/operator/reports.go
bentito: encapsulated into function on Reporting type, testing functionality now.

bentito: works again

pkg/operator/operator.go
```diff
@@ -811,6 +811,31 @@ func (op *Reporting) newPrometheusConn(promConfig promapi.Config) (prom.API, err
...
+		newDeletionTimestamp := metav1.Now()
+		report.SetDeletionTimestamp(&newDeletionTimestamp)
+		op.meteringClient.MeteringV1().Reports(report.Namespace).
+			Delete(context.TODO(), report.Name, metav1.DeleteOptions{})
+		op.logger.Infof("deleted report: %s, because expired", report.Name)
+		return nil
+	} else {
+		op.logger.Warnf("report: %s, would be deleted because expired, but is depended on", report.Name)
+		return nil
+	}
```
timflannagan1, line 836:
```suggestion
	}
	
	op.logger.Warnf("report: %s, would be deleted because expired, but is depended on", report.Name)
	return nil
```

pkg/operator/operator.go
```diff
@@ -811,6 +811,31 @@ func (op *Reporting) newPrometheusConn(promConfig promapi.Config) (prom.API, err
...
+	// check if the Report is past its retention time
+	if reportExpired := isReportExpired(op.logger, report, now); !reportExpired {
+		op.logger.Debugf("the %s Report in the %s namespace has not reached the expiration date", report.Name, report.Namespace)
+		return nil
+	}
+	// check if the Report is used by any other Report or ReportQuery, if not delete it
+	if reportIsNotInput := isReportNotUsedAsInput(report, op); reportIsNotInput {
+		op.deleteReport(report)
+		newDeletionTimestamp := metav1.Now()
+		report.SetDeletionTimestamp(&newDeletionTimestamp)
```
timflannagan1, line 828: Hmm, this seems like something that would get us in trouble. We shouldn't explicitly call `op.deleteReport(report)` as that event will get picked up when we make the client-go delete call anyways.

pkg/operator/operator.go
```diff
@@ -811,6 +811,31 @@ func (op *Reporting) newPrometheusConn(promConfig promapi.Config) (prom.API, err
...
+		op.meteringClient.MeteringV1().Reports(report.Namespace).
+			Delete(context.TODO(), report.Name, metav1.DeleteOptions{})
```
timflannagan1, line 830: We should definintely catch an error is the delete call fails.

```suggestion
		err := op.meteringClient.MeteringV1().Reports(report.Namespace).Delete(context.TODO(), report.Name, metav1.DeleteOptions{})
		if err != nil {
		    return err
		}
```

pkg/operator/operator.go
bentito: ✅

pkg/operator/operator.go
bentito: After some more looking and thinking, agreed.

pkg/operator/operator.go
bentito: ✅

pkg/operator/operator.go
```diff
@@ -811,6 +811,33 @@ func (op *Reporting) newPrometheusConn(promConfig promapi.Config) (prom.API, err
...
+	// check if the Report is used by any other Report or ReportQuery, if not delete it
+	if reportIsNotInput := isReportNotUsedAsInput(report, op); reportIsNotInput {
+		newDeletionTimestamp := metav1.Now()
+		report.SetDeletionTimestamp(&newDeletionTimestamp)
+		err := op.meteringClient.MeteringV1().Reports(report.Namespace).
+			Delete(context.TODO(), report.Name, metav1.DeleteOptions{})
+		if err != nil {
+			err = fmt.Errorf("should have deleted report: %s, because expired, but failed to delete", report.Name)
+			return err
+		}
```
timflannagan1, line 833: I realize we may want to amend this error check, and only fire an error when it's not of type `apierrrors.IsNotFound(err)` and handle that kind of error separately.

timflannagan1: That way, if we're processing a resource that's already been deleted, we can just exit early (or re-queue for later.)

pkg/operator/operator.go
```diff
@@ -7,6 +7,7 @@ import (
 	"database/sql"
 	"fmt"
 	"io/ioutil"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
```
timflannagan1, line 10: minor nit: I would move this to the kube-related imports. We're pretty consistent about this throughout the codebase.

pkg/operator/operator.go
```diff
@@ -811,6 +812,37 @@ func (op *Reporting) newPrometheusConn(promConfig promapi.Config) (prom.API, err
...
+		return nil
+	}
+	// check if the Report is used by any other Report or ReportQuery, if not delete it
+	if reportIsNotInput := isReportNotUsedAsInput(report, op); reportIsNotInput {
+		newDeletionTimestamp := metav1.Now()
+		report.SetDeletionTimestamp(&newDeletionTimestamp)
+		err := op.meteringClient.MeteringV1().Reports(report.Namespace).
+			Delete(context.TODO(), report.Name, metav1.DeleteOptions{})
+		if apierrors.IsNotFound(err) {
+			return nil
```
timflannagan1, line 832: I think we need to log a message here indicating that something potentially strange happened, whether that's another worker has deleted this resource already and we're processing it late.

pkg/operator/reports.go
```diff
@@ -734,6 +745,59 @@ func (op *Reporting) runReport(logger log.FieldLogger, report *metering.Report)
...
+	for _, reportQuery := range reportQueries {
+		// If ReportQuery has no spec.inputs it can't be depending on this report.
+		if reportQuery.Spec.Inputs == nil {
+			continue
+		}
+
+		if op.reportQueriesDependOnThisReport(reportQuery) {
+			op.logger.Infof("some ReportQueries exist that use report %s as input, will not delete though retention period has expired", report.Name)
+			return false
+		}
```
timflannagan1, line 782: Just thinking out loud here with roll-up reports in mind again. Reading this code again, it appears that if we automatically purge the monthly Report as the user-provided retention period has been met and that Report's inputs reference a weekly Report who also has a retention period configured. When we iterate over the list of Reports in the reporting-operator namespace, we'll see that it doesn't have any dependencies, but when we iterate of the list of ReportQueries, we would see one of those custom resources use that Report as an input, and we would mark the weekly Report as "expired, but has dependencies so not deleting". Is this right?

ading this code is what happens if we automatically purge the Report (monthly) whose inputs references another Report (weekly), we delete the monthyl Report that references that other Report, but we can't delete that other

timflannagan1: If so, that seems like bad UX, at least for the roll-up use-case.

pkg/operator/operator.go
bentito: cleaned up in 9be2b316

pkg/operator/operator.go
bentito: Fixed in 59e02b58

pkg/operator/operator.go
bentito: added INFO level logging about it in b12679f0

pkg/operator/reports.go
bentito: I got confused there, probably due to my lack of familiarity with roll-ups. I read rollup-reports.md, but will have to read it again to see what the answer is here...

bentito: ok so reading and thinking, it seems like the fix here is a follow-on task where we add retention times to ReportQuery so that they can be set to allow deletion of the sub-reports in the roll-up scenario?

pkg/operator/operator.go
```diff
@@ -811,6 +812,38 @@ func (op *Reporting) newPrometheusConn(promConfig promapi.Config) (prom.API, err
...
+		if err != nil {
+			err = fmt.Errorf("should have deleted report: %s, because expired, but failed to delete", report.Name)
+			return err
+		}
+		op.logger.Infof("deleted report: %s, because expired", report.Name)
+		return nil
+	} else {
+		op.logger.Warnf("report: %s, would be deleted because expired, but is depended on", report.Name)
+		return nil
+	}
```
timflannagan1, line 844: This is a redundent else clause
```suggestion
	op.logger.Warnf("report: %s, would be deleted because expired, but is depended on", report.Name)
	return nil
```

pkg/operator/operator.go
```diff
@@ -811,6 +812,38 @@ func (op *Reporting) newPrometheusConn(promConfig promapi.Config) (prom.API, err
...
+			Delete(context.TODO(), report.Name, metav1.DeleteOptions{})
+		if apierrors.IsNotFound(err) {
+			op.logger.Infof("report: %s, not deleted because it was not found at time delete attempted", report.Name)
+			return nil
+		}
+		if err != nil {
+			err = fmt.Errorf("should have deleted report: %s, because expired, but failed to delete", report.Name)
+			return err
+		}
+		op.logger.Infof("deleted report: %s, because expired", report.Name)
```
timflannagan1, line 839: I'd like to see a better log message here. Maybe something along the lines of `deleted the %s Report in the %s namespace as it reached the expiration date`.

pkg/operator/operator.go
```diff
@@ -811,6 +812,38 @@ func (op *Reporting) newPrometheusConn(promConfig promapi.Config) (prom.API, err
...
+	}
+	// check if the Report is past its retention time
+	if reportExpired := isReportExpired(op.logger, report, now); !reportExpired {
+		op.logger.Debugf("the %s Report in the %s namespace has not reached the expiration date", report.Name, report.Namespace)
+		return nil
+	}
+	// check if the Report is used by any other Report or ReportQuery, if not delete it
+	if reportIsNotInput := isReportNotUsedAsInput(report, op); reportIsNotInput {
+		newDeletionTimestamp := metav1.Now()
+		report.SetDeletionTimestamp(&newDeletionTimestamp)
```
timflannagan1, line 828: Hmm, I'm still not 100% sure we should be setting these values. Is it not sufficient to just to make the API call to delete this resource?

pkg/operator/operator.go
```diff
@@ -811,6 +812,38 @@ func (op *Reporting) newPrometheusConn(promConfig promapi.Config) (prom.API, err
...
+		err := op.meteringClient.MeteringV1().Reports(report.Namespace).
```
timflannagan1, line 829: style nit: this will fit on a single line

pkg/operator/operator.go
```diff
@@ -811,6 +812,38 @@ func (op *Reporting) newPrometheusConn(promConfig promapi.Config) (prom.API, err
 	return prom.NewAPI(client), nil
 }
 
+func (op *Reporting) handleExpiredReport(report *metering.Report, now time.Time) error {
+	// check if the Report is being deleted already
+	if report.DeletionTimestamp != nil {
+		return fmt.Errorf("report was already marked for deletion")
```
timflannagan1, line 818: Just thinking out loud here - is this the best way to handle this kind of event, i.e. should we be returning an error here, or should we just complain about it in the logs and return nil?

pkg/operator/reports.go
timflannagan1: Let's keep this implementation as-is for now, with the intention of documenting this limitation w.r.t. roll-up Reports + user-provided retention period.

pkg/operator/reports.go
```diff
@@ -859,6 +923,50 @@ func (op *Reporting) queueDependentReportsForReport(report *metering.Report) err
...
+	return true
+}
+
+func (op *Reporting) reportQueriesDependOnThisReport(reportQuery *metering.ReportQuery) bool {
+	// Look for all ReportQueries in the namespace
+	reportQueries, err := op.reportQueryLister.ReportQueries(reportQuery.Namespace).List(labels.Everything())
+	if err != nil {
+		op.logger.Errorf("was unable to list reports for report: %s, while looking for dependent queries")
+		return true
+	}
```
timflannagan1, line 951: I'm fine with making the call to list the ReportQueries in the same namespace of the `@reportQuery` parameter, but we handle this slightly different in the `op.doReportsDependOnThisReport` and pass both the target resource and the list of those resources in a namespace as parameters. I think I would prefer to make the call to list the resources we care about at the call site instead of in the helper method, but I'm fine with either implementation if we make it consistent.

pkg/operator/operator.go
```diff
@@ -811,6 +812,38 @@ func (op *Reporting) newPrometheusConn(promConfig promapi.Config) (prom.API, err
...
+	}
+	// check if the Report is past its retention time
+	if reportExpired := isReportExpired(op.logger, report, now); !reportExpired {
+		op.logger.Debugf("the %s Report in the %s namespace has not reached the expiration date", report.Name, report.Namespace)
```
timflannagan1, line 822:
```suggestion
		op.logger.Debugf("the %s Report in the %s namespace has not yet reached the expiration date", report.Name, report.Namespace)
```

pkg/operator/operator.go
```diff
@@ -811,6 +812,38 @@ func (op *Reporting) newPrometheusConn(promConfig promapi.Config) (prom.API, err
...
+		report.SetDeletionTimestamp(&newDeletionTimestamp)
+		err := op.meteringClient.MeteringV1().Reports(report.Namespace).
+			Delete(context.TODO(), report.Name, metav1.DeleteOptions{})
+		if apierrors.IsNotFound(err) {
+			op.logger.Infof("report: %s, not deleted because it was not found at time delete attempted", report.Name)
+			return nil
+		}
+		if err != nil {
+			err = fmt.Errorf("should have deleted report: %s, because expired, but failed to delete", report.Name)
+			return err
```
timflannagan1, line 837:
```suggestion
			return fmt.Errorf("should have deleted report: %s, because expired, but failed to delete", report.Name)
```

timflannagan1: And follow-up suggestion, updating the log message that gets returned:
```suggestion
			return fmt.Errorf("failed to deleted the expired %s report: %v", report.Name)
			return err
```

pkg/operator/operator.go
```diff
@@ -811,6 +812,38 @@ func (op *Reporting) newPrometheusConn(promConfig promapi.Config) (prom.API, err
...
+		}
+		op.logger.Infof("deleted report: %s, because expired", report.Name)
+		return nil
+	} else {
+		op.logger.Warnf("report: %s, would be deleted because expired, but is depended on", report.Name)
```
timflannagan1, line 842: Something that would be nice, but might also be _slightly_ out of scope for this PR would be firing off an event on top of logging this situation, such that a user can see this event message in many forms (and events are more consumable in my opinion).

timflannagan1: And to do that kind of thing, you would just invoke the event recorder type in the reporting structure:
```go
op.eventRecorder.Event(report, v1.EventTypeWarning, "ExpiredReportHasDependencies", "Skipping the deletion of the %s Report as other resources are dependent on it, despite reaching the desired expiration date.")
```

pkg/operator/operator.go
timflannagan1, line 839: Same as below - it would be pretty nice firing off an event that broadcasted the deletion of this report to a wider audience, on top of the reporting-operator container logging:
```go
op.eventRecorder.Event(report, v1.EventTypeNormal, "ExpiredReportHasBeenDeleted", "Deleted the %s Report as the configured expiration date has passed")
```

pkg/operator/reports.go
timflannagan1: Yeah agreed, it's annoying now knowing the totality of the situation, but we should aim towards at least providing some debugging mechanisms like container logs and events that can help portray the state these objects are in, why they aren't being deleted despite reaching the expiration date, etc.

pkg/operator/operator.go
bentito: fair. I added a comment to clarify the logic at that point though w/out the else.

pkg/operator/reports.go
bentito: yep, so towards that end, added logging on the thing depended on when we fail to delete report for dependencies. Done in bf132bc2

pkg/operator/operator.go
bentito: will do

pkg/operator/operator.go
bentito: Not without going over 120 chars, prefer to generally make the code read vertically without need for horizontal scrolling on a laptop screen. That was my thinking anyway.

pkg/operator/operator.go
bentito: https://kubernetes.io/docs/reference/using-api/api-concepts/ it's an async process setting it begins the process of gettting rid of the resource with all the finalizers being allowed to act once you set it. In this case I think it's okay to set and then have the client delete the resource because we actually check it ourselves at the top of this function as a way to prevent wasting time reattempting the delete.

pkg/operator/operator.go
bentito: sounds good. done.

pkg/operator/reports.go
bentito: agreed fixed up now.

pkg/operator/operator.go
bentito: done

pkg/operator/operator.go
bentito: yep, changed to along those lines.

pkg/operator/operator.go
bentito: thanks for that, changed with the suggestion.

pkg/operator/operator.go
bentito: thanks for that, changed with the suggestion

pkg/operator/reports.go
```diff
@@ -869,6 +934,47 @@ func (op *Reporting) queueDependentReportsForReport(report *metering.Report) err
...
+				return true, depReportName
+			}
+		}
+	}
+	return false, depReportName
+}
+
+// Look for all ReportQueries in the namespace
+func (op *Reporting) reportQueriesDependOnThisReport(report *metering.Report, reportQueries []*metering.ReportQuery) (bool, string) {
+	var depQueryName = ""
```
timflannagan1, line 959: This is not needed as strings default to `""` as their empty value.
```suggestion
	var depQueryName string
```

pkg/operator/reports.go
```diff
@@ -869,6 +934,47 @@ func (op *Reporting) queueDependentReportsForReport(report *metering.Report) err
...
+		deps, err := op.getReportDependencies(otherReport)
+		if err != nil {
+			op.logger.Errorf("was unable to get dependencies for report: %s, while looking for dependent reports", report.Name)
+			return true, depReportName
+		}
+		// If a Report has a dependency on the passed in report, we're done
+		for _, dep := range deps.Reports {
+			if dep.Name == report.Name {
+				depReportName = dep.Name
+				return true, depReportName
```
timflannagan1, line 950:
```suggestion
				return true, dep.Name
```

pkg/operator/reports.go
```diff
@@ -869,6 +934,47 @@ func (op *Reporting) queueDependentReportsForReport(report *metering.Report) err
...
+		deps, err := op.getQueryDependencies(reportQuery.Namespace, reportQuery.Name, nil)
+		if err != nil {
+			op.logger.Errorf("was unable to get dependencies for report: %s, while looking for dependent queries")
+			return true, ""
+		}
+		// If this ReportQuery has a dependency on the passed in report, we're done
+		for _, dep := range deps.Reports {
+			if dep.Name == report.Name {
+				depQueryName = dep.Name
+				return true, depQueryName
```
timflannagan1, line 971: Same here, no need to initialize the `depQueryName` when we can just return the `dep.Name` outright.

pkg/operator/reports.go
bentito: yep, done.

pkg/operator/reports.go
bentito: agreed, done

PR #1299, opened 2020-07-10
Expose a PodDisruptionBudget configuration for the Metering operands.

timflannagan1:
This updates the MeteringConfig CRD schema and API, exposing a configuration for managing [PodDisruptionBudget resources](https://kubernetes.io/docs/tasks/run-application/configure-pdb/) for the operand Pods. By default, this is enabled for all of the operands, except in the case where there are no Presto worker replicas. In the case where a user may need to upgrade a node and an operand's PDB resource is blocking a node from draining, you have the ability to disable that resource and re-enable once the upgrade process is complete.

For example, disabling the PDB for the reporting-operator would require editing the MeteringConfig custom resource:
```yaml
kind: MeteringConfig
...
spec:
...
  reporting-operator:
    spec:
      podDisruptionBudget:
        enabled: false
...
```

Once that custom resource has been edited, the metering-ansible-operator should purge that resource during it's next reconciliation.

I also added an e2e test for ensuring the correct number of PDB resources gets created, which is probably a good candidate for molecule testing the metering-ansible-operator if we had that testing infrastructure set up. We probably need to beef this up long term to test reconciling non-default configurations, e.g. ensuring the `enabled: false` results in a deleted PDB(s).


test/e2e/ensure_pod_disruption_budgets_exist_test.go
```diff
@@ -0,0 +1,78 @@
...
+		{
+			Name:         "hive-metastore-pdb",
+			Exists:       true,
+			MinAvailable: 1,
+		},
+		{
+			Name:         "hive-server-pdb",
+			Exists:       true,
+			MinAvailable: 1,
+		},
```
timflannagan1, line 45: We probably want to do something a bit more sophisticated than this, like read the MeteringConfig for the correct values, but this is sufficient for now.

bentito:
Did someone already require that Metering is an application such that: ```You are the owner of an application running on a Kubernetes cluster that requires high availability.```

charts/openshift-metering/values.yaml
```diff
@@ -639,6 +647,10 @@ presto:
       terminationGracePeriodSeconds: 30
       tolerations: []
 
+      podDisruptionBudget:
+        minAvailable: 0
```
bentito, line 651: I think something is off, either this number or the descriptive text, because all the text talks about minimum of one.

timflannagan1: That's probably a poor copy-paste job. We don't need to create a PDB resource for a resource that doesn't have any replicas, which is the default configuration for the Presto workers. That said, I don't think we would need to configure that kind of resource for non-coordinator resources anyways.

test/e2e/ensure_pod_disruption_budgets_exist_test.go
bentito: Can an e2e test drain a node and then check for continued functionality of the operand that was on that node?

bentito:
/lgtm
Maybe for another story just wondering about that comment about e2e test on draining a node and seeing continued functionality.

timflannagan1:
I want to think whether this kind of configuration makes sense some more before we tag it in.
/hold

test/e2e/ensure_pod_disruption_budgets_exist_test.go
timflannagan1: That should be possible in theory as we're using a kubeadmin kube client for the testing suite. Something we may want to do instead (or supplemental to your suggestion) is to patch the MeteringConfig custom resource, disabling the PDBs for all of the operands, and ensuring those resources are properly deleted by the operator.

PR #1307, opened 2020-07-13
hack,test: Re-enable upgrade testing.

timflannagan1:
During the process of testing upgrades, we first install Metering through OLM using a supported, released version of the operator through the redhat-operators CatalogSource. With these changes, we can more freely select the base channel in which we install Metering.

Before the recent changes made to the e2e suite that enabled all Metering tests to be deployed through OLM, instead of the manual workflow, we needed to use a CatalogSource containing all of the operator's metadata to do upgrade testing.

Now, before running any top-level tests, we build up an image registry containing the most up-to-date (in the context of CI) metering-ansible-operator manifest bundle, and creating all the requisite resources to expose that bundle as a custom CatalogSource. If upgrade testing is enabled, we can expect that when we base installation tests have finished, this CatalogSource containing a newer payload will exist and we can edit the existing metering subscription to point to this CatalogSource.


timflannagan1:
There's a commit in there that ensures only the upgrade test will be run.
/hold

test/e2e/main_test.go
```diff
@@ -115,10 +117,10 @@ func testMainWrapper(m *testing.M) int {
 	// the default images in the metering CSV. In the case where
 	// some version of the `make e2e-dev` has been provided, skip
 	// the deletion the local registry resources and CatalogSource CR.
-	if meteringOperatorImageRepo != "" {
+	if meteringOperatorImageRepo != "" && meteringOperatorImageTag != "" {
```
timflannagan1, line 120: We have this kind of validation in place lower in the testing suite to ensure that a context's deployment doesn't use an empty repository or tag variable, but we need to ensure that we're not overriding any default images unless the repository and tag variables have been provided.

test/deployframework/context.go
```diff
@@ -311,30 +311,10 @@ func (ctx *DeployerCtx) Upgrade(packageName, repoVersion string, purgeReports, p
...
-
-	err = CreateCatalogSource(ctx.Logger, packageName, ctx.Namespace, packageName, ctx.OLMV1Alpha1Client)
-	if err != nil {
-		return nil, fmt.Errorf("failed to create the %s CatalogSource: %v", packageName, err)
-	}
-
-	err = VerifyCatalogSourcePod(ctx.Logger, ctx.Client, packageName, ctx.Namespace)
-	if err != nil {
-		return nil, fmt.Errorf("failed to verify the %s CatalogSource was successfully created: %v", packageName, err)
-	}
```
timflannagan1, line 332: We can delete all of this logic now as we don't need to spin up a CatalogSource that references a ConfigMap anymore and can instead reference an existing CatalogSource that uses the registry image provided to the e2e suite.

timflannagan1:
Removed that TODO commit 
/hold cancel

test/deployframework/helpers.go
```diff
@@ -445,7 +445,7 @@ func VerifyConfigMap(logger logrus.FieldLogger, client kubernetes.Interface, nam
 
 // UpdateExistingSubscription is a helper function responsible for upgrading an existing metering-ocp Subscription
 // to use the newest payload and verify that the Subscription object is reporting a successful upgrade status.
-func UpdateExistingSubscription(logger logrus.FieldLogger, client olmclientv1alpha1.OperatorsV1alpha1Interface, name, upgradeChannel, namespace string) error {
+func UpdateExistingSubscription(logger logrus.FieldLogger, client olmclientv1alpha1.OperatorsV1alpha1Interface, name, namespace, catalogSourceName, catalogSourceNamespace, upgradeSubscriptionChannel string) error {
```
bentito, line 448: All the parameters from `name` on are strings, it just seems odd to call out one and not the others.

test/e2e/metering_upgrade_tests.go
```diff
@@ -9,13 +9,19 @@ import (
...
-	metering "github.com/kube-reporting/metering-operator/pkg/apis/metering/v1"
 	"github.com/kube-reporting/metering-operator/test/reportingframework"
 	"github.com/kube-reporting/metering-operator/test/testhelpers"
 )
 
+const (
+	// TODO: support upgrading from a custom registry image instead
+	// of defaulting to the redhat-operators CatalogSource and waiting
+	// for GA-ed channels to be exposed.
+	upgradeFromCatalogSource          = "redhat-operators"
```
bentito, line 21: Couldn't these move out to a test resource so as not to be compiled in?

timflannagan1:
Looks like tide is stuck :(

PR #1309, opened 2020-07-14
Add an e2e test for validating a custom node selector configuration.

timflannagan1:
This is a common MeteringConfig custom resource configuration we document downstream and following the updates in #1310, we should be testing this configuration regularly to ensure it's still working. At the moment, we can make this a required test with the intention of offloading it eventually as a periodic test that's not run as part of the regular dev PR e2e.

### DIFF
--- a/hack/e2e.sh
+++ b/hack/e2e.sh
@@ -15,6 +15,14 @@ function cleanup() {
         echo "Removing namespaces with the 'name=${METERING_NAMESPACE}-metering-testing-ns' label"
         kubectl delete ns -l "name=${METERING_NAMESPACE}-metering-testing-ns" --wait=false || true
 
+        # Remove any testing labels that may have been added during tests
+        echo "Removing any testing labels that were added to the cluster's nodes"
+        nodes=( $(kubectl get nodes -l metering-node-testing-label="true" --no-headers | awk '{ print $1 }') )
+        for i in "${nodes[@]}"
+        do
+            kubectl label node "$i" metering-node-testing-label- 2>/dev/null
+        done
+
         # Note: the `openshift-marketplace` namespace is hardcoded for now until we have the need
         # to create the registry-related resources in another namespace (e.g. testing upstream manifests).
         echo "Removing the local registry resources with the 'name=${METERING_NAMESPACE}-metering-testing-ns' label"
--- a/test/e2e/custom_node_selector_tests.go
+++ b/test/e2e/custom_node_selector_tests.go
@@ -1,0 +1,80 @@
+package e2e
+
+import (
+	"context"
+	"strings"
+	"testing"
+
+	"github.com/kube-reporting/metering-operator/test/reportingframework"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func testNodeSelectorConfigurationWorks(t *testing.T, rf *reportingframework.ReportingFramework) {
+	// note: we already know the node selector configuration that
+	// was specified in the MeteringConfig YAML manifest.
+	expectedNodeSelector := map[string]string{
+		"node-role.kubernetes.io/worker": "",
+		"metering-node-testing-label":    "true",
+	}
+
+	tt := []struct {
+		Name                 string
+		PodNameContains      string
+		ExpectedNodeSelector map[string]string
+	}{
+		{
+			Name:                 "valid-presto-coordinator-node-selector",
+			PodNameContains:      "presto-coordinator-0",
+			ExpectedNodeSelector: expectedNodeSelector,
+		},
+		{
+			Name:                 "valid-hive-server-node-selector",
+			PodNameContains:      "hive-server-0",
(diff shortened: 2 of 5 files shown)


test/e2e/metering_manual_install_tests.go
```diff
@@ -181,3 +181,34 @@ func s3InstallFunc(ctx *deployframework.DeployerCtx) error {
...
+func customNodeSelectorFunc(ctx *deployframework.DeployerCtx) error {
+	nodes, err := ctx.Client.CoreV1().Nodes().List(context.Background(), metav1.ListOptions{
+		LabelSelector: "node-role.kubernetes.io/worker=",
+	})
+	if err != nil || len(nodes.Items) == 0 {
+		return fmt.Errorf("Failed to list the worker nodes in the cluster: %v", err)
+	}
+
+	nodeTestingLabelKey := "metering-node-testing-label"
+	for _, node := range nodes.Items {
```
timflannagan1, line 199: I wasn't sure if there was a better or cleaner way of doing this. Essentially this marks every worker node with the testing node label, which should be a safe operation and we attempt to remove this label after we spin down the e2e suite.

hack/e2e.sh
```diff
@@ -15,6 +15,14 @@ function cleanup() {
         echo "Removing namespaces with the 'name=${METERING_NAMESPACE}-metering-testing-ns' label"
         kubectl delete ns -l "name=${METERING_NAMESPACE}-metering-testing-ns" --wait=false || true
 
+        # Remove any testing labels that may have been added during tests
+        echo "Removing any testing labels that were added to the cluster's nodes"
+        nodes=( $(kubectl get nodes -l metering-node-testing-label="true" --no-headers | awk '{ print $1 }') )
+        for i in "${nodes[@]}"
+        do
+            kubectl label node "$i" metering-node-testing-label- 2>/dev/null
+        done
```
timflannagan1, line 24: This was added to the cleanup function that's run anytime the EXIT pseudo-signal has been encountered. Probably need to tinker with this more further down the line, as we're not always going to be running this node selector configuration test, but it should be sufficient for now.

test/e2e/custom_node_selector_tests.go
```diff
@@ -0,0 +1,80 @@
...
+	"strings"
+	"testing"
+
+	"github.com/kube-reporting/metering-operator/test/reportingframework"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func testNodeSelectorConfigurationWorks(t *testing.T, rf *reportingframework.ReportingFramework) {
```
timflannagan1, line 14: Open to suggestions on how to make this test a bit more robust. At the moment, we just verify that the node selector configuration that was specified in the MeteringConfig custom resource has propagated to the metering operand Pods. We could do this kind of test, and then patch that same MeteringConfig custom resource with the intention of removing those node selector configurations, waiting for those changes to reconcile, and validating those node selector configurations no longer exist.

test/e2e/metering_manual_install_tests.go
```diff
@@ -181,3 +181,34 @@ func s3InstallFunc(ctx *deployframework.DeployerCtx) error {
...
+		// requires indexing into this dictionary and adding the node testing
+		// label key and value.
+		node.Labels[nodeTestingLabelKey] = "true"
+
+		_, err = ctx.Client.CoreV1().Nodes().Update(context.Background(), &node, metav1.UpdateOptions{})
+		if err != nil {
+			return fmt.Errorf("Failed to apply the %s testing label to the %s node: %v", nodeTestingLabelKey, node.Name, err)
+		}
+		ctx.Logger.Infof("Labeled the %s node with the %s node label", node.Name, nodeTestingLabelKey)
+	}
```
timflannagan1, line 211: I wasn't sure if there was a better or cleaner way of doing this. Essentially this marks every worker node with the testing node label, which should be a safe operation and we attempt to remove this label after we spin down the e2e suite.

test/e2e/custom_node_selector_tests.go
bentito: This seems fine for now. It would have caught the case where the move to v1 was wiping settings.

test/e2e/testdata/meteringconfigs/node-selector-prometheus-importer-disabled.yaml
```diff
@@ -0,0 +1,94 @@
+apiVersion: metering.openshift.io/v1
```
bentito, line 1: So this is meant to be a Metering CR with as many options used as possible?

timflannagan1: It's just the common MeteringConfig we use throughout the test installs + a node selector configuration.

test/e2e/metering_manual_install_tests.go
EmilyM1: Does this work?

timflannagan1: Yes

PR #1311, opened 2020-07-16
add sharedStorage option to CI

EmilyM1:


test/testhelpers/nfs_storageclass.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: storage.k8s.io/v1
+kind: StorageClass
+metadata:
+  name: nfs-server
+provisioner: example.com/nfs
+parameters:
+  archiveOnDelete: "false"
+reclaimPolicy: Delete
+volumeBindingMode: Immediate
```
timflannagan1, line 9: It looks like Github UI is complaining about this file missing a trailing newline.

EmilyM1: Yeah, that's a new thing. Also there's lines modded in here in manual_install_tests that I did not change.

test/e2e/metering_manual_install_tests.go
```diff
@@ -13,7 +13,9 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/yaml"
 
+	metering "github.com/kube-reporting/metering-operator/pkg/apis/metering/v1"
```
timflannagan1, line 18: meteringv1

test/e2e/metering_manual_install_tests.go
```diff
@@ -104,6 +106,75 @@ func testManualMeteringInstall(
...
+	return "", fmt.Errorf("failed to decode the %s YAML service manifest: %v", filepathNFSStorageClass, err)
+}
+//  - Wait until the Service created has a populated spec.ClusterIP (poll.Wait functionality)
+//  - Decode the NFS PV YAML manifest
+
+err := deploy.DecodeYAMLManifestToObject(filepathNFSPersistentvolume, &nfs_persistentvolume)
+if err != nil {
+	return "", fmt.Errorf("failed to decode the %s YAML service manifest: %v", filepathNFSPersistentvolume, err)
+}
+
```
timflannagan1, line 177: This isn't going to compile.

test/e2e/metering_manual_install_tests.go
```diff
@@ -104,6 +106,75 @@ func testManualMeteringInstall(
...
+	if err != nil {
+		return nil, err
+	}
+
+	if mc == nil {
+		return nil, fmt.Errorf("error: the decoded MeteringConfig object is nil")
+	}
+
+	return mc, nil
+}
```
timflannagan1, line 127: You may need to rebase against master as this was moved to the `test/testhelpers` package.

test/e2e/metering_manual_install_tests.go
EmilyM1: Yeah, still WIP, the clusterIP polling has to occur too. Was mostly wondering about the use of `deploy.DecodeYAMLManifestToObject` with multiple filepaths.

test/e2e/metering_manual_install_tests.go
```diff
@@ -104,6 +106,75 @@ func testManualMeteringInstall(
...
+		},
+	}
+	_, err = ctx.Client.CoreV1().Namespaces().Create(context.Background(), n, metav1.CreateOptions{})
+	if err != nil {
+		return err
+	}
+	ctx.Logger.Debugf("Created the %s namespace", ctx.Namespace)
+}
+//decode and create yamls except for persistentvolume
+filepathNFSServer := "../testhelpers/nfs_server.yaml"
```
pruan-rht, line 153: not very fluent at all in Go, but in general we should use the absolute paths rather than relative

timflannagan1: Yeah agreed on using absolute paths. We expose the root directory in a test flag, so joining that path with the path to the NFS-related manifests is best.

test/e2e/metering_manual_install_tests.go
EmilyM1: I just rebased and this is still there.

test/e2e/metering_manual_install_tests.go
EmilyM1: Yeah, it's WIP, I am not even fully sure a filepath can be valid as a map string, but I'll be making absolute paths for sure.

test/e2e/metering_manual_install_tests.go
```diff
@@ -104,6 +106,76 @@ func testManualMeteringInstall(
...
+	_, err = ctx.Client.CoreV1().Namespaces().Create(context.Background(), n, metav1.CreateOptions{})
+	if err != nil {
+		return err
+	}
+	ctx.Logger.Debugf("Created the %s namespace", ctx.Namespace)
+}
+
+//decode and create yamls except for persistentvolume
+var nfsManifestFiles = map[string]string{
+"filepathNFSServer": "../testhelpers/nfs_server.yaml",
```
pruan-rht, line 155: Indentation issue

pruan-rht: still relative paths

test/e2e/metering_manual_install_tests.go
```diff
@@ -104,6 +106,76 @@ func testManualMeteringInstall(
...
+"filepathNFSService": "../testhelpers/nfs_service.yaml",
+"filepathNFSStorageClass": "../testhelpers/nfs_storageclass.yaml"
+}
+
+for manifest, filepaths := range nfsManifestFiles {
+
+err := deploy.DecodeYAMLManifestToObject(manifest, &serverobject)
```
pruan-rht, line 162: entire block is incorrectly indented.

test/e2e/metering_manual_install_tests.go
timflannagan1: Sure, that's valid:
```go
deploy.DecodeYAMLManifestToObject(filepath.join(repoDir, nfsManifestsDir, nfsServerManifestFilename), ...)
```

test/e2e/metering_manual_install_tests.go
pruan-rht: also, path is is still relative.

test/testhelpers/nfs_service.yaml
```diff
@@ -0,0 +1,16 @@
...
+spec:
+  ports:
+  - name: 2049-tcp
+    port: 2049
+    protocol: TCP
+    targetPort: 2049
+  selector:
+    role: nfs-server
+  sessionAffinity: None
+  type: ClusterIP
```
timflannagan1, line 16: Missing a trailing newline.

test/e2e/metering_manual_install_tests.go
```diff
@@ -107,6 +111,107 @@ func testManualMeteringInstall(
...
+// 	return clusterIP
+// }
+// 	return pollInternal(poller(interval, timeout), condition)
+// }
+
+//  - Decode the NFS PV YAML manifest
+//get a go object
+// after we get clusterIP, save it in a var, pvc.spec.nfs.server = val
+filepathNFSPersistentvolume := "test/testhelpers/nfs_persistentvolume.yaml"
+err := deploy.DecodeYAMLManifestToObject(filepathNFSPersistentvolume, &nfs_persistentvolume)
```
timflannagan1, line 183: Go-style standard is to use camel case instead of snake casing

timflannagan1: Where is `nfs_persistentvolume` defined?

test/e2e/metering_manual_install_tests.go
```diff
@@ -107,6 +111,95 @@ func testManualMeteringInstall(
...
 }
 
+func createNFSProvisioner(ctx *deployframework.DeployerCtx)(error) {
+// cteate namespace
+	ctx.TargetPodsCount = nonHDFSTargetPodCount
+
+	_, err := ctx.Client.CoreV1().Namespaces().Get(context.Background(), ctx.Namespace, metav1.GetOptions{})
+	if err != nil && !apierrors.IsNotFound(err) {
+	return nil
+}
```
timflannagan1, line 121: Indentation is off still

test/e2e/metering_manual_install_tests.go
```diff
@@ -107,6 +111,95 @@ func testManualMeteringInstall(
...
+				"name": ctx.Namespace + "-" + testingNamespaceLabel,
+			},
+		},
+	}
+	_, err = ctx.Client.CoreV1().Namespaces().Create(context.Background(), n, metav1.CreateOptions{})
+	if err != nil {
+		return err
+	}
+	ctx.Logger.Debugf("Created the %s namespace", ctx.Namespace)
+}
```
timflannagan1, line 136: I would just create the NFS provisioner in a separate namespace like `nfs` instead of the same namespace as the deployer context.

test/e2e/metering_manual_install_tests.go
```diff
@@ -107,6 +111,95 @@ func testManualMeteringInstall(
...
+
+//decode and create yamls except for persistentvolume
+	nfsManifestFiles := [3]string{
+ 	"/test/testhelpers/nfs_server.yaml",
+	"/test/testhelpers/nfs_service.yaml",
+ 	"/test/testhelpers/nfs_storageclass.yaml",
+	}
```
timflannagan1, line 143: This doesn't look it will compile
```suggestion
	nfsManifestFiles := []string{
 	    "/test/testhelpers/nfs_server.yaml",
	    "/test/testhelpers/nfs_service.yaml",
 	   "/test/testhelpers/nfs_storageclass.yaml",
	}
```

test/e2e/metering_manual_install_tests.go
```diff
@@ -107,6 +111,95 @@ func testManualMeteringInstall(
...
+
+for _, filepath := range nfsManifestFiles {
+// err = deploy.DecodeYAMLManifestToObject(filepath.join(repoDir, nfsManifestsDir, nfsServerManifestFilename), ...)
+
+	err := deploy.DecodeYAMLManifestToObject(manifest, &serverobject)
+	if err != nil {
+		return "", fmt.Errorf("failed to decode the %s YAML service manifest: %v", filepath, err)
+	}
+}
```
timflannagan1, line 152: While iterating over this array, you're decoding the current file path to the same object (`serverObject`), so you're overwriting the variable that contains the decoded object.

test/e2e/metering_manual_install_tests.go
```diff
@@ -107,6 +111,95 @@ func testManualMeteringInstall(
...
+	nfsManifestFiles := [3]string{
+ 	"/test/testhelpers/nfs_server.yaml",
+	"/test/testhelpers/nfs_service.yaml",
+ 	"/test/testhelpers/nfs_storageclass.yaml",
+	}
+
+for _, filepath := range nfsManifestFiles {
+// err = deploy.DecodeYAMLManifestToObject(filepath.join(repoDir, nfsManifestsDir, nfsServerManifestFilename), ...)
+
+	err := deploy.DecodeYAMLManifestToObject(manifest, &serverobject)
```
timflannagan1, line 148: The `serverobject` variable doesn't look like it's defined anywhere.

test/e2e/metering_manual_install_tests.go
```diff
@@ -107,6 +111,95 @@ func testManualMeteringInstall(
 	assert.NoError(t, err, "capturing logs and uninstalling metering should produce no error")
 }
 
+func createNFSProvisioner(ctx *deployframework.DeployerCtx)(error) {
```
timflannagan1, line 114:
```suggestion
func createNFSProvisioner(ctx *deployframework.DeployerCtx) (error) {
```

test/e2e/metering_manual_install_tests.go
```diff
@@ -107,6 +111,95 @@ func testManualMeteringInstall(
...
+	"/test/testhelpers/nfs_service.yaml",
+ 	"/test/testhelpers/nfs_storageclass.yaml",
+	}
+
+for _, filepath := range nfsManifestFiles {
+// err = deploy.DecodeYAMLManifestToObject(filepath.join(repoDir, nfsManifestsDir, nfsServerManifestFilename), ...)
+
+	err := deploy.DecodeYAMLManifestToObject(manifest, &serverobject)
+	if err != nil {
+		return "", fmt.Errorf("failed to decode the %s YAML service manifest: %v", filepath, err)
```
timflannagan1, line 150: This is the wrong return type looking at the function signature (only returns an error).

test/e2e/metering_manual_install_tests.go
```diff
@@ -107,6 +111,95 @@ func testManualMeteringInstall(
...
+	pvc, err := client.CoreV1.Services(ctx.Namespace).Get(content.TODO(),service.nfs-service, metav1.GetOptions{})
+// if theres an error don't poll
+	if err != nil {
+		return false, err
+	}
+// clusterIP is populated, stop polling
+	return pvc.Spec.nfs.server != "", nil })
+// no service exists, stop polling
+	if err != nil {
+		return "", fmt.ErrorF("still waiting for clusterIP of persistentvolume to populate. %v", err)
```
timflannagan1, line 166:
```suggestion
		return fmt.Errorf("still waiting for clusterIP of persistentvolume to populate. %v", err)
```

test/e2e/metering_manual_install_tests.go
```diff
@@ -107,6 +111,127 @@ func testManualMeteringInstall(
...
+	}
+		if nfsManifestItem.action == "CREATE" {
+			ctx.Client.CoreV1().Namespaces().Create(context.Background(nfs_server,nfs_service, nfs_storageclass), &nfs_server, &nfs_service, &nfs_storageclass, metav1.CreateOptions{})
+	}
+			if err != nil {
+				return err
+			}
+			ctx.Logger.Debugf("Created the %s nfs_server %s nfs_service %s nfs_storageclass", nfs_server, nfs_service, nfs_storageclass)
+	} //
+	 	if nfsManifestItem.action == GET {
```
timflannagan1, line 176:
```suggestion
	 	if nfsManifestItem.action == "GET" {
```

test/e2e/metering_manual_install_tests.go
```diff
@@ -107,6 +111,127 @@ func testManualMeteringInstall(
...
+		resourceFile: "/test/testhelpers/nfs_storageclass.yaml",
+		action: [2]string{"CREATE"}
+	}
+}
+
+
+for _, manifests := range nfsManifestFiles {
+// err = deploy.DecodeYAMLManifestToObject(filepath.join(repoDir, nfsManifestsDir, nfsServerManifestFilename), ...)
+
+	err := deploy.DecodeYAMLManifestToObject(manifest, &nfs_server, &nfs_service, &nfs_storageclass)
```
timflannagan1, line 164: This doesn't match the function signature of that helper function. You're currently passing three objects and a single manifest. That function expects a single manifest and object.

test/e2e/metering_manual_install_tests.go
```diff
@@ -107,6 +111,127 @@ func testManualMeteringInstall(
...
+				return err
+			}
+			ctx.Logger.Debugf("Created the %s nfs_server %s nfs_service %s nfs_storageclass", nfs_server, nfs_service, nfs_storageclass)
+	} //
+	 	if nfsManifestItem.action == GET {
+			_, err = ctx.Client.CoreV1().Service(ctx.Namespace).Get(context.Background(), "nfs-service", metav1.GetOptions{})
+			if err != nil && !apierrors.IsNotFound(err) {
+				return nil
+			}
+
```
timflannagan1, line 181: Hmm, this could be problematic, especially in the case where the resource already exists. It's better to try and get the resource first, check if it already exists, and if it doesn't, create it.

test/e2e/metering_manual_install_tests.go
```diff
@@ -107,6 +111,127 @@ func testManualMeteringInstall(
...
+		}
+
+
+//  - Wait for service to populate spec.ClusterIP (poll.Wait functionality)
+// poll has a closure that polls for the server IP to get populated
+err := wait.Poll(3*time.Second, 5*time.Minute, func () (done bool, err error) {
+
+	pvc, err := client.CoreV1.Services(ctx.Namespace).Get(content.TODO(),service.nfs-service, metav1.GetOptions{})
```
timflannagan1, line 189: This won't compile

test/e2e/metering_manual_install_tests.go
```diff
@@ -107,6 +108,141 @@ func testManualMeteringInstall(
...
+					"name": ctx.Namespace + "-" + testingNamespaceLabel,
+				},
+			},
+		}
+		_, err = ctx.Client.CoreV1().Namespaces().Create(context.Background(), n, metav1.CreateOptions{})
+		if err != nil {
+			return err
+		}
+		ctx.Logger.Debugf("Created the %s namespace", ctx.Namespace)
+	}
```
timflannagan1, line 133: I would throw this into a separate namespace, like `nfs` or something like that, or you need to update the `ctx.TargetPodCount = nonHDFSTargetPodCount + 1` to account for the NFS provisioner Pod that will be spun up here.

test/e2e/metering_manual_install_tests.go
```diff
@@ -107,6 +108,141 @@ func testManualMeteringInstall(
...
+			ctx.Logger.Debugf("Created the %s nfs_server %s nfs_service %s nfs_storageclass", nfsManifestItems["nfsServer"],nfsManifestItems["nfsService"].resourceFile, nfsManifestItems["nfsStorageclass"].resourceFile)
+			if action == "GET" {
+
+				foo, err := ctx.Client.CoreV1().Services(ctx.Namespace).Get(context.Background(), "nfs-service", metav1.GetOptions{})
+				if err != nil && !apierrors.IsNotFound(err) {
+					return nil
+				}
+			}
+		}
+	}
```
timflannagan1, line 197: I would highly suggest we shy away from doing this kind of pattern and making this more complex than it needs to be. Currently, this is pretty difficult to read, and we can cut down on the amount of code dedicated to creating these resources by doing something like the following:

```go
	var (
		deployment *appsv1.Deployment
		service *corev1.Service
		storageclass *storagev1.StorageClass
	)
	resources := map[string]interface{}{
		"testhelpers/manifests/nfs/deployment.yaml": deployment,
		"testhelpers/manifests/nfs/service.yaml": service,
		"testhelpers/manifests/nfs/storageclass.yaml": storageclass, 
	}
	
	for path, resource := range resources {
		/*
		Steps:
		1. GET the resource
		   a. if the resource exists, continue to next item
		   b. else, decode the @path YAML manifest to the @resource object
		2. Create the @resource
	}
...
```

test/e2e/metering_manual_install_tests.go
```diff
@@ -107,6 +108,141 @@ func testManualMeteringInstall(
...
+			return false, err
+		}
+		// clusterIP is populated, stop polling
+		return pvc.Spec.nfs.server != "", nil
+	})
+	// no service exists, stop polling
+	if err != nil {
+		return err
+	}
+	return pvc.Spec.nfs.ClusterIP, nil
```
timflannagan1, line 215: This won't compile

test/e2e/metering_manual_install_tests.go
```diff
@@ -107,6 +108,141 @@ func testManualMeteringInstall(
...
+				}
+			}
+		}
+	}
+
+	//  - Wait for service to populate spec.ClusterIP (poll.Wait functionality)
+	// poll has a closure that polls for the server IP to get populated
+	err := wait.Poll(3*time.Second, 5*time.Minute, func() (done bool, err error) {
+
+		pvc, err := ctx.Client.CoreV1().Services(ctx.Namespace).Get(context.TODO(), "nfs-service", metav1.GetOptions{})
```
timflannagan1, line 203: It looks like you're querying for the "nfs-service" Service in the `ctx.Namespace` namespace, but assigning to a variable called `pvc`.

timflannagan1: Also, you should query for this resource (`.CoreV1().Service(...).Get(...)`) before polling. That way, if the resource already exists and the spec.ClusterIP has already been populated, you don't need to poll and continue to ping the apiserver. If the spec.ClusterIP has not been populated yet, then you run the poll logic, overwriting the variable name you used for the first GET, which should solve the scoping problem of the poll conditional function.

test/e2e/metering_manual_install_tests.go
```diff
@@ -107,6 +108,141 @@ func testManualMeteringInstall(
...
+		// if theres an error don't poll
+		if err != nil {
+			return false, err
+		}
+		// clusterIP is populated, stop polling
+		return pvc.Spec.nfs.server != "", nil
+	})
+	// no service exists, stop polling
```
timflannagan1, line 211: This comment is misleading, but I would remove it anyways as it doesn't serve much purpose when reading the code.

test/e2e/metering_manual_install_tests.go
```diff
@@ -107,6 +108,149 @@ func testManualMeteringInstall(
...
+	var service, err = ctx.Client.CoreV1().Services(ctx.Namespace).Get(context.TODO(), "nfs-service", metav1.GetOptions{})
+
+	// if theres an error don't poll
+	if err != nil {
+		err := wait.Poll(3*time.Second, 5*time.Minute, func() (done bool, err error) {
+			service, err = ctx.Client.CoreV1().Services(ctx.Namespace).Get(context.TODO(), "nfs-service", metav1.GetOptions{})
+			return false, err
+		}
+		// clusterIP is populated, stop polling
+		return service.Spec.ClusterIP != "", nil
```
timflannagan1, line 194: You're returning this conditional outside of the wait.Poll conditional function.

test/e2e/metering_manual_install_tests.go
```diff
@@ -107,6 +108,149 @@ func testManualMeteringInstall(
...
+
+		// Update the NFS PV configuration w/ the populated spec.ClusterIP from the NFS Service
+		clusterIP := service.spec.nfs.server
+	}
+
+	_, err = ctx.Client.CoreV1().Namespaces().Create(context.Background(), &nfs_persistentvolume, metav1.CreateOptions{})
+	if err != nil {
+		return err
+	}
+	ctx.Logger.Debugf("Created the %s nfs_clusterIP", clusterIP)
```
timflannagan1, line 226: This can be cleaned up a bit as there's no need to decode the YAML manifest if we first get the NFS persistentvolume resource first, check if it exists, and if not, decode it, update the pvc.nfs.server to point to the svc.Spec.ClusterIP and create it. If it does exist, we don't need to do anything, at least in this current implementation.

test/e2e/metering_manual_install_tests.go
```diff
@@ -107,6 +108,149 @@ func testManualMeteringInstall(
...
+			},
+		}
+		_, err = ctx.Client.CoreV1().Namespaces().Create(context.Background(), namespace, metav1.CreateOptions{})
+		if err != nil {
+			return namespace, nil
+		}
+		ctx.Logger.Debugf("Created the %s namespace", ctx.Namespace)
+	}
+	return namespace, nil
+}
```
timflannagan1, line 252: This is fine to keep as a helper function as we can use it in other spots too, but we wouldn't want to pass an entire `deployframework.DeployerCtx` structure as a parameter as we're only interested in the actual name of the ctx.Namespace string. We would want to also modify the name of this, replacing `createNamespace` with something like `createTestingNamespace` and go from there.

test/e2e/metering_manual_install_tests.go
```diff
@@ -107,6 +108,149 @@ func testManualMeteringInstall(
 	assert.NoError(t, err, "capturing logs and uninstalling metering should produce no error")
 }
 
+func createNFSProvisioner(ctx *deployframework.DeployerCtx) error {
+
+	ctx.TargetPodsCount = nonHDFSTargetPodCount
```
timflannagan1, line 113: This needs to be updated to reflect the NFS provisioner Pod that will be deployed in the ctx.Namespace as `nonHDFSTargetPodCount` is set to an integer value of five, but we're going to be provisioning six pods in total in this testing namespace, so we risk marking the metering installation as "ready to run post-install tests" but something like the reporting-operator Pod won't be healthy yet, which would be problematic.

test/e2e/metering_manual_install_tests.go
EmilyM1: At this point it hasn't been decoded yet, but the 3 other manifests (server, service, storageclass) have - so I believe I do need to decode it. I think you are saying to still decode it, but within a conditional, which would make sense. Is that the thought here?

test/e2e/metering_manual_install_tests.go
EmilyM1: Fixed

test/e2e/metering_manual_install_tests.go
EmilyM1: Actually, there is not a way this could be created before because it would not have the clusterIP in it. What is a case where it would already be created?

test/e2e/metering_manual_install_tests.go
```diff
@@ -107,6 +108,160 @@ func testManualMeteringInstall(
...
+			}
+			ctx.Logger.Debugf("Created the %s nfs_server %s nfs_service %s nfs_storageclass", nfsManifestItems["nfsServer"], nfsManifestItems["nfsService"].resourceFile, nfsManifestItems["nfsStorageclass"].resourceFile)
+			if action == "GET" {
+				_, err := ctx.Client.CoreV1().Services(ctx.Namespace).Get(context.TODO(), "nfs-service", metav1.GetOptions{})
+				if err != nil && !apierrors.IsNotFound(err) {
+					return nil
+				}
+			}
+		}
+	}
```
timflannagan1, line 181: The whole idea behind using a loop is to handle the process of decoding and creating these manifests in a single iteration. Currently, you're making the call to decode several manifests inside a single loop iteration.

test/e2e/metering_manual_install_tests.go
```diff
@@ -3,22 +3,25 @@ package e2e
 import (
 	"context"
 	"fmt"
+	"github.com/kube-reporting/metering-operator/pkg/deploy"
+	appsv1 "k8s.io/api/apps/v1"
+	"k8s.io/api/storage/v1beta1"
```
timflannagan1, line 8: Is there a reason we can't use the v1 version of the storage API and need to use the v1beta1 versioning?
```suggestion
	storagev1beta1 "k8s.io/api/storage/v1beta1"
```

test/e2e/metering_manual_install_tests.go
```diff
@@ -107,6 +112,137 @@ func testManualMeteringInstall(
...
+			_, err = ctx.Client.CoreV1().Namespaces().Create(context.Background(), &nfsPersistentVolume, metav1.CreateOptions{})
+			if err != nil {
+				return err
+			}
+			ctx.Logger.Debugf("Created the %s nfs_clusterIP", clusterIP)
+
+			return err
+
+
+}
```
timflannagan1, line 217: I would shy away from this kind of implementation, and just try and emulate what we worked on offline for how to handle the other resources. Instead, decode the persistent volume into a Go variable, update that variable's `spec.NFS.Server = svc.Spec.ClusterIP` we got previously and pass that now updated object using the following client-go API call:

```go
	err = ctx.Client.CoreV1().PersistentVolumes().Create(context.TODO(), pv, metav1.CreateOptions{})
```

test/e2e/metering_manual_install_tests.go
EmilyM1: It threw an error when I tried.

test/e2e/metering_manual_install_tests.go
EmilyM1: Agree. Have simplified.

test/e2e/metering_manual_install_tests.go
timflannagan1: Can you expand on what error it threw? It looks like the `StorageClass` is a valid object in the v1 package: https://godoc.org/k8s.io/api/storage/v1#StorageClass

test/e2e/metering_manual_install_tests.go
```diff
@@ -107,6 +112,105 @@ func testManualMeteringInstall(
 	assert.NoError(t, err, "capturing logs and uninstalling metering should produce no error")
 }
 
+func createNFSProvisioner(ctx *deployframework.DeployerCtx) error {
+	//Waiting for Metering pods
+	ctx.TargetPodsCount = nonHDFSTargetPodCount
+
+	_, err := createTestingNamespace(ctx.Client, "nfs")
```
timflannagan1, line 119: I would recommend moving this as a constant variable somewhere as we use the "nfs" string throughout and this could help improve readability/maintenance by using a constant variable instead.

```go
const nfsTestingNamespace = "nfs"
```

test/e2e/metering_manual_install_tests.go
```diff
@@ -107,6 +112,105 @@ func testManualMeteringInstall(
...
+
+			return svc.Spec.ClusterIP != "", err
+
+		})
+
+	}
+
+	nfsFile := "/test/testhelpers/nfs_persistentvolume.yaml"
+	nfs_pv := &corev1.PersistentVolume{}
+	nfs_pv.Spec.NFS.Server = svc.Spec.ClusterIP
```
timflannagan1, line 179: This is the wrong order of operations here. Decode first, then update the `nfs_pv` structure once it's been populated by `deploy.DecodeYAMLManifestToObject`.

test/e2e/metering_manual_install_tests.go
```diff
@@ -107,6 +112,105 @@ func testManualMeteringInstall(
...
+	_, err = ctx.Client.CoreV1().PersistentVolumes().Create(context.TODO(), nfs_pv, metav1.CreateOptions{})
+	if err != nil && !apierrors.IsAlreadyExists(err) {
+		return err
+	}
+
+}// end of function
+
+func createTestingNamespace(client kubernetes.Interface, namespace string) (*corev1.Namespace, error) {
+	ns, err := client.CoreV1().Namespaces().Get(context.Background(), namespace, metav1.GetOptions{})
+	//Get(ctx context.Context, name string, opts metav1.GetOptions) (*v1.Namespace, error)
```
timflannagan1, line 193: Remove this comment

test/e2e/metering_manual_install_tests.go
```diff
@@ -107,6 +112,105 @@ func testManualMeteringInstall(
...
+	if err != nil && !apierrors.IsNotFound(err) {
+		return ns, err
```
timflannagan1, line 195: I think we should just return `return nil, err` for better readability in the case the GET call failed and it failed in a manner we weren't expecting.

test/testhelpers/nfs_persistentvolume.yaml
```diff
@@ -0,0 +1,17 @@
...
+  capacity:
+    storage: 5Gi
+  accessModes:
+  - ReadWriteMany
+  storageClassName: nfs-server
+  nfs:
+    path: "/"
+    server: ""
+  persistentVolumeReclaimPolicy: Delete
+
```
timflannagan1, line 17: Remove the extra trailing newline.

test/testhelpers/nfs_server.yaml
```diff
@@ -0,0 +1,23 @@
...
+          containerPort: 2049
+      securityContext:
+        privileged: true
+      volumeMounts:
+      - mountPath: "/mnt/data"
+        name: local
+  volumes:
+    - name: local
+      emptyDir: {}
+
```
timflannagan1, line 23: Remove the extra trailing newline.

test/e2e/metering_manual_install_tests.go
```diff
@@ -3,22 +3,27 @@ package e2e
 import (
 	"context"
 	"fmt"
+	"github.com/aws/aws-sdk-go/aws/client"
```
timflannagan1, line 6: Remove this import.

test/e2e/metering_manual_install_tests.go
```diff
@@ -3,22 +3,27 @@ package e2e
...
+	"github.com/kube-reporting/metering-operator/pkg/deploy"
+	appsv1 "k8s.io/api/apps/v1"
+	"k8s.io/api/storage/v1beta1"
+	apiextv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
```
timflannagan1, line 10: Remove this import.

test/e2e/metering_manual_install_tests.go
```diff
@@ -107,6 +112,105 @@ func testManualMeteringInstall(
...
+
+	files := map[string]string{
+		"pod":          "/test/testhelpers/nfs_server.yaml",
+		"service":      "/test/testhelpers/nfs_service.yaml",
+		"storageClass": "/test/testhelpers/nfs_storageclass.yaml",
+
+	}
+	server := &corev1.Pod{}
+	service := &corev1.Service{}
+	storageClass := &v1beta1.StorageClass{}
```
timflannagan1, line 132: After renaming the `k8s.io/api/storage/v1beta1` to `storagev1 "k8s.io/api/storage/v1`, use the following:
```suggestion
	storageClass := &storagev1.StorageClass{}`
```

test/e2e/metering_manual_install_tests.go
timflannagan1: Let's move this call to update the `.Spec.NFS.Server` value to after lines 181-183 (after making the call to decode the YAML manifest into the `nfs_pv` empty object).

test/e2e/metering_manual_install_tests.go
```diff
@@ -107,6 +112,105 @@ func testManualMeteringInstall(
...
+	nfs_pv.Spec.NFS.Server = svc.Spec.ClusterIP
+	//decode pvc
+	if err := deploy.DecodeYAMLManifestToObject(nfsFile, nfs_pv); err != nil {
+		return err
+	}
+	_, err = ctx.Client.CoreV1().PersistentVolumes().Create(context.TODO(), nfs_pv, metav1.CreateOptions{})
+	if err != nil && !apierrors.IsAlreadyExists(err) {
+		return err
+	}
+
```
timflannagan1, line 188:
```suggestion
return nil
```

test/e2e/main_test.go
```diff
@@ -344,6 +344,32 @@ func TestManualMeteringInstall(t *testing.T) {
 			},
 			MeteringConfigManifestFilename: "s3.yaml",
 		},
+		{
+			Name:                      "ValidateNFS-ReportDynamicInputData",
+			MeteringOperatorImageRepo: meteringOperatorImageRepo,
+			MeteringOperatorImageTag:  meteringOperatorImageTag,
+			Skip:                      !runAllInstallTests,
```
timflannagan1, line 351: Unfortunately, this needs to be set to `Skip: false` or else the e2e suite for this PR won't run this test. We may want to revert back to `Skip: !runAllInstallTests` once we get closer to cleaning this PR up.

test/e2e/main_test.go
```diff
@@ -344,6 +344,32 @@ func TestManualMeteringInstall(t *testing.T) {
...
+						"REPORTING_OPERATOR_PROMETHEUS_METRICS_IMPORTER_CHUNK_SIZE=5m",
+						"REPORTING_OPERATOR_PROMETHEUS_METRICS_IMPORTER_INTERVAL=5m",
+						"REPORTING_OPERATOR_PROMETHEUS_METRICS_IMPORTER_STEP_SIZE=60s",
+					},
+				},
+				{
+					Name:         "testEnsureS3BucketIsDeleted",
+					TestFunc:     testEnsureS3BucketIsDeleted,
+					ExtraEnvVars: []string{},
+				},
```
timflannagan1, line 369: Not need to run this post-install tests as no s3 bucket has been configured for metering.

test/e2e/main_test.go
```diff
@@ -344,6 +344,27 @@ func TestManualMeteringInstall(t *testing.T) {
 			},
 			MeteringConfigManifestFilename: "s3.yaml",
 		},
+		{
+			Name:                      "ValidateNFS-ReportDynamicInputData",
```
timflannagan1, line 348:
```suggestion
			Name:                      "ValidNFS-ReportDynamicInputData",
```

test/e2e/metering_manual_install_test.go
```diff
@@ -107,6 +110,105 @@ func testManualMeteringInstall(
...
+
+	_, err := createTestingNamespace(ctx.Client, "nfsTestingNamespace")
+	if err != nil {
+		return err
+	}
+
+	files := map[string]string{
+		"pod":          "/test/testhelpers/nfs_server.yaml",
+		"service":      "/test/testhelpers/nfs_service.yaml",
+		"storageClass": "/test/testhelpers/nfs_storageclass.yaml",
```
timflannagan1, line 125: This needs to be referencing the root of the directory. I'm not sure if the `deployframework.DeployerCtx` type has knowledge of the root directory, but the `deployframework.DeployFramework` would. Check if there's any existing fields in the ctx structure that points to the path of the testing resources.

test/testhelpers/nfs_persistentvolume.yaml
timflannagan1, line 17: Let's move all of these manifests to test/e2e/manifests/nfs and deflate the `nfs_*` prefix to just the resource name.

test/e2e/main_test.go
EmilyM1: This is your editor width.

test/deployframework/context.go
```diff
@@ -76,19 +76,7 @@ type DeployerCtx struct {
...
-	reportingOperatorImageRepo,
-	reportingOperatorImageTag,
-	catalogSourceName,
-	catalogSourceNamespace,
-	subscriptionChannel,
-	outputPath string,
-	extraLocalEnvVars []string,
-	spec metering.MeteringConfigSpec,
-) (*DeployerCtx, error) {
+func (df *DeployFramework) NewDeployerCtx(namespace, meteringOperatorImageRepo, meteringOperatorImageTag, reportingOperatorImageRepo, reportingOperatorImageTag, catalogSourceName, catalogSourceNamespace, subscriptionChannel, outputPath string, extraLocalEnvVars []string, spec metering.MeteringConfigSpec, nfsm *metering.MeteringConfig) (*DeployerCtx, error) {
```
pruan-rht, line 79: I would suggest to follow the format as before... I don't think PR reviewers would appreciate one long line like that.

test/e2e/main_test.go
timflannagan1: I'm not sure I follow. I was suggesting a change to the test name.

test/e2e/metering_manual_install_test.go
```diff
@@ -57,6 +60,9 @@ func testManualMeteringInstall(
 	mc, err := testhelpers.DecodeMeteringConfigManifest(repoPath, testMeteringConfigManifestsPath, manifestFilename)
 	require.NoError(t, err, "failed to successfully decode the YAML MeteringConfig manifest")
 
+	nfsm, err := testhelpers.DecodeMeteringConfigManifest(repoPath, testNFSManifestPath, manifestFilename)
+	require.NoError(t, err, "failed to successfully decode the YAML MeteringConfig manifest")
```
timflannagan1, line 64: Why do we need to make this call to decode the meteringconfig manifest twice? The MeteringConfig YAML manifest that uses NFS as a storage backend should be located in the same directory as the other testing MeteringConfig YAML manifests.

test/e2e/main_test.go
```diff
@@ -47,6 +47,7 @@ var (
 	postUpgradeTestDirName          = "post-upgrade"
 	gatherTestArtifactsScript       = "gather-test-install-artifacts.sh"
 	testMeteringConfigManifestsPath = "/test/e2e/manifests/meteringconfigs/"
+	testNFSManifestPath 			= "/test/e2e/manifest/nfs/"
```
timflannagan1, line 50: This is the wrong path. It's `test/e2e/manifests/nfs` and not `test/e2e/*manifest*/nfs`

test/e2e/metering_manual_install_test.go
```diff
@@ -107,6 +114,105 @@ func testManualMeteringInstall(
 	assert.NoError(t, err, "capturing logs and uninstalling metering should produce no error")
 }
 
+func createNFSProvisioner(ctx *deployframework.DeployerCtx) error {
+	//Waiting for Metering pods
+	ctx.TargetPodsCount = nonHDFSTargetPodCount
+
+	_, err := createTestingNamespace(ctx.Client, "nfsTestingNamespace")
```
timflannagan1, line 121: The `nfsTestingNamespace` is an invalid name for a namespace.

test/e2e/metering_manual_install_test.go
```diff
@@ -107,6 +114,105 @@ func testManualMeteringInstall(
...
+	files := map[string]string{
+		"pod":          "/test/e2e/manifest/nfs/nfs-server.yaml",
+		"service":      "/test/e2e/manifest/nfs/nfs-service.yaml",
+		"storageClass": "/test/e2e/manifest/nfs/nfs-storageclass.yaml",
+	}
+	server := &corev1.Pod{}
+	service := &corev1.Service{}
+	storageClass := &storagev1beta1.StorageClass{}
+	for name, file := range files {
+
```
timflannagan1, line 135: Let's just build up the path to these manifests here.
```suggestion
absFile := filepath.Join(repoPath, file)
```
And then we can use that `absFile` value in the call to deploy.DecodeYAMLManifestsToObject. For example:

```go
deploy.DecodeYAMLManifestToObject(absFile, server)
```

test/e2e/metering_manual_install_test.go
```diff
@@ -107,6 +114,105 @@ func testManualMeteringInstall(
...
+			svc, err = ctx.Client.CoreV1().Services(ctx.Namespace).Get(context.TODO(), "nfs-service", metav1.GetOptions{})
+
+			return svc.Spec.ClusterIP != "", err
+
+		})
+
+	}
+
+	nfsFile := "/test/e2e/manifest/nfs/nfs-persistentvolume.yaml"
```
timflannagan1, line 177: You're referencing an invalid path here. The directory name manifests, and not manifest. This also needs to be updated to reflect the absolute path that I mentioned during the loop iteration.

test/e2e/metering_manual_install_test.go
```diff
@@ -107,6 +114,105 @@ func testManualMeteringInstall(
...
+	nfs_pv := &corev1.PersistentVolume{}
```
timflannagan1, line 178: We shouldn't use snake casing for defining variables in Golang. Should always be camel cased: `nfsPV` is the direct substitution.

test/e2e/main_test.go
EmilyM1: Gotcha. Keeping as validate

timflannagan1: ValidateNFS doesn't make much sense here in the context of what we're trying to accomplish here. We're passing a valid NFS storage configuration, and as part of the pre-installation function, not validating anything.

EmilyM1: Validating the NFS Storage config works. Seems fine to me.

test/e2e/metering_manual_install_test.go
```diff
@@ -46,6 +49,7 @@ func testManualMeteringInstall(
 ) {
 	// create a directory used to store the @testCaseName container and resource logs
 	testCaseOutputBaseDir := filepath.Join(testOutputPath, testCaseName)
+	os.RemoveAll(testCaseOutputBaseDir)
```
timflannagan1, line 52: Revert - this isn't needed.

test/e2e/metering_manual_install_test.go
```diff
@@ -86,8 +90,8 @@ func testManualMeteringInstall(
 	}
 
 	rf, err := deployerCtx.Setup(deployerCtx.Deployer.InstallOLM, expectInstallErr)
-	canSafelyRunTest := testhelpers.AssertCanSafelyRunReportingTests(t, err, expectInstallErr, expectInstallErrMsg)
```
timflannagan1, line 89: Revert this change too.

test/e2e/metering_manual_install_test.go
```diff
@@ -107,6 +111,105 @@ func testManualMeteringInstall(
 	assert.NoError(t, err, "capturing logs and uninstalling metering should produce no error")
 }
 
+func createNFSProvisioner(ctx *deployframework.DeployerCtx) error {
+	//Waiting for Metering pods
+	ctx.TargetPodsCount = nonHDFSTargetPodCount
+
+	_, err := createTestingNamespace(ctx.Client, nfsTestingNamespace)
```
timflannagan1, line 118: :+1:

test/e2e/metering_manual_install_test.go
```diff
@@ -107,6 +111,105 @@ func testManualMeteringInstall(
...
+		}
+	}
+	// wait and Poll , get service
+	svc, err := ctx.Client.CoreV1().Services(nfsTestingNamespace).Get(context.TODO(), "nfs-service", metav1.GetOptions{})
+	if apierrors.IsNotFound(err) && svc.Spec.ClusterIP == "" {
+
+		err = wait.Poll(3*time.Second, 5*time.Minute, func() (bool, error) {
+
+			svc, err = ctx.Client.CoreV1().Services(ctx.Namespace).Get(context.TODO(), "nfs-service", metav1.GetOptions{})
+
```
timflannagan1, line 167: You missing the error handling for this client-go API call. If you're creating the service resource for the first time, you're risking an error down the line when attempting to grab the `svc.Spec.ClusterIP` value.

I would suggest doing something like the following:
```go
	svc, err = ctx.Client.CoreV1().Services(ctx.Namespace).Get(context.TODO(), "nfs-service", metav1.GetOptions{})
		return false, err
	}
	if err != nil {
		return false, nil
	}

	return svc.Spec.ClusterIP != "", nil
```

test/e2e/metering_manual_install_test.go
```diff
@@ -107,6 +111,105 @@ func testManualMeteringInstall(
...
+			return svc.Spec.ClusterIP != "", err
+
+		})
+
+	}
+
+	nfsFile := "persistentvolume.yaml"
+	nfsPv := &corev1.PersistentVolume{}
+	absFile := filepath.Join(repoPath, testNFSManifestPath, nfsFile)
```
timflannagan1, line 176: :+1:

test/e2e/metering_manual_install_test.go
```diff
@@ -107,6 +111,105 @@ func testManualMeteringInstall(
...
+	_, err = ctx.Client.CoreV1().PersistentVolumes().Create(context.TODO(), nfsPv, metav1.CreateOptions{})
+	if err != nil && !apierrors.IsAlreadyExists(err) {
+		return err
+	}
+	return nil
+}
+
+func createTestingNamespace(client kubernetes.Interface, namespace string) (*corev1.Namespace, error) {
+	ns, err := client.CoreV1().Namespaces().Get(context.Background(), namespace, metav1.GetOptions{})
+
```
timflannagan1, line 192: Remove this extra newline between the client-go call and the error handling.

test/e2e/metering_manual_install_test.go
EmilyM1: There's a test error that required it
        metering_manual_install_test.go:53: 
            	Error Trace:	metering_manual_install_test.go:53
            	            				main_test.go:386
            	            	mkdir ValidateNFS-ReportDynamicInputData: file exists
            	Test:       	TestManualMeteringInstall/ValidateNFS-ReportDynamicInputData
            	Messages:   	creating the test case output directory should produce no error

test/e2e/main_test.go
timflannagan1: I'm not going to hold up merging on this, but I don't see how validating making sense as a test name when there's no validation occurring during this test. We're just testing a valid NFS storage configuration.

test/e2e/metering_manual_install_test.go
EmilyM1: So this line removes that error when running single test

test/e2e/metering_manual_install_test.go
EmilyM1: Strange, I don't have that line subtracted. It is in the file. So going to leave alone

test/e2e/metering_manual_install_test.go
timflannagan1: You're most likely overriding the default tmp directory output location. If that's erroring for you, let's tackle fixing that outside of this PR.

timflannagan1: But we should revert for this PR or else we risk CI not being able to grab the testing artifacts properly, which is a big part of validating whether or not this e2e test is working properly if something errors post-install.

test/e2e/metering_manual_install_test.go
EmilyM1: This makes sense, but this code above is a bit off from what is in there.
The biggest difference is, that I have a long closure going on, which makes it difficult to see where the error handling should nest ...
I have a guess but would you mind showing this with lines 161 -172

test/e2e/metering_manual_install_test.go
EmilyM1: removed

test/e2e/metering_manual_install_test.go
```diff
@@ -107,6 +110,111 @@ func testManualMeteringInstall(
...
+	if err != nil || svc.Spec.ClusterIP == "" {
+		err = wait.Poll(3*time.Second, 5*time.Minute, func() (bool, error) {
+			svc, err = ctx.Client.CoreV1().Services(ctx.Namespace).Get(context.TODO(), "nfs-service", metav1.GetOptions{})
+			if apierrors.IsNotFound(err) {
+				return false, err
+			}
+			if err != nil {
+				return false, nil
+			}
+			return svc.Spec.ClusterIP != "" , nil
```
timflannagan1, line 173: As mentioned offline, you need to run `go fmt` on this as verify is complaining about `return svc.Spec.ClusterIP != "" ,` have a comma between the two return types.

test/e2e/metering_manual_install_test.go
```diff
@@ -107,6 +110,111 @@ func testManualMeteringInstall(
...
+	if err := deploy.DecodeYAMLManifestToObject(absFile, nfsPv); err != nil {
+		return err
+	}
+
+	nfsPv.Spec.NFS.Server = svc.Spec.ClusterIP
+	_, err = ctx.Client.CoreV1().PersistentVolumes().Create(context.TODO(), nfsPv, metav1.CreateOptions{})
+	if err != nil && !apierrors.IsAlreadyExists(err) {
+		return err
+	}
+	return nil
```
timflannagan1, line 193: As a follow-up, we should aim towards making this more idempotent.

test/e2e/metering_manual_install_test.go
```diff
@@ -107,6 +110,111 @@ func testManualMeteringInstall(
...
+				},
+			},
+		}
+		_, err = client.CoreV1().Namespaces().Create(context.Background(), ns, metav1.CreateOptions{})
+		if err != nil {
+			return ns, nil
+		}
+	}
+	return ns, nil
+}
```
timflannagan1, line 216: As a follow-up, we should move this towards a more centralized package as we do something similar in the other tests.

test/e2e/metering_manual_install_test.go
EmilyM1: Thanks!

test/e2e/manifests/meteringconfigs/nfs.yaml
```diff
@@ -0,0 +1,55 @@
...
+  storage:
+    type: hive
+    hive:
+      type: sharedPVC
+      sharedPVC:
+        createPVC: true
+        storageClass: nfs-server
+        size: 5Gi
+
+  reporting-operator:
```
bentito, line 17: Couldn't you pare this down by just going with all defaults below here?

test/e2e/metering_manual_install_test.go
bentito: I think Tim means the whitespace change v. master is not needed, the location of the empty line was changed.

test/e2e/metering_manual_install_test.go
```diff
@@ -107,6 +110,111 @@ func testManualMeteringInstall(
...
+				return err
+			}
+			ctx.Client.StorageV1beta1().StorageClasses().Create(context.TODO(), storageClass, metav1.CreateOptions{})
+			if err != nil && !apierrors.IsAlreadyExists(err) {
+				return err
+			}
+		}
+	}
+
+	svc, err := ctx.Client.CoreV1().Services(nfsTestingNamespace).Get(context.TODO(), "nfs-service", metav1.GetOptions{})
```
bentito, line 160: I think there is a reason for "pre" Get? If so, a line comment here as to why would be helpful

test/e2e/metering_manual_install_test.go
bentito: Agreed, generalized, this could go in `testhelpers`

test/e2e/metering_manual_install_test.go
bentito: Aside from the test output dir fail thing, wouldn't running this test over and over have the same results and effects?

test/e2e/metering_manual_install_test.go
timflannagan1: There's no guarantee that the service we created beforehand will have a populated `svc.Spec.ClusterIP` as that's populated by kube at runtime and not creation time. We could do something like `time.Sleep(3*time.Second)` before attempting to grab the service but that's a poor implementation, so we instead attempt to grab the service, check if that IP has been populated, and if not, continue pinging the apiserver until it has been populated.

test/e2e/manifests/meteringconfigs/nfs.yaml
timflannagan1: I would agree for this most part, except for the metrics importer configuration, but even that attempts to set a few default values, and we do this throughout most of the MeteringConfig YAML manifests anyways, and fixing that is out of the scope of this PR anyways.

test/e2e/metering_manual_install_test.go
timflannagan1: I realize the code block I'm highlighting here is for the PV setup, and idempotent was a poor choice of words - we can further refine this implementation by checking if these resources already exist, and if not, attempt to create the resource. I don't think we should care about the upgrade scenario anyways as PV's are immutable post creation and that just introduces extra work when tearing down the namespace is sufficient (which we do when exiting the e2e suite in non-dev mode) in most cases.

test/e2e/metering_manual_install_test.go
timflannagan1: But yeah agreed, we should probably reflect that behavior in a comment as that's a valid question.

timflannagan1:
@bentito I'm going to tag this. We can address anything else cosmetic as a follow-up as it's test code and nothing user-facing.
/lgtm
/approve

EmilyM1:
Can we bypass upgrades?

PR #1314, opened 2020-07-17
test/e2e: Add an e2e test for configuring a MySQL instance for Hive metastore.

timflannagan1:
In our OCP documentation, we recommend overriding the default, local embedded Derby database that Hive metastore is configured to use, with either a MySQL or PostgreSQL instance.
    
This would add a fairly straight-forward MeteringConfig YAML manifest that is configured to reference to a MySQL database instance in the `mysql` namespace. We create that namespace and the database instance using the `oc new-app ...` mysql 5.7 imagestream before firing off the metering installation responsible for using MySQL as an underlying hive metastore database.

### DIFF
--- a/test/e2e/main_test.go
+++ b/test/e2e/main_test.go
@@ -254,6 +254,27 @@ func TestManualMeteringInstall(t *testing.T) {
 			MeteringConfigManifestFilename: "node-selector-prometheus-importer-disabled.yaml",
 		},
 		{
+			Name:                      "ValidHDFS-MySQLDatabase",
+			MeteringOperatorImageRepo: meteringOperatorImageRepo,
+			MeteringOperatorImageTag:  meteringOperatorImageTag,
+			Skip:                      false,
+			PreInstallFunc:            createMySQLDatabase,
+			InstallSubTests: []InstallTestCase{
+				{
+					Name:     "testReportingProducesData",
+					TestFunc: testReportingProducesData,
+					ExtraEnvVars: []string{
+						"REPORTING_OPERATOR_PROMETHEUS_DATASOURCE_MAX_IMPORT_BACKFILL_DURATION=15m",
+						"REPORTING_OPERATOR_PROMETHEUS_METRICS_IMPORTER_INTERVAL=30s",
+						"REPORTING_OPERATOR_PROMETHEUS_METRICS_IMPORTER_CHUNK_SIZE=5m",
+						"REPORTING_OPERATOR_PROMETHEUS_METRICS_IMPORTER_INTERVAL=5m",
+						"REPORTING_OPERATOR_PROMETHEUS_METRICS_IMPORTER_STEP_SIZE=60s",
+					},
+				},
+			},
+			MeteringConfigManifestFilename: "mysql.yaml",
+		},
+		{
 			Name:                      "S3-ReportDynamicInputData",
 			MeteringOperatorImageRepo: meteringOperatorImageRepo,
 			MeteringOperatorImageTag:  meteringOperatorImageTag,
--- a/test/e2e/manifests/meteringconfigs/mysql.yaml
+++ b/test/e2e/manifests/meteringconfigs/mysql.yaml
@@ -1,0 +1,78 @@
+apiVersion: metering.openshift.io/v1
+kind: MeteringConfig
+metadata:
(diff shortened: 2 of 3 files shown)


test/e2e/main_test.go
```diff
@@ -253,6 +253,27 @@ func TestManualMeteringInstall(t *testing.T) {
...
+					ExtraEnvVars: []string{
+						"REPORTING_OPERATOR_PROMETHEUS_DATASOURCE_MAX_IMPORT_BACKFILL_DURATION=15m",
+						"REPORTING_OPERATOR_PROMETHEUS_METRICS_IMPORTER_INTERVAL=30s",
+						"REPORTING_OPERATOR_PROMETHEUS_METRICS_IMPORTER_CHUNK_SIZE=5m",
+						"REPORTING_OPERATOR_PROMETHEUS_METRICS_IMPORTER_INTERVAL=5m",
+						"REPORTING_OPERATOR_PROMETHEUS_METRICS_IMPORTER_STEP_SIZE=60s",
+					},
+				},
+			},
+			MeteringConfigManifestFilename: "mysql.yaml",
```
timflannagan1, line 275: Open to suggestions on any additional post-installation functions. I didn't see much value in adding a post-install function that aims to emulate something like the prometheus connector up test, as we should be able to tell if the Hive metastore + mysql connection is working with the `testReportingProducesData` sub-test.

test/e2e/main_test.go
```diff
@@ -253,6 +253,27 @@ func TestManualMeteringInstall(t *testing.T) {
 			},
 			MeteringConfigManifestFilename: "node-selector-prometheus-importer-disabled.yaml",
 		},
+		{
+			Name:                      "ValidHDFS-MySQLDatabase",
+			MeteringOperatorImageRepo: meteringOperatorImageRepo,
+			MeteringOperatorImageTag:  meteringOperatorImageTag,
+			Skip:                      false,
```
timflannagan1, line 260: Probably should change this to `!runAllInstallTests` in the future once we add some periodic testing support, but this is fine for now.

bentito:
Is there something special about the ancient version of MySQL?

test/e2e/manifests/meteringconfigs/mysql.yaml
```diff
@@ -0,0 +1,78 @@
+apiVersion: metering.openshift.io/v1
```
bentito, line 1: this is so much better, I'm really glad we do that is ingested files now.

test/e2e/metering_manual_install_tests.go
```diff
@@ -212,3 +215,76 @@ func customNodeSelectorFunc(ctx *deployframework.DeployerCtx) error {
...
+			return err
+		}
+		ctx.Logger.Debugf("Created the %s namespace", mysqlNamespace)
+	}
+
+	cmd := exec.Command(
+		"oc",
+		"-n", mysqlNamespace,
+		"new-app",
+		"--image-stream", "mysql:5.7",
```
bentito, line 252: Would this work a more recent MySQL version but it's not available in CI?

timflannagan1: Nope, we have [open BZ](https://bugzilla.redhat.com/show_bug.cgi?id=1838802) for Metering where we can't work with mysql8. This kind of e2e test would be nice in verifying that the proposed fix (updating the hadoop image to pull an updated mysql connector jar) would still work with mysql5.7, in addition to a more recent release.

timflannagan1:
Looks like that was an e2e flake, but the hive metastore e2e test case appears to be running okay. This is from the metastore logs:
```
20/07/21 15:33:49 [main]: INFO metastore.ObjectStore: Setting MetaStore object pin classes with hive.metastore.cache.pinobjtypes="Table,StorageDescriptor,SerDeInfo,Partition,Database,Type,FieldSchema,Order"
20/07/21 15:33:52 [main]: INFO metastore.MetaStoreDirectSql: Using direct SQL, underlying DB is MYSQL
20/07/21 15:33:52 [main]: INFO metastore.ObjectStore: Initialized ObjectStore
20/07/21 15:33:52 [main]: WARN metastore.ObjectStore: Version information not found in metastore. hive.metastore.schema.verification is not enabled so recording the schema version 2.3.0
```

PR #1317, opened 2020-07-19
hack: Gather the InstallPlan resources in the test metering namespace.

timflannagan1:


timflannagan1:
Welp, not entirely sure what's happening here. It looks like the test cluster is using a 1.18.x kubelet/node/apiserver, but this is still being rendered as a 1.17.0 server version. After some light investigation, it looks like OLM determines this using a discovery client, so that shouldn't be an issue. Probably need to ping DPTP for further investigations as I imagine we're not the only component getting hit by this.
```
- group: operators.coreos.com
    kind: ClusterServiceVersion
    message: 'CSV version requirement not met: minKubeVersion (1.18.0) > server version
      (1.17.0)'
    name: metering-operator.v4.6.0
    status: PresentNotSatisfied
    version: v1alpha1
```

timflannagan1:
Okay, looks like these recent e2e failures are due to a recent rebase with the openshift apiserver. Just going to force-merge this change for now as it's purely for test debugging.

PR #1318, opened 2020-07-19
charts,manifests: Bump the minimum kubernetes server version to 1.18.3

timflannagan1:
This is a follow-up PR to #1322 which lowered the `minKubeVersion` in our CSV's to unblock CI as the server was reporting a kube version of 1.17.0 instead of 1.18.3 at the time. It appears this is now fixed, so this reverts back to 1.18.3, which is the version our kube dependencies are currently pinned to now.

### DIFF
--- a/manifests/deploy/openshift/olm/bundle/4.6/meteringoperator.v4.6.0.clusterserviceversion.yaml
+++ b/manifests/deploy/openshift/olm/bundle/4.6/meteringoperator.v4.6.0.clusterserviceversion.yaml
@@ -257,7 +257,7 @@ spec:
 
   keywords: [metering metrics reporting prometheus chargeback]
   version: "4.6.0"
-  minKubeVersion: "1.17.0"
+  minKubeVersion: "1.18.3"
   maturity: stable
   maintainers:
     - email: sd-operator-metering@redhat.com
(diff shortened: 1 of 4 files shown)


timflannagan1:
Need to wait for the apiserver fixes before we can merge this as it will cause CI to break again.
/hold

timflannagan1:
This appears to be resolved now as e2e is passing again.
/hold cancel

manifests/deploy/openshift/olm/bundle/4.6/meteringoperator.v4.6.0.clusterserviceversion.yaml
```diff
@@ -253,7 +253,7 @@ spec:
 
   keywords: [metering metrics reporting prometheus chargeback]
   version: "4.6.0"
-  minKubeVersion: "1.17.0"
+  minKubeVersion: "1.18.3"
```
bentito, line 256: We're setting a different patch version here: https://github.com/kube-reporting/metering-operator/blob/e3165873ffda515bf641da2371122f4c2ae41187/charts/metering-ansible-operator/upstream-values.yaml#L50 and in one other place.

timflannagan1: Yeah, good catch. Updated that chart and re-made the manifests in the last force-push.

PR #1319, opened 2020-07-20
Documentation,charts,manifests: Update the downstream logo.

timflannagan1:
This updates the various values.yaml helm charts that are used to generate the YAML manifests.

In order to generate this icon binary, I ran the following command:

```bash
$ base64 -i -w 0 Documentation/downstream_metering_logo.svg
```

And I verified that was working properly but running the following:

```bash
$ output=$(base64 -i -w 0 Documentation/downstream_metering_logo.svg)
$ echo $output | base64 -d | display
```

Note(s):
- Added a csv.icon configuration for the upstream-values.yaml to distinguish between the upstream and downstream logos, as before they pointed to the same logo.
- While testing, I rename the package.yaml we use to avoid conflicting with the metering-ocp packagemanifest that gets loaded in the redhat-operators CatalogSource. It looks like the console attempts to load the first metering-ocp packagemanifest that gets returned when I was poking around chrome's element inspection view.

Screenshot from 2020-07-20 17-46-54


timflannagan1:
Let's hold off on merging this until Sergio is back on PTO.
/hold

charts/metering-ansible-operator/upstream-values.yaml
```diff
@@ -80,6 +80,10 @@ olm:
 
       * **Configure AWS Billing Data Source** - To assign Pod $$ costs on AWS, create a [read-only IAM user](https://github.com/kube-reporting/metering-operator/blob/master/Documentation/configuring-aws-billing.md) ([example-policy](https://github.com/kube-reporting/metering-operator/blob/master/Documentation/aws/read-only.json)) and [configure Metering](https://github.com/kube-reporting/metering-operator/blob/master/manifests/metering-config/aws-billing.yaml) to use it.
 
+    icon:
```
bentito, line 83: We should put the commit message into perhaps its own `dev-notes.txt` doc in https://github.com/kube-reporting/setup perhaps? Or somewhere devs can find it to be able to repeat this rarely done technique.

timflannagan1: Yeah that's not a bad idea, but it should be sufficient for now to just refer back to the PR description/commit message for this and emulate the steps listed there.

PR #1322, opened 2020-07-21
charts,manifests: Lower the minKubeVersion to 1.17.0 to unblock CI.

timflannagan1:
Right now, every PR is failing e2e as the minKubeVersion specified in our CSV resource is 1.18.3, which is greater than the server version (1.17.0) that the apiserver is advertising. In order to unblock CI, we can lower this value to 1.17.0 with the intention of reverting it once this issue is resolved.

### DIFF
--- a/manifests/deploy/openshift/olm/bundle/4.6/meteringoperator.v4.6.0.clusterserviceversion.yaml
+++ b/manifests/deploy/openshift/olm/bundle/4.6/meteringoperator.v4.6.0.clusterserviceversion.yaml
@@ -253,7 +253,7 @@ spec:
 
   keywords: [metering metrics reporting prometheus chargeback]
   version: "4.6.0"
-  minKubeVersion: "1.18.0"
+  minKubeVersion: "1.17.0"
   maturity: stable
   maintainers:
     - email: sd-operator-metering@redhat.com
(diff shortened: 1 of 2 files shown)


manifests/deploy/openshift/olm/bundle/4.6/meteringoperator.v4.6.0.clusterserviceversion.yaml
```diff
@@ -253,7 +253,7 @@ spec:
 
   keywords: [metering metrics reporting prometheus chargeback]
   version: "4.6.0"
-  minKubeVersion: "1.18.0"
+  minKubeVersion: "1.17.0"
```
bentito, line 256: Not sure if this in unavoidable because of Helm, but it's bad to have to make changes like this in more than one place, it's error-prone and a dev burden.

timflannagan1: This is unavoidable because we're required to have a list of rendered YAML manifests that ART can consume when building our operator in a disconnected environment (OSBS). In general, though, that's what helm is doing, rendering YAML templating and spitting out rendered YAML manifests.

bentito: ✔️

PR #1328, opened 2020-07-23
hack,olm_deploy: Add a script and manifests that enables horizontal machine autoscaling.

timflannagan1:
Pretty straight-forward script:
- Create the ClusterAutoScaler resource that controls and monitors the size of the cluster to help scale up nodes when Pods fail to schedule due to insufficient resources.
- For each `machineset` in the openshift-machine-api namespace, grab the metadata.name of the current `machineset` we're iterating over and create the corresponding MachineAutoScaler resource using `envsubst`.

For more information on these resources, check out the [4.5 autoscaling documentation](https://docs.openshift.com/container-platform/4.5/machine_management/applying-autoscaling.html).


hack/e2e.sh
```diff
@@ -77,6 +78,11 @@ if [ -n "${IMAGE_FORMAT:-}" ]; then
     echo "\$IMAGE_FORMAT=$IMAGE_FORMAT"
 fi
 
+if [[ $TEST_ENABLE_MACHINE_AUTOSCALING == "true" ]]; then
```
timflannagan1, line 81: We would update our release CI configuration to explicitly set this value to true:

```bash
$ make e2e TEST_ENABLE_MACHINE_AUTOSCALE="true" ...
```

timflannagan1:
CC @pruan-rht

hack/enable-machine-autoscaling.sh
```diff
@@ -0,0 +1,17 @@
+#! /bin/bash
+
+set -eou pipefail
```
bentito, line 3:
```suggestion
set -eoux pipefail
```
Will allow for easier debugging if script has problems running in CI

hack/enable-machine-autoscaling.sh
```diff
@@ -0,0 +1,17 @@
...
+
+if ! kubectl get clusterautoscalers default > /dev/null 2>&1; then
+    kubectl create -f ${MANIFESTS_BASE_PATH}/clusterautoscaler.yaml
+fi
+
+MACHINESETS=( $(oc -n openshift-machine-api get machinesets --no-headers | awk '{ print $1 }') )
+for machine in "${MACHINESETS[@]}"; do
+    export MACHINE_NAME=$(oc -n openshift-machine-api get machineset $machine -o jsonpath='{.metadata.name}')
+    envsubst < ${MANIFESTS_BASE_PATH}/machineautoscaler.yaml | kubectl apply -f -
+done
```
bentito: everyone seems to think trailing newlines are a great thing

hack/enable-machine-autoscaling.sh
timflannagan1: Yeah, good catch - added in the last force-push.

PR #1329, opened 2020-07-23
add unit testing for report util methods.

pweil-:
Just throwing in some tests while I poke around the codebase.  

A couple items marked with TODOs that probably need a closer eye from someone more familiar with the usages:

1.  we accept a pointer in SetReportCondition but don't account for nils by either throwing an error or returning.  I added a return but that may be just hiding a programming error - your call.

2.  we accept a pointer in RemoveReportCondition but don't account for nils by either throwing an error or returning.  I added a return but that may be just hiding a programming error - your call.

3.  we don't consider a change in message a change in SetReportCondition.  Not sure if that's on purpose or not.

### DIFF
--- a/pkg/apis/metering/v1/util/report_util.go
+++ b/pkg/apis/metering/v1/util/report_util.go
@@ -71,6 +71,11 @@ func GetReportCondition(status metering.ReportStatus, condType metering.ReportCo
 // SetReportCondition updates the report to include the provided condition. If the condition that
 // we are about to add already exists and has the same status and reason then we are not going to update.
 func SetReportCondition(status *metering.ReportStatus, condition metering.ReportCondition) {
+	// TODO we accept a pointer but do not account for nil.  This seems like a programmatic error, should we
+	// throw an error rather than return since the caller is definitely expecting an action to happen?
+	if status == nil {
+		return
+	}
 	currentCond := GetReportCondition(*status, condition.Type)
 	if currentCond != nil && currentCond.Status == condition.Status && currentCond.Reason == condition.Reason {
 		return
@@ -85,6 +90,11 @@ func SetReportCondition(status *metering.ReportStatus, condition metering.Report
 
 // RemoveReportCondition removes the report condition with the provided type.
 func RemoveReportCondition(status *metering.ReportStatus, condType metering.ReportConditionType) {
+	// TODO we accept a pointer but do not account for nil.  This seems like a programmatic error, should we
+	// throw an error rather than return since the caller is definitely expecting an action to happen?
+	if status == nil {
+		return
+	}
 	status.Conditions = filterOutCondition(status.Conditions, condType)
 }
 
--- a/pkg/apis/metering/v1/util/report_util.go
+++ b/pkg/apis/metering/v1/util/report_util.go
@@ -1,6 +1,7 @@
 package util
 
 import (
+	"errors"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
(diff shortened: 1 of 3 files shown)


pkg/apis/metering/v1/util/report_util.go
```diff
@@ -71,6 +71,11 @@ func GetReportCondition(status metering.ReportStatus, condType metering.ReportCo
 // SetReportCondition updates the report to include the provided condition. If the condition that
 // we are about to add already exists and has the same status and reason then we are not going to update.
 func SetReportCondition(status *metering.ReportStatus, condition metering.ReportCondition) {
+	// TODO we accept a pointer but do not account for nil.  This seems like a programmatic error, should we
+	// throw an error rather than return since the caller is definitely expecting an action to happen?
```
bentito, line 75: Indeed, we should throw and the several call sites should also throw on hearing about the error it looks like. I'll need to look more myself to see if we have existing error handling code that could deal with this at some point up the chain.

It looks like 2 of the 3 current call sites can't send a null pointer, but one is just passing along `report.status` from a `report` it got, so it could, this is in `updateReportStatus()`. It does look like all the callers of `updateReportStatus` are prepared to hear about errors, so I that makes me even more confident about the above error throwing.

pweil-: Thanks @bentito.  If you think we should throw here then I'll take a look at what refactoring that would cause and add it in.

pweil-:
You touched it first so.....

/assign @bentito 

:smile:

pweil-:
/hold

hold for throwing the error instead of check and return

pweil-:
updated to throw errors and pass them along in the calling methods

/hold cancel

pweil-:
infra errors
```
error: failed to load configuration: response from configresolver == 504 (Gateway Timeout)
```
/retest

pweil-:
May not be related to the test timeout failures (?) but there seems to be a lot of failures in the log in the Prometheus importer along the lines of `many-to-many matching not allowed: matching labels must be unique on one side`:

```
time="2020-07-29T02:20:46Z" level=error msg="error collecting metrics" app=metering chunkSize=5m0s component=PrometheusImporter endTime="2020-07-29 02:20:45.998682502 +0000 UTC" error="failed to perform Prometheus query: execution: found duplicate series for the match group {namespace=\"metering-hdfs-olm-upgrade\", persistentvolumeclaim=\"hive-metastore-db-data\"} on the right hand-side of the operation: [{namespace=\"metering-hdfs-olm-upgrade\", persistentvolumeclaim=\"hive-metastore-db-data\", storageclass=\"gp2\", volumename=\"pvc-dd80e4ad-6864-408c-babd-2e82f268f948\"}, {namespace=\"metering-hdfs-olm-upgrade\", persistentvolumeclaim=\"hive-metastore-db-data\", storageclass=\"gp2\"}];many-to-many matching not allowed: matching labels must be unique on one side" logID=oyKHpHbpoN namespace=metering-hdfs-olm-upgrade reportDataSource=persistentvolumeclaim-request-bytes startTime="2020-07-29 02:05:45.998682502 +0000 UTC" stepSize=1m0s tableName=hive.metering.datasource_metering_hdfs_olm_upgrade_persistentvolumeclaim_request_bytes

time="2020-07-29T02:20:46Z" level=debug msg="PrometheusImporter ImportFromLastTimestamp finished" app=metering chunkSize=5m0s component=PrometheusImporter logID=oyKHpHbpoN namespace=metering-hdfs-olm-upgrade reportDataSource=persistentvolumeclaim-request-bytes stepSize=1m0s tableName=hive.metering.datasource_metering_hdfs_olm_upgrade_persistentvolumeclaim_request_bytes

time="2020-07-29T02:20:46Z" level=error msg="error syncing ReportDataSource \"metering-hdfs-olm-upgrade/persistentvolumeclaim-request-bytes\", adding back to queue" ReportDataSource=metering-hdfs-olm-upgrade/persistentvolumeclaim-request-bytes app=metering component=reportDataSourceWorker error="ImportFromLastTimestamp errored: failed to perform Prometheus query: execution: found duplicate series for the match group {namespace=\"metering-hdfs-olm-upgrade\", persistentvolumeclaim=\"hive-metastore-db-data\"} on the right hand-side of the operation: [{namespace=\"metering-hdfs-olm-upgrade\", persistentvolumeclaim=\"hive-metastore-db-data\", storageclass=\"gp2\", volumename=\"pvc-dd80e4ad-6864-408c-babd-2e82f268f948\"}, {namespace=\"metering-hdfs-olm-upgrade\", persistentvolumeclaim=\"hive-metastore-db-data\", storageclass=\"gp2\"}];many-to-many matching not allowed: matching labels must be unique on one side" logID=BlsYESR0bo
```

timflannagan1:
@pweil- yeah that's what is causing the test timeout failures as the test we're running expects to have at least one row of data that gets populated from metrics we pull from Prometheus. The problem is that we're getting many-to-many errors on the PVC-related metrics every time we query Prometheus, which results in no data being populated in those tables. I was looking into ignoring those types of errors, but I can't seem to find a clean way of doing something like this:

```go
t.Run(...) {
  ...
}
if t.Failed() && strings.Contains(t.Stdout, "timed out waiting for the condition") {
  t.Logf("Test failed due to known problem, see BZ: ...")
  t.SkipNow()
}
```

The problem seems that there's no way to grab the stdout or stderr from a failed report (on the testing or testify Golang packages) and `t.SkipNow()` will still result in a failed test despite marking that test as skipped.

pweil-:
Right, that was broached upstream a while ago and rejected as part of the api (skipping an already run and failed test).  We'd need to prevent the error, wait for the right data, or be able to detect it and skip before test failures occur.  

Sounds like you may already have a BZ filed anyway.  Thanks for the retests!

PR #1331, opened 2020-07-23
remove duplicate import and redundant alias for hive

pweil-:
Looks like (probably) the blank import was there to get the driver registered via the init function and someone needed to call `presto.RegisterCustomClient` later so a dupe was added.

Also cleaned up the redundant alias for the `hive` import - the import path is `go-hive-driver` but the actual package is already `hive`.


timflannagan1:
Overriding as it passed everything but the node selector test which has been permanently failing recently

PR #1334, opened 2020-07-24
test/e2e: Disable the node selector e2e test by default.

timflannagan1:
This test is causing frequent e2e flakes as we're adding this custom testing label
to the available sets of nodes in the prior to firing off any metering installations.
The result is that all of the metering installations fight for their pods to be scheduled,
which is fine as the machineautoscaler will provision new machines. With this node selector
test, that's problematic as the new machines that the machineautoscaler provisions will not
have the custom testing node label, so pods will never schedule.

### DIFF
--- a/test/e2e/main_test.go
+++ b/test/e2e/main_test.go
@@ -179,6 +179,43 @@ func TestManualMeteringInstall(t *testing.T) {
 			MeteringConfigManifestFilename: "missing-storage.yaml",
 		},
 		{
+			Name:                      "ValidHDFS-ValidNodeSelector",
+			MeteringOperatorImageRepo: meteringOperatorImageRepo,
+			MeteringOperatorImageTag:  meteringOperatorImageTag,
+			// TODO: transistion this to a periodic test and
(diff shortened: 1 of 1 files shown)


bentito:
Is it actually treated as an ordered list?

timflannagan1:
> Is it actually treated as an ordered list?

Nope - I'm just going to disable this test for the time being once I have time to fix it.

```go
    --- FAIL: TestManualMeteringInstall/ValidHDFS-ValidNodeSelector (1215.74s)
        metering_manual_install_tests.go:89: 
            	Error Trace:	assert.go:112
            	            				metering_manual_install_tests.go:89
            	            				main_test.go:320
            	            	error waiting for metering pods to become ready: the pods failed to report a ready status before the timeout period occurred: timed out waiting for the condition
            	Test:       	TestManualMeteringInstall/ValidHDFS-ValidNodeSelector
            	Messages:   	expected there would be no error
```

timflannagan1:
/retitle test/e2e: Disable the node selector e2e test by default.

PR #1335, opened 2020-07-24
charts,manifests: Add the zap-level argument to the ansible-operator container.

timflannagan1:
This field controls the internal ansible-operator logging level. We can default this logging level to logging both Ansible playbook logs and the internal ansible-operator logs, with the intention of editing this value on the deployment/CSV objects if you want to change the verbosity of the default logging.

Other available values for this argument are 'debug' and 'error' (or providing the verbosity of the logs using an integer):
- https://github.com/operator-framework/operator-sdk/blob/be718958884e04b4c4adfab5e0054524bba5226b/pkg/log/zap/flags.go#L52

In the case where you only want to stream the Ansible playbook logs and not the internal ansible-operator logs, you would edit the value of this argument to '--zap-level=error' in the metering-ansible-operator deployment and that would reduce the general verbosity of the container logs:

```bash
$ oc -n openshift-metering edit deployment metering-operator
...
containers:
- name: operator
  ...
  args:
  - --zap-level=error
  ...
```

### DIFF
--- a/charts/metering-ansible-operator/templates/_helpers.tpl
+++ b/charts/metering-ansible-operator/templates/_helpers.tpl
@@ -27,6 +27,8 @@ template:
     - name: operator
       image: "{{ .Values.operator.image.repository }}:{{ .Values.operator.image.tag }}"
       imagePullPolicy: {{ .Values.operator.image.pullPolicy }}
+      args:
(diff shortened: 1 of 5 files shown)


timflannagan1:
zzz not sure if the prow status was broken there or not

PR #1336, opened 2020-07-24
test/e2e: Add a post-install test that verifies that the metering-ansible-operator is exposing metrics properly.

timflannagan1:
This is a follow-up to the fixes made in https://github.com/kube-reporting/metering-operator/pull/1235 to make sure we don't regress in future releases.

At the moment, all this post-install test does is exec into the `prometheus-k8s` Pod in the openshift-monitoring namespace and ensures that the workqueue_work_duration_seconds_count metric returns a successful status. The ansible-operator, which will expose this metric, creates a ServiceMonitor for the metering-operator-metrics service name, so we should be able to query for the value of this metric in the test namespace that has deployed Metering.

### DIFF
--- a/test/e2e/metering_ansible_operator_metrics_test.go
+++ b/test/e2e/metering_ansible_operator_metrics_test.go
@@ -1,0 +1,38 @@
+package e2e
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/kube-reporting/metering-operator/test/reportingframework"
+	"github.com/kube-reporting/metering-operator/test/testhelpers"
+)
+
+const (
+	prometheusPodName       = "presto-k8s-0"
+	prometheusNamespace     = "openshift-monitoring"
+	prometheusContainerName = "prometheus"
+	prometheusQueryEndpoint = "http://127.0.0.1:9090/api/v1/query"
+)
+
+// testMeteringAnsibleOperatorMetricsWork ensures that the ServiceMonitor object the ansible-operator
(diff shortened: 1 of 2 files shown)


test/e2e/metering_ansible_operator_metrics_test.go
```diff
@@ -0,0 +1,38 @@
...
+	}
+	options := testhelpers.NewExecOptions(prometheusPodName, prometheusNamespace, prometheusContainerName, false, cmd)
+	stdoutBuf, stderrBuf, err := testhelpers.ExecPodCommandWithOptions(rf.KubeConfig, rf.KubeClient, options)
+	t.Logf("Stdout output: %s", stdoutBuf.String())
+	t.Logf("Stderr output: %s", stderrBuf.String())
+	require.Nil(t, err, "expected running the prometheus query would produce no error")
+	require.Containsf(t, stdoutBuf.String(), "workqueue_work_duration_seconds_count", "expected the prometheus query would return the 'workqueue_work_duration_seconds_count' job name")
+	require.Containsf(t, stdoutBuf.String(), "success", "expected the prometheus query would return a successful status")
+	require.Len(t, stderrBuf.String(), 0, "expected that the stderr buffer would return nothing")
+}
```
timflannagan1, line 38: Probably could do something a bit more complex, like using the Prometheus go-client, or grabbing the prometheus-k8s route from the openshift-monitoring namespace, which would be cleaner, but this seemed like the easiest way to fire off a query (and we already had this exec a command in a Pod helper function).

timflannagan1:
Looks like CI is struggling a bit this morning, and this has already passed e2e a couple of times now.

PR #1340, opened 2020-07-28
Documentation: Update the metadata.name used for the aggregation ReportQuery example.

timflannagan1:
This is just to help the process of copying and pasting the examples provided as the metadata.name of the `ReportQuery` example shown conflicts with the default `namespace-cpu-usage` ReportQuery.

### DIFF
--- a/Documentation/rollup-reports.md
+++ b/Documentation/rollup-reports.md
@@ -31,7 +31,7 @@ It contains a few a custom inputs: most importantly, `NamespaceCPUUsageReportNam
 apiVersion: metering.openshift.io/v1
 kind: ReportQuery
 metadata:
-  name: namespace-cpu-usage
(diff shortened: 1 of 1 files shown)


Documentation/rollup-reports.md
```diff
@@ -31,7 +31,7 @@ It contains a few a custom inputs: most importantly, `NamespaceCPUUsageReportNam
 apiVersion: metering.openshift.io/v1
 kind: ReportQuery
 metadata:
-  name: namespace-cpu-usage
+  name: custom-namespace-cpu-usage
```
pruan-rht, line 34: @timflannagan1 if this doesn't exists, I'm afraid it will cause confusion for the end-users.

timflannagan1: You would be creating this query, not referencing an existing one.

PR #1342, opened 2020-07-29
test/e2e: Ignore any test failures that occur during the pre-upgrade installation.

timflannagan1:


timflannagan1:
Hmm, it would appear that I didn't read the documentation for the testing stdlib well enough as `t.Skip()` doesn't skip failed tests, which is what this PR attempts to do. Closing.

PR #1343, opened 2020-07-29
pkg/operator: Update the scheme used for the event record to include the metering v1 scheme.

timflannagan1:
In order to broadcast k8s events to help report errors for the metering-related resources, we need to add the metering scheme that gets generated from the `pkg/apis/metering/v1/register.go` implementation on top of the default scheme the controller was using the report the leader election event. Changing the event recorder to use the metering scheme, in addition to the default scheme allows us to broadcast events to different object's lifecycles.

### DIFF
--- a/pkg/operator/operator.go
+++ b/pkg/operator/operator.go
@@ -25,6 +25,9 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/clock"
 	"k8s.io/apimachinery/pkg/util/wait"
+
+	meteringv1scheme "github.com/kube-reporting/metering-operator/pkg/generated/clientset/versioned/scheme"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/client-go/kubernetes/scheme"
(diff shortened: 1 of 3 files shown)


pkg/operator/operator.go
```diff
@@ -305,13 +308,21 @@ func newReportingOperator(
 		reporting.NewReportListerGetter(reportInformer.Lister()),
 	)
 
+	logger.Infof("setting up event broadcasters")
+	utilruntime.Must(meteringv1scheme.AddToScheme(scheme.Scheme))
```
bentito, line 312: a)`Must` is a candidate for worst function name ever; b) why are we wrapping with it? Like why do we want to cause a panic there, how will it help matters?

bentito: Could rename `Must` to `PanicOnError` and delete its comment.

timflannagan1: Yeah agreed on the poor function name selection. But to answer your question, `AddToScheme(...)` returns an error, but the Reporting constructor function does not, so I thought a panic would be applicable here instead of ignoring that error.

bentito: Not our func, nvm

PR #1344, opened 2020-07-29
pkg/operator: Improve the debugging experience for invalid Prometheus configurations.

timflannagan1:
Our current health checks are contingent on queries from Presto being successful, but debugging installations where no Prometheus metrics are being imported to the data source tables can be difficult at times. In order to help alleviate this, I added an initial poll that queries the Prometheus `up` metric for data and checks whether the response of that query has a length greater than zero characters. This kind of check provides an avenue toward weeding out the invalid reporting-operator + Prometheus configurations that are passed to the MeteringConfig custom resource pretty quickly in the reporting-operator container logs.

Here is an example of the container logs when an invalid Prometheus host is passed:
```
time="2020-07-29T17:50:49Z" level=info msg="writes to Presto are succeeding" app=metering
time="2020-07-29T17:50:49Z" level=info msg="testing the ability to query from Prometheus" app=metering
time="2020-07-29T17:50:55Z" level=error msg="failed to successfully query Prometheus: Get https://thanos-querier.openshift-monitoring.svc:8080/api/v1/query?query=up&time=2020-07-29T17%3A50%3A52.641742567Z: dial tcp 172.30.48.65:8080: connect: no route to host" app=metering
time="2020-07-29T17:50:55Z" level=info msg="basic initialization completed" app=metering
time="2020-07-29T17:50:55Z" level=info msg="starting the informers" app=metering
time="2020-07-29T17:50:55Z" level=info msg="waiting for caches to sync" app=metering
time="2020-07-29T17:50:55Z" level=info msg="starting leader election" app=metering
```

On top of the initial probe, I updated the way we handle failed Prometheus queries when importing metrics for the ReportDataSource custom resources, firing off a warning event indicating something problematic may have occurred.

Using the sample situation as the above example where the Prometheus host is invalid, here are the events that are created as a result of a failed Prometheus query when importing metrics:
```bash
$ oc get events --field-selector reason="FailedPrometheusQuery"
LAST SEEN   TYPE      REASON                  OBJECT                                                    MESSAGE
6m59s       Warning   FailedPrometheusQuery   reportdatasource/node-allocatable-cpu-cores               Unable to import metrics after Prometheus query failure. Check the reporting-operator container logs for more information.
37s         Warning   FailedPrometheusQuery   reportdatasource/node-allocatable-cpu-cores               Unable to import metrics after Prometheus query failure. Check the reporting-operator container logs for more information.
6m56s       Warning   FailedPrometheusQuery   reportdatasource/node-allocatable-memory-bytes            Unable to import metrics after Prometheus query failure. Check the reporting-operator container logs for more information.
37s         Warning   FailedPrometheusQuery   reportdatasource/node-allocatable-memory-bytes            Unable to import metrics after Prometheus query failure. Check the reporting-operator container logs for more information.
6m53s       Warning   FailedPrometheusQuery   reportdatasource/node-capacity-cpu-cores                  Unable to import metrics after Prometheus query failure. Check the reporting-operator container logs for more information.
37s         Warning   FailedPrometheusQuery   reportdatasource/node-capacity-cpu-cores                  Unable to import metrics after Prometheus query failure. Check the reporting-operator container logs for more information.
6m56s       Warning   FailedPrometheusQuery   reportdatasource/node-capacity-memory-bytes               Unable to import metrics after Prometheus query failure. Check the reporting-operator container logs for more information.
31s         Warning   FailedPrometheusQuery   reportdatasource/node-capacity-memory-bytes               Unable to import metrics after Prometheus query failure. Check the reporting-operator container logs for more information.
7m3s        Warning   FailedPrometheusQuery   reportdatasource/persistentvolumeclaim-capacity-bytes     Unable to import metrics after Prometheus query failure. Check the reporting-operator container logs for more information.
22s         Warning   FailedPrometheusQuery   reportdatasource/persistentvolumeclaim-capacity-bytes     Unable to import metrics after Prometheus query failure. Check the reporting-operator container logs for more information.
6m56s       Warning   FailedPrometheusQuery   reportdatasource/persistentvolumeclaim-phase              Unable to import metrics after Prometheus query failure. Check the reporting-operator container logs for more information.
40s         Warning   FailedPrometheusQuery   reportdatasource/persistentvolumeclaim-phase              Unable to import metrics after Prometheus query failure. Check the reporting-operator container logs for more information.
6m59s       Warning   FailedPrometheusQuery   reportdatasource/persistentvolumeclaim-request-bytes      Unable to import metrics after Prometheus query failure. Check the reporting-operator container logs for more information.
25s         Warning   FailedPrometheusQuery   reportdatasource/persistentvolumeclaim-request-bytes      Unable to import metrics after Prometheus query failure. Check the reporting-operator container logs for more information.
6m53s       Warning   FailedPrometheusQuery   reportdatasource/persistentvolumeclaim-usage-bytes        Unable to import metrics after Prometheus query failure. Check the reporting-operator container logs for more information.
40s         Warning   FailedPrometheusQuery   reportdatasource/persistentvolumeclaim-usage-bytes        Unable to import metrics after Prometheus query failure. Check the reporting-operator container logs for more information.
6m56s       Warning   FailedPrometheusQuery   reportdatasource/pod-limit-cpu-cores                      Unable to import metrics after Prometheus query failure. Check the reporting-operator container logs for more information.
25s         Warning   FailedPrometheusQuery   reportdatasource/pod-limit-cpu-cores                      Unable to import metrics after Prometheus query failure. Check the reporting-operator container logs for more information.
6m56s       Warning   FailedPrometheusQuery   reportdatasource/pod-limit-memory-bytes                   Unable to import metrics after Prometheus query failure. Check the reporting-operator container logs for more information.
37s         Warning   FailedPrometheusQuery   reportdatasource/pod-limit-memory-bytes                   Unable to import metrics after Prometheus query failure. Check the reporting-operator container logs for more information.
7m6s        Warning   FailedPrometheusQuery   reportdatasource/pod-persistentvolumeclaim-request-info   Unable to import metrics after Prometheus query failure. Check the reporting-operator container logs for more information.
40s         Warning   FailedPrometheusQuery   reportdatasource/pod-persistentvolumeclaim-request-info   Unable to import metrics after Prometheus query failure. Check the reporting-operator container logs for more information.
6m56s       Warning   FailedPrometheusQuery   reportdatasource/pod-request-cpu-cores                    Unable to import metrics after Prometheus query failure. Check the reporting-operator container logs for more information.
40s         Warning   FailedPrometheusQuery   reportdatasource/pod-request-cpu-cores                    Unable to import metrics after Prometheus query failure. Check the reporting-operator container logs for more information.
6m56s       Warning   FailedPrometheusQuery   reportdatasource/pod-request-memory-bytes                 Unable to import metrics after Prometheus query failure. Check the reporting-operator container logs for more information.
40s         Warning   FailedPrometheusQuery   reportdatasource/pod-request-memory-bytes                 Unable to import metrics after Prometheus query failure. Check the reporting-operator container logs for more information.
7m6s        Warning   FailedPrometheusQuery   reportdatasource/pod-usage-cpu-cores                      Unable to import metrics after Prometheus query failure. Check the reporting-operator container logs for more information.
37s         Warning   FailedPrometheusQuery   reportdatasource/pod-usage-cpu-cores                      Unable to import metrics after Prometheus query failure. Check the reporting-operator container logs for more information.
6m56s       Warning   FailedPrometheusQuery   reportdatasource/pod-usage-memory-bytes                   Unable to import metrics after Prometheus query failure. Check the reporting-operator container logs for more information.
40s         Warning   FailedPrometheusQuery   reportdatasource/pod-usage-memory-bytes                   Unable to import metrics after Prometheus query failure. Check the reporting-operator container logs for more information.
```


pkg/operator/datasources.go
```diff
@@ -230,6 +230,7 @@ func (op *Reporting) handlePrometheusMetricsDataSource(logger log.FieldLogger, d
 	// run the import
 	results, err := importer.ImportFromLastTimestamp(context.Background())
 	if err != nil {
+		op.eventRecorder.Event(dataSource, v1.EventTypeWarning, "FailedPrometheusQuery", "Unable to import metrics after Prometheus query failure. Check the reporting-operator container logs for more information.")
```
timflannagan1, line 233: We may want to expose a configuration for disabling this kind of check as this could be potentially expensive if every ReportDataSource promQL query is constantly failing.

timflannagan1, line 233: Also, we could include the actual error in the event's message, but it can pretty long and difficult to parse, especially when we get an error that's in this kind of form:
```
Get https://thanos-querier.openshift-monitoring.svc:8080/api/v1/query?query=up&time=2020-07-29T17%3A50%3A52.641742567Z: dial tcp 172.30.48.65:8080: connect: no route to host
```

timflannagan1: Well maybe expensive at first, but that should normalize over time as we're not going to be updating these events as there's going to be no ReportDataSource events to process in the cache as we're constantly failing to import metrics, at least in this scenario.

pkg/operator/operator.go
```diff
@@ -598,10 +599,33 @@ func (op *Reporting) Run(ctx context.Context) error {
...
+	// indicating there was a problem and ignore the error.
+	if !op.cfg.DisablePrometheusMetricsImporter {
+		op.logger.Info("testing the ability to query from Prometheus")
+		err = wait.Poll(3*time.Second, 15*time.Second, func() (bool, error) {
+			res, err := op.promConn.Query(context.TODO(), "up", time.Now())
+			if err != nil {
+				return false, err
+			}
+
+			return len(res.String()) != 0, nil
```
bentito, line 618: I don't think the cast to String is reliable on this complex type.

pkg/operator/operator.go
```diff
@@ -598,10 +599,33 @@ func (op *Reporting) Run(ctx context.Context) error {
...
+	// been disabled, try to fire off a test Prometheus query against the `up`
+	// metric. This should weed out any invalid Prometheus configurations, e.g.
+	// the service account token or CA passed to the reporting-operator is invalid,
+	// and we should complain about that, but not require this as part of the
+	// reporting-operator health checks. Note: in the case where we encounter
+	// something like a timeout error while polling, just log the message
+	// indicating there was a problem and ignore the error.
+	if !op.cfg.DisablePrometheusMetricsImporter {
+		op.logger.Info("testing the ability to query from Prometheus")
+		err = wait.Poll(3*time.Second, 15*time.Second, func() (bool, error) {
```
pweil-, line 612: A couple questions with this implementation.  I am understanding that the objective here is to test that the configuration we have for prometheus is correct.  With that assumption the questions are:

1.  Should we attempt to query prom multiple times - ie. will an incorrect configuration, network issue, etc resolve itself in the 15 seconds we're polling?
2.  Do we care about the actual state of prometheus at this point (the result of the query)?
3.  Do we care about the length of time the query to prometheus takes?

On the assumption that the answer to 2 is no (we're just logging the results of errors and prom going down at runtime needs to be accounted for anyway) and the answer to 3 is yes there are two suggestions.

First, no matter the answer to 1 - we cap the prom query by using a real context via a `context.TODO`.

If the answer to 1 is yes - bad config could resolve itself in the 15 second polling period we should adjust to not return an error (which stops polling right away) and log in the poll method, letting it time out if we don't resolve by the time `time.Duration` hits (15 seconds):

```go
err = wait.Poll(3*time.Second, 15*time.Second, func() (bool, error) {
   _, err := op.promConn.Query(context.TODO(), "up", time.Now())
   if err != nil {
      // give it another try, don't return the error
      op.logger.Errorf("failed to successfully query Prometheus: %v", err)
      return false, nil
   }

   return true, nil
})
```

If the answer to 1 is no, config will not change in the 15 seconds so we have no need to keep retrying it and we're just going to log unavailable prometheus anyway this boils down to just:

```go
_, err := op.promConn.Query(context.TODO(), "up", time.Now())
if err != nil {
   op.logger.Errorf("failed to successfully query Prometheus: %v", err)
}
```

test/e2e/ensure_no_failed_prometheus_events_test.go
```diff
@@ -0,0 +1,27 @@
...
+const failedPrometheusQueryEventReason = "FailedPrometheusQuery"
+
+func testFailedPrometheusQueryEvents(t *testing.T, rf *reportingframework.ReportingFramework) {
+	events, err := rf.KubeClient.CoreV1().Events(rf.Namespace).List(context.Background(), metav1.ListOptions{})
+	require.NoError(t, err, "failed to list the events in the test namespace")
+
+	var found bool
+	for _, event := range events.Items {
+		if event.Reason == failedPrometheusQueryEventReason {
+			found = true
```
pweil-, line 22: break here, no reason to keep looping through n events

timflannagan1: Whoops, not sure how I missed that one. Good catch.

pkg/operator/operator.go
timflannagan1: The main objective I had in mind for this code block is to provide some sort of logging or event that can provide an avenue for quickly debugging potential problems with our, or the user-provided, Prometheus configuration. With that in mind, firing off a query for the `up` metric in Prometheus at the beginning of the operator's workflow seemed like a decent health check. Note: we could uncover a subset of these errors while processing the ReportDataSource promQL queries too, so this kind of check could be redundant, but those failures could be due to something that's not configuration-based (e.g. a many-to-many error).

I wanted to account for a potential bad query to Prometheus (for whatever reason) so that's why I used `poll.Wait` to fire off subsequent ones, but the 15-second timeout was arbitrarily chosen with the goal that our operator would still pass the reporting-operator's health checks as we're not polling for a substantial amount of time. But to answer your questions, the configuration passed to the reporting-operator most likely isn't going to change within a 15-second window, so a single Prometheus query could suffice, assuming we update the context to have a deadline.

timflannagan1: And we could also fire off an event indicating a potential problem, on top of complaining about this failed query in the reporting-operator container logs.

pweil-: cool, thanks.  Ok, so if you want to try multiple times for whatever reason (maybe there is a network error, whatever) then not returning the error (like in the first example) will keep you polling like you wanted :+1: 

Other items of note.  You may consider changing `op.logger.Errorf` to warning level and also logging a success before `return true, nil` so when reading the log you can see if it ever resolved successfully (maybe even with the `result.String()` if that's helpful info).  

Thanks for the clarity.

timflannagan1: Made a couple of changes based on your feedback and tried to improve the logging. In the case where the host of the Prometheus instance is invalid, the following logs would be produced:
```
time="2020-07-30T15:39:09Z" level=info msg="testing the ability to query from Prometheus" app=metering
time="2020-07-30T15:39:16Z" level=warning msg="failed to succesfully query Prometheus: Get https://thanos-querier.openshift-monitoring.svc:8080/api/v1/query?query=up&time=2020-07-30T15%3A39%3A12.983041081Z: dial tcp 172.30.245.171:8080: connect: no route to host" app=metering
time="2020-07-30T15:39:19Z" level=warning msg="failed to succesfully query Prometheus: Get https://thanos-querier.openshift-monitoring.svc:8080/api/v1/query?query=up&time=2020-07-30T15%3A39%3A18.983069575Z: dial tcp 172.30.245.171:8080: connect: no route to host" app=metering
time="2020-07-30T15:39:25Z" level=warning msg="failed to succesfully query Prometheus: Get https://thanos-querier.openshift-monitoring.svc:8080/api/v1/query?query=up&time=2020-07-30T15%3A39%3A21.983024157Z: dial tcp 172.30.245.171:8080: connect: no route to host" app=metering
time="2020-07-30T15:39:28Z" level=warning msg="failed to succesfully query Prometheus: Get https://thanos-querier.openshift-monitoring.svc:8080/api/v1/query?query=up&time=2020-07-30T15%3A39%3A25.064029555Z: dial tcp 172.30.245.171:8080: connect: no route to host" app=metering
time="2020-07-30T15:39:28Z" level=warning msg="queries from Prometheus are failing: timed out waiting for the condition" app=metering
```
I'm not sure if every failure should be a warning level, but that's probably better than making them debug level, seeing that `queries from Prometheus are failing: timed out waiting for the condition` isn't super helpful and would require redeploying the reporting-operator with a higher log verbosity.

timflannagan1: Alternatively, could just return the error we grab from the poll.Wait condition function instead of nil, so the outer scope can just report on that error instead of the `timed out waiting for the condition`.

pweil-: > Alternatively, could just return the error we grab from the poll.Wait condition function instead of nil, so the outer scope can just report on that error instead of the timed out waiting for the condition.

If you return the error you stop polling so lose the "try multiple times" component depending on when the error happens.  You could get fancier and aggregate the errors (but not return them) and report it after poll finishes if it doesn't finish with success (where you nil out the aggregated error on success or something)

pweil-: but fwiw, it seems ok to be simple, to me.

PR #1352, opened 2020-08-02
pkg/operator: Add more comments and perform some general maintenance for the expired reports methods/functions.

timflannagan:
This just aims to perform some light maintenance on the recently merged for supporting automatic GCing of Reports based on a user-provided retention period configured for a particular Report custom resource.

### DIFF
--- a/pkg/operator/reports.go
+++ b/pkg/operator/reports.go
@@ -354,38 +354,68 @@ func getReportPeriod(now time.Time, logger log.FieldLogger, report *metering.Rep
 	return reportPeriod, nil
 }
 
+// handleExpiredReport is responsible for determining if the @report object we're
+// currently processing has reached it's configured retention period. In the case
+// where the Report object hasn't been configured with an expiration, exit early,
+// else attempt to determine if we can safely delete this object without breaking
+// other Report custom resources that rely on this particular Report.
 func (op *Reporting) handleExpiredReport(logger log.FieldLogger, report *metering.Report, now time.Time) error {
-	// check if the Report is being deleted already
+	// check if the Report deletion timestamp has already been set, exit early.
 	if report.DeletionTimestamp != nil {
-		logger.Warnf("report was already marked for deletion")
+		// TODO: should we be requeuing here?
+		logger.Warnf("the %s Report in the %s namespace was already marked for deletion", report.Name, report.Namespace)
 		return nil
 	}
-	// check if the Report is past its retention time
-	if reportExpired := isReportExpired(op.logger, report, now); !reportExpired {
+
+	// check if the report we're currently processing has reached the
+	// user-provided retention period. If it hasn't, exit early and
+	// let the call site determine how to handle this Report object.
+	if reportExpired := isReportExpired(logger, report, now); !reportExpired {
 		logger.Debugf("the %s Report in the %s namespace has not yet reached the expiration date", report.Name, report.Namespace)
 		return nil
 	}
-	// check if the Report is used by any other Report or ReportQuery, if not delete it
-	if reportIsNotInput := isReportNotUsedAsInput(logger, report, op); reportIsNotInput {
+	// Determine whether this Report object can be safely deleted.
+	// In the case of something like a roll-up Report, where an
+	// aggregate Report uses another Report object as an input,
+	// we need to ensure that we're not deleting any Report that's
+	// listed as a dependency, even if the user-provided retention
+	// period has passed.
+	reportIsNotInput, err := op.isReportNotUsedAsInput(logger, report)
+	if err != nil {
+		logger.Warnf("failed to determine if the %s Report is listed as a dependency: %v", report.Name, err)
+		return nil
+	}
+	// If the Report is not listed as a dependency for another
+	// custom resource, then attempt to make the client-go API
+	// call to delete this resource. Deleting this resource will
+	// result in the informer picking up this deletion event and will run
+	// op.deleteReport once that event has been picked up in the cache.
+	if reportIsNotInput {
 		newDeletionTimestamp := metav1.Now()
 		report.SetDeletionTimestamp(&newDeletionTimestamp)
-		err := op.meteringClient.MeteringV1().Reports(report.Namespace).
-			Delete(context.TODO(), report.Name, metav1.DeleteOptions{})
+
+		err := op.meteringClient.MeteringV1().Reports(report.Namespace).Delete(context.TODO(), report.Name, metav1.DeleteOptions{})
 		if apierrors.IsNotFound(err) {
-			logger.Infof("report: %s, not deleted because it was not found at time delete attempted", report.Name)
+			logger.Infof("failed to delete the %s Report in the %s namespace but it does not exist anymore", report.Name, report.Namespace)
 			return nil
 		}
 		if err != nil {
-			return fmt.Errorf("failed to delete the expired report: %s. err: %s", report.Name, err)
+			return fmt.Errorf("failed to delete the %s expired report: %v", report.Name, err)
 		}
-		logger.Infof("deleted Report: %s in the Namespace: %s because it reached the expiration time",
+
+		logger.Infof("deleted the Report: %s in the Namespace: %s because it reached the expiration time",
 			report.Name, report.Namespace)
 		op.eventRecorder.Event(report, v1.EventTypeNormal, "ExpiredReportHasBeenDeleted",
 			fmt.Sprintf("Deleted the %s Report as the configured expiration date has passed", report.Name))
 
 		return nil
 	}
-	// if here, warn about dependency before return
+
+	// If we reach this part of the function, we have a good idea that
+	// we're processing an "expired" Report, but we cannot safely delete
+	// this object as it's listed as a dependent for another custom resource,
+	// so we log and fire an event describing this scenario but don't take any
+	// further action.
 	logger.Warnf("report: %s, would be deleted because expired, but is depended on", report.Name)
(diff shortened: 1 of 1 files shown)


timflannagan1:
The e2e test failed due to a known issue and passed everything else

pkg/operator/reports.go
```diff
@@ -844,21 +874,31 @@ func isReportNotUsedAsInput(logger log.FieldLogger, report *metering.Report, op
...
 		}
 	}
-	return true
+	return true, nil
 }
 
+// isReportExpired is a helper function that is responsible for inspecting
+// the @report custom resource and checking whether the user has provided
+// a retention period. In the case that field has been populated, check
+// if the we're currently processing an "expired" report and return true,
```
bentito, line 887:
```suggestion
// if we're currently processing an "expired" report and return true,
```

pkg/operator/reports.go
```diff
@@ -844,21 +874,31 @@ func isReportNotUsedAsInput(logger log.FieldLogger, report *metering.Report, op
...
 	if report.Spec.Expiration != nil {
 		currentTimeNano := currentTime.UTC().Truncate(time.Millisecond).UnixNano()
 		reportCreationNano := report.ObjectMeta.CreationTimestamp.UTC().Truncate(time.Millisecond).UnixNano()
 		retentionAmountNano := report.Spec.Expiration.Nanoseconds()
+
 		if currentTimeNano > (reportCreationNano + retentionAmountNano) {
 			return true
 		}
-		logger.Infof("curr: %d, creat: %d, retain: %d", currentTimeNano, reportCreationNano, retentionAmountNano)
+		logger.Infof("current timestamp: %d, creation timestamp: %d, retention timestamp: %d", currentTimeNano, reportCreationNano, retentionAmountNano)
```
bentito, line 901:
```suggestion
		logger.Debugf("current timestamp: %d, creation timestamp: %d, retention timestamp: %d", currentTimeNano, reportCreationNano, retentionAmountNano)
```
Meant to delete this or we can make it debug...

pkg/operator/reports.go
```diff
@@ -354,38 +354,68 @@ func getReportPeriod(now time.Time, logger log.FieldLogger, report *metering.Rep
...
+// where the Report object hasn't been configured with an expiration, exit early,
+// else attempt to determine if we can safely delete this object without breaking
+// other Report custom resources that rely on this particular Report.
 func (op *Reporting) handleExpiredReport(logger log.FieldLogger, report *metering.Report, now time.Time) error {
-	// check if the Report is being deleted already
+	// check if the Report deletion timestamp has already been set, exit early.
 	if report.DeletionTimestamp != nil {
-		logger.Warnf("report was already marked for deletion")
+		// TODO: should we be requeuing here?
+		logger.Warnf("the %s Report in the %s namespace was already marked for deletion", report.Name, report.Namespace)
```
timflannagan1, line 366: I realize this kind of logging may be a bit too verbose as the logger we're passing to this method already contains fields about the report name and namespace.

pkg/operator/reports.go
```diff
@@ -354,38 +354,68 @@ func getReportPeriod(now time.Time, logger log.FieldLogger, report *metering.Rep
...
+// currently processing has reached it's configured retention period. In the case
+// where the Report object hasn't been configured with an expiration, exit early,
+// else attempt to determine if we can safely delete this object without breaking
+// other Report custom resources that rely on this particular Report.
 func (op *Reporting) handleExpiredReport(logger log.FieldLogger, report *metering.Report, now time.Time) error {
-	// check if the Report is being deleted already
+	// check if the Report deletion timestamp has already been set, exit early.
 	if report.DeletionTimestamp != nil {
-		logger.Warnf("report was already marked for deletion")
+		// TODO: should we be requeuing here?
```
timflannagan1, line 365: We can leave this here for now, but I assume we should be re-queueing the resource here instead of logging and exiting early.

pkg/operator/reports.go
```diff
@@ -354,38 +354,68 @@ func getReportPeriod(now time.Time, logger log.FieldLogger, report *metering.Rep
 	return reportPeriod, nil
 }
 
+// handleExpiredReport is responsible for determining if the @report object we're
+// currently processing has reached it's configured retention period. In the case
```
bentito, line 358:
```suggestion
// currently processing has reached its configured retention period. In the case
```

pkg/operator/reports.go
bentito, line 365: returning `nil` from this func allows `syncReport` to keep going on the Report object -- has no effect on the handling of the Report, right?

pkg/operator/reports.go
```diff
@@ -354,38 +354,68 @@ func getReportPeriod(now time.Time, logger log.FieldLogger, report *metering.Rep
...
+	if reportIsNotInput {
 		newDeletionTimestamp := metav1.Now()
 		report.SetDeletionTimestamp(&newDeletionTimestamp)
-		err := op.meteringClient.MeteringV1().Reports(report.Namespace).
-			Delete(context.TODO(), report.Name, metav1.DeleteOptions{})
+
+		err := op.meteringClient.MeteringV1().Reports(report.Namespace).Delete(context.TODO(), report.Name, metav1.DeleteOptions{})
 		if apierrors.IsNotFound(err) {
-			logger.Infof("report: %s, not deleted because it was not found at time delete attempted", report.Name)
+			logger.Infof("failed to delete the %s Report in the %s namespace but it does not exist anymore", report.Name, report.Namespace)
```
bentito, line 399: The rewrite of the log line reads more like normal English, the problem is when there are null values.
Then the old log line reads like this:
`report: , not deleted because it was not found at time delete attempted`
whereas the rewrite reads like this:
`failed to delete the Report in the namespace but it does not exist anymore`
Unless you're a log reader who has made themselves very familiar with the code base, that latter log line makes you question the programmer's sanity for throwing this log line in the first place. The colon and comma are used to give a hint that something is missing.

pkg/operator/reports.go
```diff
@@ -844,21 +874,31 @@ func isReportNotUsedAsInput(logger log.FieldLogger, report *metering.Report, op
...
+
+			return false, nil
 		}
 	}
-	return true
+	return true, nil
 }
 
+// isReportExpired is a helper function that is responsible for inspecting
+// the @report custom resource and checking whether the user has provided
```
bentito, line 885: @Report? (capital R)

pkg/operator/reports.go
```diff
@@ -844,21 +874,31 @@ func isReportNotUsedAsInput(logger log.FieldLogger, report *metering.Report, op
...
+// a retention period. In the case that field has been populated, check
+// if we're currently processing an "expired" report and return true,
+// else return false. We define "expired" here using the following calculation:
+// time.Now() > report.Metadata.CreationTimestamp + report.Spec.Expiration.
 func isReportExpired(logger log.FieldLogger, report *metering.Report, currentTime time.Time) bool {
+	// TODO: should this be contingent on the Report object reporting
```
bentito, line 891: Good point. I guess this implementation is probably too literal, the user set an expiration, but it's going to happen before the Report runs, probably not the user's intention.

pkg/operator/reports.go
timflannagan1: Yeah returning nil here would allow `op.syncReport` to continue processing the resource. I think we may want to just re-queue here with the assumption that if the deletion timestamp is not nil, we should assume it's been deleted either by us or the user, but let the informer determine what's the best course of action to take.

timflannagan1: Actually @pweil- are you aware of any best practices as to when we should be re-queuing resources we're currently processing vs. exiting early and letting the call site determine how to handle the resource from there (assuming there's sufficient logic for the latter case.)

pkg/operator/reports.go
timflannagan1: I think that's a decent situation to consider, but we've essentially performed these validation checks at the call site for this method when we attempt to use the client the object from the cache's metadata key. That said, we already configured the logger for this resource to have the report name and namespace as fields, so this kind of logging may be redundant anyways.

pkg/operator/reports.go
pweil-: Is the more appropriate question to ask "why are we even in this method for a deleted or unexpired report"?  The orchestrator (`syncReport` or `handleReport`) can take care of this for us by calling `isReportExpired` and only calling `handleExpiredReport` when necessary (ie. it's expired and not deleted) or requeing the object, no?

Then you aren't tangling up the orchestration logic with the handler logic as much.  Which incidentally is my answer to your original question: avoid coupling the orchestrator that determines state and next action with the handler that performs a discreet piece of work.

Handlers that defend against bad input are good though but no work to do is success in most cases.

timflannagan1: Yeah agreed, that sounds more reasonable than how we're currently handing processing resources during upgrades. @bentito I think if we do some light refactoring of these methods based on what Paul's comment, we can also further weed out situations where we don't need to create a deep copy (e.g. are we processing an expired or finished report?) and save the overhead of making that kind of expensive call.

bentito: I agree, the changes to those methods should save unneeded work.

pkg/operator/reports.go
bentito: I think @pweil- mentioned on his overall recent refactor PR, we could use some dev standards documented in a file in the code base. Things like, new functions need a comment;  always try to format log lines like XXX; we use Go style YYY.  We'll get code and log lines we're all more familiar with that way after awhile.

PR #1353, opened 2020-08-03
test/e2e: Migrate to using the 10.2 MariaDB imagestream in tests.

timflannagan1:
It looks like the MySQL 5.7 imagestream was recently removed from 4.6 clusters, and we still have the open BZ for Metering cannot work with the most recently GA'd MySQL release, which is causing this e2e test to constantly fail.

Looking at the MariaDB vs. MySQL compatibility chart, it looks like 10.2/10.3/10.4 are all compatible with MySQL 5.7:
- https://mariadb.com/kb/en/mariadb-vs-mysql-compatibility/

These changes just update the imagestream we create in the e2e test to use the mariadb:10.2 version.

### DIFF
--- a/test/e2e/metering_manual_install_test.go
+++ b/test/e2e/metering_manual_install_test.go
@@ -357,7 +357,7 @@ func createMySQLDatabase(ctx *deployframework.DeployerCtx) error {
 		"oc",
 		"-n", mysqlNamespace,
 		"new-app",
-		"--image-stream", "mysql:5.7",
+		"--image-stream", "mariadb:10.2",
 		"MYSQL_USER=testuser",
 		"MYSQL_PASSWORD=testpass",
 		"MYSQL_DATABASE=metastore",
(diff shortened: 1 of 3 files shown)


test/e2e/metering_manual_install_test.go
```diff
@@ -249,7 +249,7 @@ func createMySQLDatabase(ctx *deployframework.DeployerCtx) error {
 		"oc",
 		"-n", mysqlNamespace,
 		"new-app",
-		"--image-stream", "mysql:5.7",
+		"--image-stream", "mariadb:10.2",
```
timflannagan1, line 252: Just a side note too, I would prefer to not outright disable this test as it's nice to have and could provide a way to quickly test whether or not the Mysql8 + metering BZ works or not, depending on how we prioritize that work.

timflannagan1:
This is blocking all e2e runs at the moment, so let's just hard disable in #1358 and circle back to this later.

test/e2e/metering_manual_install_test.go
pruan-rht: @timflannagan1 perhaps we should put a more about hack to delete the pod to get the hive-metastore to start creation `delete pod -l app=hive`

PR #1354, opened 2020-08-03
Dockerfile: Get the faq dependency from a released linux binary.

timflannagan1:


timflannagan1:
This affects only our CI configuration and we need to force-merge this kind of change as this is the CI src Dockerfile.

PR #1356, opened 2020-08-03
Operator refactor - validation and import cleanup

pweil-:
Round 1 - go!

Let's see what clean up we can do here @timflannagan1.

This has the following commits that can be reviewed independently

1.  move the type declarations to their own file and try to comment them up for future readers
2.  since the reporting operator was not exporting anything, introduce an interface and unexport the type.  Consumers deal with the interface now
3.  move all the smaller validation blurbs from the various sections that are called during `New`, `newOperator`, and `Run` methods and consolidate on a single validation call to ensure the config is good during `New`.  No good config == no operator to run.  Bonus: Add tests for all validations
4.  rename cb* imports to remove references to chargeback

There are a couple questions sprinkled in where things weren't obvious.

### DIFF
--- a/pkg/operator/operator.go
+++ b/pkg/operator/operator.go
@@ -14,7 +14,7 @@ import (
 	"time"
 
 	"github.com/kube-reporting/metering-operator/pkg/db"
-	cbClientset "github.com/kube-reporting/metering-operator/pkg/generated/clientset/versioned"
+	clientset "github.com/kube-reporting/metering-operator/pkg/generated/clientset/versioned"
 	meteringv1scheme "github.com/kube-reporting/metering-operator/pkg/generated/clientset/versioned/scheme"
 	factory "github.com/kube-reporting/metering-operator/pkg/generated/informers/externalversions"
 	listers "github.com/kube-reporting/metering-operator/pkg/generated/listers/metering/v1"
@@ -57,7 +57,7 @@ type defaultReportingOperator struct {
 
 	kubeClient        corev1.CoreV1Interface
 	coordinatorClient coordinatorv1.CoordinationV1Interface
-	meteringClient    cbClientset.Interface
+	meteringClient    clientset.Interface
 	eventRecorder     record.EventRecorder
 
 	informerFactory factory.SharedInformerFactory
@@ -149,7 +149,7 @@ func New(logger log.FieldLogger, cfg Config) (ReportingOperator, error) {
 	}
 
 	logger.Debugf("setting up Metering client...")
-	meteringClient, err := cbClientset.NewForConfig(kubeConfig)
+	meteringClient, err := clientset.NewForConfig(kubeConfig)
 	if err != nil {
 		return nil, fmt.Errorf("Unable to create Metering client: %v", err)
 	}
@@ -177,7 +177,7 @@ func newReportingOperator(
 	kubeConfig *rest.Config,
 	kubeClient corev1.CoreV1Interface,
 	coordinatorClient coordinatorv1.CoordinationV1Interface,
-	meteringClient cbClientset.Interface,
+	meteringClient clientset.Interface,
 	informerNamespace string,
 ) ReportingOperator {
 	informerFactory := factory.NewSharedInformerFactoryWithOptions(meteringClient, defaultResyncPeriod, factory.WithNamespace(informerNamespace), factory.WithTweakListOptions(nil))
--- a/pkg/operator/operator.go
+++ b/pkg/operator/operator.go
@@ -52,7 +52,7 @@ import (
 )
 
 // defaultReportingOperator is the concrete implementation of a ReportingOperator.
-type Reporting struct {
+type defaultReportingOperator struct {
 	cfg        Config
 	kubeConfig *rest.Config
 
@@ -116,7 +116,7 @@ func (cfg *TLSConfig) Valid() error {
 	return nil
 }
 
-func New(logger log.FieldLogger, cfg Config) (*Reporting, error) {
+func New(logger log.FieldLogger, cfg Config) (ReportingOperator, error) {
 	if err := cfg.APITLSConfig.Valid(); err != nil {
 		return nil, err
 	}
@@ -198,7 +198,7 @@ func newReportingOperator(
 	coordinatorClient coordinatorv1.CoordinationV1Interface,
 	meteringClient cbClientset.Interface,
 	informerNamespace string,
-) *Reporting {
+) ReportingOperator {
 	informerFactory := factory.NewSharedInformerFactoryWithOptions(meteringClient, defaultResyncPeriod, factory.WithNamespace(informerNamespace), factory.WithTweakListOptions(nil))
 
 	prestoTableInformer := informerFactory.Metering().V1().PrestoTables()
@@ -237,7 +237,7 @@ func newReportingOperator(
 	eventBroadcaster.StartRecordingToSink(&corev1.EventSinkImpl{Interface: kubeClient.Events(cfg.OwnNamespace)})
 	eventRecorder := eventBroadcaster.NewRecorder(scheme.Scheme, v1.EventSource{Component: cfg.Hostname})
 
-	op := &Reporting{
+	op := &defaultReportingOperator{
 		logger:            logger,
 		cfg:               cfg,
 		kubeConfig:        kubeConfig,
@@ -315,7 +315,7 @@ func newReportingOperator(
 	return op
 }
 
-func (op *Reporting) Run(ctx context.Context) error {
+func (op *defaultReportingOperator) Run(ctx context.Context) error {
 	var wg sync.WaitGroup
 	// buffered big enough to hold the errs of each server we start.
 	srvErrChan := make(chan error, 3)
@@ -655,7 +655,7 @@ func (op *Reporting) Run(ctx context.Context) error {
 	return nil
 }
 
-func (op *Reporting) newPrometheusConnFromURL(url string) (prom.API, error) {
+func (op *defaultReportingOperator) newPrometheusConnFromURL(url string) (prom.API, error) {
 	transportConfig := &transport.Config{}
 	if op.cfg.PrometheusConfig.CAFile != "" {
 		if _, err := os.Stat(op.cfg.PrometheusConfig.CAFile); err == nil {
@@ -694,7 +694,7 @@ func (op *Reporting) newPrometheusConnFromURL(url string) (prom.API, error) {
 	})
 }
 
-func (op *Reporting) startWorkers(ctx context.Context, wg *sync.WaitGroup) {
+func (op *defaultReportingOperator) startWorkers(ctx context.Context, wg *sync.WaitGroup) {
 	stopCh := ctx.Done()
 
 	startWorker := func(threads int, workerFunc func(id int)) {
@@ -746,20 +746,20 @@ func (op *Reporting) startWorkers(ctx context.Context, wg *sync.WaitGroup) {
 	})
 }
 
-func (op *Reporting) setInitialized() {
+func (op *defaultReportingOperator) setInitialized() {
 	op.initializedMu.Lock()
 	op.initialized = true
 	op.initializedMu.Unlock()
 }
 
-func (op *Reporting) isInitialized() bool {
+func (op *defaultReportingOperator) isInitialized() bool {
 	op.initializedMu.Lock()
 	initialized := op.initialized
 	op.initializedMu.Unlock()
 	return initialized
 }
 
-func (op *Reporting) newPrometheusConn(promConfig promapi.Config) (prom.API, error) {
+func (op *defaultReportingOperator) newPrometheusConn(promConfig promapi.Config) (prom.API, error) {
 	client, err := promapi.NewClient(promConfig)
 	if err != nil {
 		return nil, fmt.Errorf("can't connect to prometheus: %v", err)
--- a/pkg/operator/operator.go
+++ b/pkg/operator/operator.go
@@ -14,20 +14,21 @@ import (
 	"sync"
 	"time"
 
-	"github.com/davecgh/go-spew/spew"
-	"github.com/prestodb/presto-go-client/presto"
-	promapi "github.com/prometheus/client_golang/api"
-	prom "github.com/prometheus/client_golang/api/prometheus/v1"
-	"github.com/prometheus/client_golang/prometheus/promhttp"
-	log "github.com/sirupsen/logrus"
-	"github.com/taozle/go-hive-driver"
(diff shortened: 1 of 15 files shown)


pkg/operator/operator.go
```diff
@@ -10,24 +10,24 @@ import (
...
 
-	"github.com/davecgh/go-spew/spew"
-	"github.com/prestodb/presto-go-client/presto"
-	promapi "github.com/prometheus/client_golang/api"
-	prom "github.com/prometheus/client_golang/api/prometheus/v1"
-	"github.com/prometheus/client_golang/prometheus/promhttp"
-	log "github.com/sirupsen/logrus"
-	"github.com/taozle/go-hive-driver"
+	"github.com/kube-reporting/metering-operator/pkg/db"
+	cbClientset "github.com/kube-reporting/metering-operator/pkg/generated/clientset/versioned"
```
timflannagan1, line 17: If we're tackling this kind of work now, I would expect to also rename this import and all of its occurrences too, to something like `meteringClientset` or anything that doesn't involve the old "chargeback" name.

pkg/operator/operator.go
```diff
@@ -194,12 +103,10 @@ type Reporting struct {
 	importers   map[string]*prestostore.PrometheusImporter
 }
 
-func New(logger log.FieldLogger, cfg Config) (*Reporting, error) {
-	if err := cfg.APITLSConfig.Valid(); err != nil {
-		return nil, err
-	}
-	if err := cfg.MetricsTLSConfig.Valid(); err != nil {
-		return nil, err
+func New(logger log.FieldLogger, cfg Config) (ReportingOperator, error) {
```
timflannagan1, line 106: This is unrelated to the changes you've made so far to this constructor, but I would generally expect to see the initialization of all of the clientsets we may need throughout the operator, in something like cmd/reporting-operator/start.go instead of the actual constructor declaration. WDYT?

timflannagan1:
Something else that I would like to tackle, but is probably out of the scope of this kind of work, is improving (or adding) comments to types/methods/functions to the various reporting-operator components, like what does `New() ...` or `op.Run` achieve such that a person not familiar with controllers would be able to see what we're trying to achieve with those functions.

pkg/operator/operator.go
```diff
@@ -194,12 +103,10 @@ type Reporting struct {
...
+	errs := IsValidConfig(&cfg)
+	if errs != nil {
+		return nil, errs
 	}
```
timflannagan1, line 110: Building off my comment from above, I wonder if we should be doing all of the reporting configuration validation in the cmd/reporting-operator driver function that provides the entry point for this operator, rather than in the constructor.

pkg/operator/operator.go
pweil-: > If we're tackling this kind of work now

We don't have to merge this now, I think it should be no-op so it shouldn't be too risky but bugs first and we can merge these when master opens if that't the right thing to do.



> I would expect to also rename this import and all of its occurrences too, to something like `meteringClientset` or anything that doesn't involve the old "chargeback" name.

Easy enough, I'll add a new commit for this refactor

pkg/operator/operator.go
pweil-: > WDYT

Yes, after reducing the file I think there are a few more follow ons we can do.  (Side goal is to also keep these refactor commits small and reviewable so some may be better in follow on PRs)

I feel like a lot that happens in the private method could be moved into `New` or refactored into helpers that `New` just orchestrates for better readability so +1 to that suggestion

pweil-:
> improving (or adding) comments to types/methods/functions to the various reporting-operator components,

this is 100% a goal here.  Anything we touch in a refactor should get comments IMO, hence the API comments.

pkg/operator/operator.go
pweil-: it can be called from the `cmd` package but it should not be done there.  Commands in Kube/OpenShift follow a pattern of `Complete`, `Validate`, `Run`.  We should refactor ours to match and call the `IsValidConfig` method in the command's `Validate` method.  However we should keep it in new as well to protect against programmatic access to the operator no matter how it's started in the future.

pweil-: Example of the recommended pattern: https://github.com/openshift/oc/blob/a695d74ef1aee9a3f605d38dd8b6fae2062b63fc/pkg/cli/tag/tag.go#L93-L116

pweil-:
Some other observations for the discussion (perhaps this PR, perhaps future)

1.  figure out why we have so many threads in a single operator pod and if we can reduce and maintain performance (possibly removing the need for any mutex guarded code depending on the answer)

2.  reduce the configuration complexity where possible.  Example is the namespace configuration where we require `AllNamespaces` to be set with `TargetNamespaces` when we could just key on `TargetNamespaces` and do the right thing.  Same for the TLS configs with a boolean that turns it off or on when we could just have the existence of the certs enable it.

3.  move items that are ancillary to the operator into their own packages

pkg/operator/operator.go
pweil-: don't think I read this right the first time.  I would need to look a bit more at that suggestion.  I think I agree - let the operator take the client as configuration and let the caller determine how they want to create it.  All we really care about in the context of the operator is the interface methods, not how the client was instantiated.

pkg/operator/operator.go
```diff
@@ -57,7 +57,7 @@ type defaultReportingOperator struct {
 
 	kubeClient        corev1.CoreV1Interface
```
timflannagan1, line 58: As a side note, this also seems a bit strange to me as to why we would use a `corev1.Corev1Interface` instead of a `kubernetes.Interface`, but this isn't something we need to tackle here.

pkg/operator/operator.go
```diff
@@ -41,108 +41,17 @@ import (
...
-	LeaderLeaseDuration time.Duration
-
-	APITLSConfig     TLSConfig
-	MetricsTLSConfig TLSConfig
-	PrometheusConfig PrometheusConfig
-}
-
-type Reporting struct {
+// defaultReportingOperator is the concrete implementation of a ReportingOperator.
+type defaultReportingOperator struct {
```
bentito, line 54: Would it work to just call this `reportingOperator` with the lower case indicating its implementation status. I only ask because it's used so many places, it seems like keeping it shorter would be good and yet still code legible. I'm not clear that lower case would do that for Go, but I'm hoping.

timflannagan1: Lower case variables/functions/methods are for defining global scope throughout a package: https://tour.golang.org/basics/3

So here we're essentially saying you wouldn't have access to this `defaultReportingOperator` type if you're not interacting with the interfaces, instead of the explicit type.

pkg/operator/operator.go
timflannagan1: It looks like the MCO does something similar but abstracts it a bit more using a custom type: https://github.com/openshift/machine-config-operator/blob/master/cmd/machine-config-controller/start.go

pkg/operator/operator.go
pweil-: if the question is would `ReportingOperator` and `reportingOperator` types conflict the answer is no but I think that's probably too much mental gymnastics.  

I'm ok with shortening the name but didn't have an immediate go to.  I went with `defaultReportingOperator` in the vein of "if we made another one how would we name it?".  Probably something like `instrumentedReportingOperator` or `mockReportingOperator`.  I'm 100% not tied to the name though :smile:

I'll note that referencing it can/should be shorter like we do with `func (op *defaultReportingOperator)` so most of the interactions would simply be `op.Foo` not `defaultReportingOperator.Foo`.

pkg/operator/operator.go
bentito: Like `cubeClientSet`, maybe `kubeClientSet`?

pkg/operator/operator.go
pweil-: I'm going to start an issue with a list of these :smile:

pweil-: https://github.com/kube-reporting/metering-operator/issues/1357

pkg/operator/operator.go
bentito: Sound reasonable.

pweil-:
```
time="08-04-2020 13:35:39" level=debug msg="Created the mysql namespace" context=metering-validhdfs-mysqldatabase
error: multiple images or templates matched "mysql:5.7"

The argument "mysql:5.7" could apply to the following Docker images, OpenShift image streams, or templates:

* Image stream "mysql" (tag "8.0") in project "openshift"
  Use --image-stream="openshift/mysql:8.0" to specify this image or template

* Image stream "mysql" (tag "latest") in project "openshift"
  Use --image-stream="openshift/mysql:latest" to specify this image or template
```

timflannagan1:
> ```
> time="08-04-2020 13:35:39" level=debug msg="Created the mysql namespace" context=metering-validhdfs-mysqldatabase
> error: multiple images or templates matched "mysql:5.7"
> 
> The argument "mysql:5.7" could apply to the following Docker images, OpenShift image streams, or templates:
> 
> * Image stream "mysql" (tag "8.0") in project "openshift"
>   Use --image-stream="openshift/mysql:8.0" to specify this image or template
> 
> * Image stream "mysql" (tag "latest") in project "openshift"
>   Use --image-stream="openshift/mysql:latest" to specify this image or template
> ```

https://github.com/kube-reporting/metering-operator/pull/1353

pweil-:
Removing WIP from this - it's probably big enough that we shouldn't add anything else too it yet if we want to merge.  I am going to call out a few pieces (in review comments) that I think we need to be aware of though before we merge.

pkg/operator/validation.go
```diff
@@ -0,0 +1,178 @@
...
+	}
+	return nil
+}
+
+// isValidHostPort attempts to split a non empty hp into host and port, returning any errors found.
+// TODO this is only validating non-empty strings.  We may want to check for empty strings an report errors.
+// TODO this requires a port to be specified, is that one of our requirements?
+func isValidHostPort(hp string, name string) error {
+	if len(hp) > 0 {
+		if _, _, err := net.SplitHostPort(hp); err != nil {
```
pweil-, line 173: this will fail if a port is not given - can someone confirm that that is a correct behavior?

pkg/operator/validation.go
```diff
@@ -0,0 +1,178 @@
...
+	if cfg.PrometheusConfig.CAFile != "" {
+		if _, err := os.Stat(cfg.PrometheusConfig.CAFile); err != nil {
+			errs = append(errs, err)
+		}
+	}
+
+	// PrometheusDataSourceMaxBackfillImportDuration overrides PrometheusDataSourceGlobalImportFromTime
+	// don't set both.
+	if cfg.PrometheusDataSourceGlobalImportFromTime != nil && cfg.PrometheusDataSourceMaxBackfillImportDuration > 0 {
+		errs = append(errs, fmt.Errorf("prometheusDataSourceGlobalImportFromTime and prometheusDataSourceMaxBackfillImportDuration cannot both be set"))
```
pweil-, line 146: previously we stated that one overrides the other.  Introducing this will cause explicit failure.  Acceptable?

pkg/operator/validation.go
```diff
@@ -0,0 +1,178 @@
...
+	return nil
+}
+
+// IsValidHiveConfig ensure all Hive* fields are valid if provided.
+func IsValidHiveConfig(cfg *Config) error {
+	errs := []error{}
+
+	errs = append(errs, isValidHostPort(cfg.HiveHost, "hiveHost"))
+
+	if !cfg.HiveUseTLS {
```
pweil-, line 97: previously we allowed config to be set even if this was set to false.  Should we continue?

pkg/operator/validation.go
```diff
@@ -0,0 +1,178 @@
...
+	return nil
+}
+
+// IsValidPrestoConfig ensure all Presto* fields are valid if provided.
+func IsValidPrestoConfig(cfg *Config) error {
+	errs := []error{}
+
+	errs = append(errs, isValidHostPort(cfg.PrestoHost, "prestoHost"))
+
+	if !cfg.PrestoUseTLS {
```
pweil-, line 64: previously we allowed config to be set even if this was set to false.  Should we continue?

timflannagan1: I would imagine that makes sense, i.e. rejecting a configuration where you explicitly set TLS-related flags yet disable TLS entirely for the reporting-operator -> presto communications.

pkg/operator/validation.go
timflannagan1: Same as my comment below. I think this is fine UX, reject any configurations like this if you're disabling TLS on the reporting-operator side for that component.

pkg/operator/validation.go
timflannagan1: I'm not sure where I land on this validation check. I think it sounds reasonable looking at the codebase and making sure we reject configurations that won't be respected. @bentito any opinion on this validation check.

pweil-: @bentito - thoughts on this one?

bentito: Reading the usage in `start.go`, it seems like this validates the stated usage. If it's Prom import question, they're both just ways of getting at how far back to try to grab Prom data when metering somehow gets out of sync. I think `PrometheusDataSourceGlobalImportFromTime` has more potential to be pretty awful for cluster workload, but I don't think that's being asked here.

pkg/operator/validation.go
bentito: This gets called on api, metrics, pprof, presto, hive
 
I think it's valid to make the admin config the port for our API.
Metrics = Prom, also has a default.
pprof seems like it has a default port as well, so maybe okay to not spec.
Presto has a default port, might be ok to not spec.
Hive has a default thrift service port as well, so maybe that might not need being requeired to specify either.

pweil-: Ok, I will remove the calls/tests for apis that have default ports.

pkg/operator/validation.go
pweil-: just validation of input. Sounds like I can leave it.  Thanks

pweil-:
@bentito @timflannagan1 updates from the review in the last commit.  Thanks.

timflannagan1:
@pweil- @bentito where are we at with this? There's nothing in here that I think we should hold off until 4.7 master opens up so merging this sooner than later is probably best.

pweil-:
@timflannagan1 all review comments from @bentito have been addressed (left the `prometheusDataSourceGlobalImportFromTime`/`prometheusDataSourceMaxBackfillImportDuration` and removed the non-api host/port validations and tests).

Unless there is anything else I think this can merge.

timflannagan1:
:tada:

PR #1366, opened 2020-08-10
olm_deploy: Use the ubi8-minimal base image for the registry Dockerfile.

timflannagan1:
This updates the base image to use ubi8 instead of the previous Golang image, which helps significantly cut down the size of this image and we don't need much more besides a basic linux environment to utilize this image throughout the testing suite.


timflannagan1:
This doesn't touch anything upgrade-related, and that test was a flake.

PR #1367, opened 2020-08-10
e2e: report expires with no dependency & does not expire with dependency

bentito:
We need to test via e2e testing, the following situations:

- [X] Does a report with no dependencies with an expiration set, properly expire?
- [x] Does a report that is depended on by another report not get deleted despite reaching its expiration date?
~- [ ] Does a report that is depended on by a ReportQuery not get deleted despite reaching its expiration date?~
[We should add a test on the last item in another pass]

### DIFF
--- a/pkg/operator/http_test.go
+++ b/pkg/operator/http_test.go
@@ -118,7 +118,7 @@ func TestAPIV1ReportsGet(t *testing.T) {
 	}{
 		"report-finished-no-results": {
 			reportName: testReportName,
-			report:     testhelpers.NewReport(testReportName, namespace, testQueryName, reportStart, reportEnd, metering.ReportStatus{}, nil, false, nil),
+			report:     testhelpers.NewReport(testReportName, namespace, testQueryName, nil, reportStart, reportEnd, metering.ReportStatus{}, nil, false, nil),
 			query: testhelpers.NewReportQuery(testQueryName, namespace, []metering.ReportQueryColumn{
 				{
 					Name: "timestamp",
@@ -146,7 +146,7 @@ func TestAPIV1ReportsGet(t *testing.T) {
 		},
 		"report-finished-with-results": {
 			reportName: testReportName,
-			report:     testhelpers.NewReport(testReportName, namespace, testQueryName, reportStart, reportEnd, metering.ReportStatus{}, nil, false, nil),
+			report:     testhelpers.NewReport(testReportName, namespace, testQueryName, nil, reportStart, reportEnd, metering.ReportStatus{}, nil, false, nil),
 			query: testhelpers.NewReportQuery(testQueryName, namespace, []metering.ReportQueryColumn{
 				{
 					Name: "timestamp",
@@ -187,7 +187,7 @@ func TestAPIV1ReportsGet(t *testing.T) {
 		},
 		"report-finished-db-errored": {
 			reportName: testReportName,
-			report:     testhelpers.NewReport(testReportName, namespace, testQueryName, reportStart, reportEnd, metering.ReportStatus{}, nil, false, nil),
+			report:     testhelpers.NewReport(testReportName, namespace, testQueryName, nil, reportStart, reportEnd, metering.ReportStatus{}, nil, false, nil),
 			query: testhelpers.NewReportQuery(testQueryName, namespace, []metering.ReportQueryColumn{
 				{
 					Name: "timestamp",
@@ -232,7 +232,7 @@ func TestAPIV1ReportsGet(t *testing.T) {
 		},
 		"mismatched-results-schema-to-table-schema": {
 			reportName: testReportName,
-			report:     testhelpers.NewReport(testReportName, namespace, testQueryName, reportStart, reportEnd, metering.ReportStatus{}, nil, false, nil),
+			report:     testhelpers.NewReport(testReportName, namespace, testQueryName, nil, reportStart, reportEnd, metering.ReportStatus{}, nil, false, nil),
 			query: testhelpers.NewReportQuery(testQueryName, namespace, []metering.ReportQueryColumn{
 				{
 					Name: "timestamp",
@@ -394,7 +394,7 @@ func TestAPIV2ReportsFull(t *testing.T) {
 	}{
 		"report-finished-with-results": {
 			reportName: testReportName,
-			report:     testhelpers.NewReport(testReportName, namespace, testQueryName, reportStart, reportEnd, metering.ReportStatus{}, nil, false, nil),
+			report:     testhelpers.NewReport(testReportName, namespace, testQueryName, nil, reportStart, reportEnd, metering.ReportStatus{}, nil, false, nil),
 			apiPath:    apiReportV2URLFull(namespace, testReportName) + testFormat,
 			query: testhelpers.NewReportQuery(testQueryName, namespace, []metering.ReportQueryColumn{
 				{
@@ -444,7 +444,7 @@ func TestAPIV2ReportsFull(t *testing.T) {
 		},
 		"report-finished-no-results": {
 			reportName: testReportName,
-			report:     testhelpers.NewReport(testReportName, namespace, testQueryName, reportStart, reportEnd, metering.ReportStatus{}, nil, false, nil),
+			report:     testhelpers.NewReport(testReportName, namespace, testQueryName, nil, reportStart, reportEnd, metering.ReportStatus{}, nil, false, nil),
 			apiPath:    apiReportV2URLFull(namespace, testReportName) + testFormat,
 			query: testhelpers.NewReportQuery(testQueryName, namespace, []metering.ReportQueryColumn{
 				{
@@ -475,7 +475,7 @@ func TestAPIV2ReportsFull(t *testing.T) {
 		},
 		"report-finished-db-errored": {
 			reportName: testReportName,
-			report:     testhelpers.NewReport(testReportName, namespace, testQueryName, reportStart, reportEnd, metering.ReportStatus{}, nil, false, nil),
+			report:     testhelpers.NewReport(testReportName, namespace, testQueryName, nil, reportStart, reportEnd, metering.ReportStatus{}, nil, false, nil),
 			apiPath:    apiReportV2URLFull(namespace, testReportName) + testFormat,
 			query: testhelpers.NewReportQuery(testQueryName, namespace, []metering.ReportQueryColumn{
 				{
@@ -524,7 +524,7 @@ func TestAPIV2ReportsFull(t *testing.T) {
 		},
 		"report-format-not-specified": {
 			reportName:            testReportName,
-			report:                testhelpers.NewReport(testReportName, namespace, testQueryName, reportStart, reportEnd, metering.ReportStatus{}, nil, false, nil),
+			report:                testhelpers.NewReport(testReportName, namespace, testQueryName, nil, reportStart, reportEnd, metering.ReportStatus{}, nil, false, nil),
 			apiPath:               apiReportV2URLFull(namespace, testReportName),
 			expectedStatusCode:    http.StatusBadRequest,
 			expectedAPIError:      "the following fields are missing or empty: format",
@@ -533,7 +533,7 @@ func TestAPIV2ReportsFull(t *testing.T) {
 		},
 		"report-format-non-existent": {
 			reportName:            testReportName,
-			report:                testhelpers.NewReport(testReportName, namespace, testQueryName, reportStart, reportEnd, metering.ReportStatus{}, nil, false, nil),
+			report:                testhelpers.NewReport(testReportName, namespace, testQueryName, nil, reportStart, reportEnd, metering.ReportStatus{}, nil, false, nil),
 			apiPath:               apiReportV2URLFull(namespace, testReportName) + "?format=doesntexist",
 			expectedStatusCode:    http.StatusBadRequest,
 			expectedAPIError:      "format must be one of: csv, json or tabular",
@@ -542,7 +542,7 @@ func TestAPIV2ReportsFull(t *testing.T) {
 		},
 		"mismatched-results-schema-to-table-schema": {
 			reportName: testReportName,
-			report:     testhelpers.NewReport(testReportName, namespace, testQueryName, reportStart, reportEnd, metering.ReportStatus{}, nil, false, nil),
+			report:     testhelpers.NewReport(testReportName, namespace, testQueryName, nil, reportStart, reportEnd, metering.ReportStatus{}, nil, false, nil),
 			apiPath:    apiReportV2URLFull(namespace, testReportName) + testFormat,
 			query: testhelpers.NewReportQuery(testQueryName, namespace, []metering.ReportQueryColumn{
 				{
@@ -682,7 +682,7 @@ func TestAPIV2ReportsTable(t *testing.T) {
 	}{
 		"report-finished-with-results": {
 			reportName: testReportName,
-			report:     testhelpers.NewReport(testReportName, namespace, testQueryName, reportStart, reportEnd, metering.ReportStatus{}, nil, false, nil),
+			report:     testhelpers.NewReport(testReportName, namespace, testQueryName, nil, reportStart, reportEnd, metering.ReportStatus{}, nil, false, nil),
 			apiPath:    apiReportV2URLTable(namespace, testReportName) + testFormat,
 			query: testhelpers.NewReportQuery(testQueryName, namespace, []metering.ReportQueryColumn{
 				{
@@ -734,7 +734,7 @@ func TestAPIV2ReportsTable(t *testing.T) {
 		},
 		"report-finished-no-results": {
 			reportName: testReportName,
-			report:     testhelpers.NewReport(testReportName, namespace, testQueryName, reportStart, reportEnd, metering.ReportStatus{}, nil, false, nil),
+			report:     testhelpers.NewReport(testReportName, namespace, testQueryName, nil, reportStart, reportEnd, metering.ReportStatus{}, nil, false, nil),
 			apiPath:    apiReportV2URLTable(namespace, testReportName) + testFormat,
 			query: testhelpers.NewReportQuery(testQueryName, namespace, []metering.ReportQueryColumn{
 				{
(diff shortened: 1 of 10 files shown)


test/e2e/report_dynamic_input_data_test.go
```diff
@@ -151,7 +158,7 @@ func testReportsProduceData(t *testing.T, testReportingFramework *reportingframe
 
 			t.Logf("reportStart: %s, reportEnd: %s", reportStart, reportEnd)
 
-			report := test.newReportFunc(test.name, test.queryName, test.schedule, &reportStart, &reportEnd)
+			report := test.newReportFunc(test.name, test.queryName, test.schedule, &reportStart, &reportEnd, nil)
```
EmilyM1, line 161: Is the nil needed?

test/e2e/report_dynamic_input_data_test.go
```diff
@@ -163,3 +170,145 @@ func testReportsProduceData(t *testing.T, testReportingFramework *reportingframe
...
+func testReportIsDeletedWhenNoDeps(t *testing.T, testReportingFramework *reportingframework.ReportingFramework) {
+	// cron schedule to run every minute
+	cronSchedule := &metering.ReportSchedule{
+		Period: metering.ReportPeriodCron,
+		Cron: &metering.ReportScheduleCron{
+			Expression: "*/1 * * * *",
+		},
+	}
+	start := time.Now()
+	var foundOnce = false
```
timflannagan1, line 183: You can just specify `var foundOnce bool` here as Golang has this notion of default types for empty variables.

test/e2e/report_dynamic_input_data_test.go
bentito: Yes, because I updated `newReportFunc` to take an expiration so all the pre-existing usages needed to be updated and "Go does not have optional parameters nor does it support method overloading"

test/e2e/report_dynamic_input_data_test.go
bentito: ah, yeah simpler, thanks.

test/testhelpers/helpers.go
```diff
@@ -39,6 +51,7 @@ func NewReport(name, namespace, testQueryName string, reportStart, reportEnd *ti
 		},
 		Spec: metering.ReportSpec{
 			QueryName:      testQueryName,
+			Inputs: inputs,
```
timflannagan1, line 54: You may need to setup automatic formatting on save in your IDE. I think the `make verify` check will complain about this as it's going to run `gofmt -w ...` and the `inputs` variable will be re-indented with the rest of the variables.

test/e2e/report_dynamic_input_data_test.go
```diff
@@ -163,3 +169,129 @@ func testReportsProduceData(t *testing.T, testReportingFramework *reportingframe
...
+		return &v
+	}
+
+	report := testhelpers.NewReport(
+		"report-depends-on-subreport",
+		testReportingFramework.Namespace,
+		testQueryName,
+		[]metering.ReportQueryInputValue{
+			metering.ReportQueryInputValue{
+				Name: "Report",
```
timflannagan1, line 261: This is [not a valid input](https://storage.googleapis.com/origin-ci-test/pr-logs/pull/kube-reporting_metering-operator/1367/pull-ci-kube-reporting-metering-operator-master-metering-e2e-aws/1301934731100688384/artifacts/metering-e2e-aws/metering_test_output/e2e/ReportExpiration/logs/namespaces/metering-reportexpiration/metering.openshift.io/reports/report-depends-on-subreport.yaml).

test/e2e/report_dynamic_input_data_test.go
```diff
@@ -163,3 +169,129 @@ func testReportsProduceData(t *testing.T, testReportingFramework *reportingframe
...
+			}
+		}
+		// if we're exiting due to timeout, it's correct, because the subReport should not be deleted
+		return false, nil
+	})
+	assert.True(t, strings.Contains(
+		err.Error(),
+		"timed out waiting"),
+		"should have timed out waiting on subReport expiration as subReport should not have been deleted",
+	)
```
timflannagan1, line 296: We may want to also validate that an event got created with the `ExpiredReportHasDependencies` warning type.

test/e2e/report_dynamic_input_data_test.go
```diff
@@ -163,3 +169,129 @@ func testReportsProduceData(t *testing.T, testReportingFramework *reportingframe
...
+	err := wait.Poll(time.Second*5, waitTimeWhileCheckDeletion, func() (bool, error) {
+		_, err := testReportingFramework.GetMeteringReport("report-should-delete")
+		// `foundOnce` provides a latch that allows us to wait for report creation then exit on deletion
+		if err == nil {
+			foundOnce = true
+		}
+		if err != nil {
+			if foundOnce && strings.Contains(err.Error(), "not found") {
+				return true, errors.New("report was deleted after being created")
+			}
```
timflannagan1, line 209: We should avoid doing the string comparison ourselves and just let the `apierrors.IsNotFound(err)` return the correct boolean value.

test/e2e/report_dynamic_input_data_test.go
```diff
@@ -163,3 +169,129 @@ func testReportsProduceData(t *testing.T, testReportingFramework *reportingframe
...
+	)
+	t.Logf("creating report %s and waiting %s to finish", report.Name, waitTimeWhileCheckDeletion)
+	createErr := testReportingFramework.CreateMeteringReport(report)
+	if createErr != nil {
+		assert.Failf(t, "Creating report should not have an error", createErr.Error())
+	}
+	err := wait.Poll(time.Second*5, waitTimeWhileCheckDeletion, func() (bool, error) {
+		_, err := testReportingFramework.GetMeteringReport("report-should-delete")
+		// `foundOnce` provides a latch that allows us to wait for report creation then exit on deletion
+		if err == nil {
```
timflannagan1, line 203: Same here, we should use `if apierrors.IsAlreadyExists(err) { foundOnce = true }`.

test/e2e/report_dynamic_input_data_test.go
```diff
@@ -163,3 +169,129 @@ func testReportsProduceData(t *testing.T, testReportingFramework *reportingframe
...
+		},
+	}
+	var foundOnce = false
+	expiration := &metav1.Duration{Duration: 30.0 * time.Second}
+	// If a subReport is in dependency unmet state it won't be deleted and it can take awhile for import to happen.
+	// 4 minutes seems to be enough time for that setup, and the delete, to reliably happen.
+	waitTimeWhileCheckDeletion := 4 * time.Minute
+
+	subReport := testReportingFramework.NewSimpleReport(
+		"subreport-should-not-delete",
```
timflannagan1, line 236: Can we make this a const variable like `testQueryName` even if we're not reusing it for readability?

test/e2e/report_dynamic_input_data_test.go
```diff
@@ -163,3 +169,129 @@ func testReportsProduceData(t *testing.T, testReportingFramework *reportingframe
...
+		nil,
+		nil,
+		expiration,
+	)
+	t.Logf("creating subReport %s and waiting %s to finish", subReport.Name, waitTimeWhileCheckDeletion)
+
+	createErr := testReportingFramework.CreateMeteringReport(subReport)
+	if createErr != nil {
+		assert.Failf(t, "Creating subReport should not have an error", createErr.Error())
+	}
```
timflannagan1, line 248: `assert.NoError(t, err, "creating sub report should produce no error")`

test/e2e/report_dynamic_input_data_test.go
```diff
@@ -163,3 +169,129 @@ func testReportsProduceData(t *testing.T, testReportingFramework *reportingframe
...
+		metering.ReportStatus{},
+		cronSchedule,
+		false,
+		nil,
+	)
+
+	createErr = testReportingFramework.CreateMeteringReport(report)
+	if createErr != nil {
+		assert.Failf(t, "Creating report should not have an error", createErr.Error())
+	}
```
timflannagan1, line 276: `assert.NoError(t, err, "creating report should produce no error)`

test/e2e/report_dynamic_input_data_test.go
```diff
@@ -163,3 +169,129 @@ func testReportsProduceData(t *testing.T, testReportingFramework *reportingframe
...
+
+	err := wait.Poll(time.Second*5, waitTimeWhileCheckDeletion, func() (bool, error) {
+		_, err := testReportingFramework.GetMeteringReport("subReport-should-not-delete")
```
timflannagan1, line 279: Ah yeah, looks like we use this in multiple places (make a comment earlier in the review). Let's make this a variable instead of hard coding it.

test/e2e/report_dynamic_input_data_test.go
```diff
@@ -163,3 +169,129 @@ func testReportsProduceData(t *testing.T, testReportingFramework *reportingframe
...
+		// `foundOnce` provides a latch that allows us to wait for subReport creation then exit on deletion
+		if err == nil {
+			foundOnce = true
+		}
+		if err != nil {
+			if foundOnce && strings.Contains(err.Error(), "not found") {
+				return true, errors.New("subReport was deleted after being created")
+			}
+		}
```
timflannagan1, line 288: Same comment here as above - let's use the `apierrors` package to do the error handling and string comparisons for us.

test/e2e/report_dynamic_input_data_test.go
timflannagan1: We can also use [`assert.Contains`](https://godoc.org/github.com/stretchr/testify/assert#Contains) here instead of doing a blanket `assert.True` that's difficult to read.

test/testhelpers/helpers.go
timflannagan1, line 54: Same comment as before. This is going to fail the verify check until you run gofmt locally and recommit.

bentito: I just need to push the format buttons more :-) thanks!

test/e2e/report_dynamic_input_data_test.go
bentito: I tried swapping out a valid input, `NamespaceCPUUsageReportName` and that yields a valid report, so how would I programmatically link the sub Report to that input then is what I am still wondering.

test/e2e/report_dynamic_input_data_test.go
bentito: ah I like a typed check better than the strings checking anyway.

test/e2e/report_dynamic_input_data_test.go
bentito: fixed in both locations

test/e2e/report_dynamic_input_data_test.go
bentito: changed in all 3 locations

test/e2e/report_dynamic_input_data_test.go
bentito: yeah, esp. cuz I got the casing wrong!

test/e2e/report_dynamic_input_data_test.go
EmilyM1: That seems sensible

pkg/operator/http_test.go
```diff
@@ -118,7 +118,7 @@ func TestAPIV1ReportsGet(t *testing.T) {
 	}{
 		"report-finished-no-results": {
 			reportName: testReportName,
-			report:     testhelpers.NewReport(testReportName, namespace, testQueryName, reportStart, reportEnd, metering.ReportStatus{}, nil, false, nil),
+			report:     testhelpers.NewReport(testReportName, namespace, testQueryName, nil, reportStart, reportEnd, metering.ReportStatus{}, nil, false, nil),
```
EmilyM1, line 121: Why the nil before reportStart parameter?

EmilyM1: Looks like reportStart is the first key in the TestAPIV1ReportsGet struct.

test/e2e/report_dynamic_input_data_test.go
```diff
@@ -163,3 +170,125 @@ func testReportsProduceData(t *testing.T, testReportingFramework *reportingframe
...
+		if err == nil {
+			foundOnce = true
+		}
+		if err != nil {
+			if foundOnce && apierrors.IsNotFound(err) {
+				return true, errors.New("subReport was deleted after being created")
+			}
+		}
+		// if we're exiting due to timeout, it's correct, because the subReport should not be deleted
+		return false, nil
```
timflannagan1, line 287:
```go
_, err := testReportingFramework.GetMeteringReport(subReportName)
if apierrors.IsAlreadyExists(err) {
    foundOnce = true
    return false, nil
}
...

File: test/e2e/report_dynamic_input_data_test.go [hunk 15]
```diff
@@ -291,4 +292,19 @@ func testReportIsNotDeletedWhenReportDependsOnIt(t *testing.T, testReportingFram
 		"timed out waiting"),
 		"should have timed out waiting on subReport expiration as subReport should not have been deleted",
 	)
+	listOptions := metav1.ListOptions{}
```
timflannagan1, line 295:
```suggestion
	listOptions := metav1.ListOptions{
	    FieldSelector: "involvedObject.Name="subreport-should-not-delete,type=Warning"
	```
...
require.NoError(t, err, "failed to list the events in the test namespace")
require.GreaterOrEqual(len(events.Items), 1, "expected the list of events matching the %s field selector to be greater than zero")
```
```

File: test/e2e/main_test.go [hunk 16]
```diff
@@ -366,6 +366,29 @@ func TestManualMeteringInstall(t *testing.T) {
...
+					TestFunc:     testReportIsNotDeletedWhenReportDependsOnIt,
+					ExtraEnvVars: []string{},
+				},
+				//{
+				//	Name:     "testReportIsNotDeletedWhenReportQueryDependsOnIt",
+				//	TestFunc: testReportIsNotDeletedWhenReportQueryDependsOnIt,
+				//},
+			},
+			MeteringConfigManifestFilename: "prometheus-metrics-importer-enabled.yaml",
+		},
```
timflannagan1, line 391: Once we get closer to merging this, I don't think this kind of testing coverage requires it's own dedicated test installation. Should we deflate those sub-tests into existing test installations?

test/e2e/report_dynamic_input_data_test.go
bentito: done in latest commit. though I couldn't quickly work out the filtering. I don't think there are many events to loop on in test so not a big deal, IMO

test/e2e/report_dynamic_input_data_test.go
bentito: yep, better, trying now.

test/e2e/main_test.go
bentito: agreed, I'll move them into another top level test.

pkg/operator/http_test.go
bentito: This PR adds `inputs` in that slot.

test/e2e/report_dynamic_input_data_test.go
bentito: returning false for the condition function on `Poll` means stop polling, not what we want here. 

Also, I don't find the api call more readable than just the check to see if there is no err directly, it's certainly slower.

test/e2e/report_dynamic_input_data_test.go
bentito:
```                	Error:      	Received unexpected error:
                	            	field label not supported: involvedObject.Name```

Comment (bentito) [reply test/e2e/report_dynamic_input_data_test.go:295]: Going to leave as is, there are not many events to loop through.

File: test/e2e/main_test.go (hunk 16 above)
Comment (bentito) [reply test/e2e/main_test.go:391]: Going with sticking them under: `ValidHDFS-ReportStaticInputData` 

FWIW: None of the top-level test "suite" names make much sense to me.

RANT: Go's test infra is primitive, making coders look through a long struct for tests and subtests is bogus.

Comment (timflannagan1) [reply test/e2e/main_test.go]: Yeah, agreed on the naming of tests doesn't make much sense. I'm under the impression that we were going to have tests that caught some of the common static validation errors we were seeing (e.g. `spec.storage` wasn't properly configured) but never got to that part in the e2e suite. I would also expect that kind of validation to live in its own top-level tests, so we can rename any of the `Valid*-<Storage>-...` tests.

But for the go test infra part, most of the test code is custom logic and we only utilize a portion of the stdlib testing suite, and the recommendation to use table driven tests.

File: test/e2e/manifests/meteringconfigs/prometheus-metrics-importer-enabled.yaml [hunk 17]
```diff
@@ -39,6 +39,8 @@ spec:
           requests:
             cpu: 1
             memory: 1Gi
+      image:
+        tag: "4.5"
```
timflannagan1, line 43: Remove this

test/e2e/report_dynamic_input_data_test.go
```diff
@@ -163,3 +171,139 @@ func testReportsProduceData(t *testing.T, testReportingFramework *reportingframe
...
+	testQueryName := "namespace-cpu-usage"
+	subReportName := "subreport-should-not-delete"
+	// cron schedule to run every minute
+	cronSchedule := &metering.ReportSchedule{
+		Period: metering.ReportPeriodCron,
+		Cron: &metering.ReportScheduleCron{
+			Expression: "*/1 * * * *",
+		},
+	}
+	var foundOnce = false
```
timflannagan1, line 231: `var foundOnce bool`

test/e2e/manifests/meteringconfigs/prometheus-metrics-importer-enabled.yaml
bentito: whoops, yep.

test/e2e/report_dynamic_input_data_test.go
bentito: yep

timflannagan1:
Not entirely sure why tide is retesting all these jobs.

PR #1368, opened 2020-08-11
test: Upgrade from metering version that contains the 4.5 many-to-many fix.

timflannagan1:
We're seeing consistent upgrade failures due to the 4.5 metering-ansible-operator version in the redhat-operators CatalogSource custom resource not containing the PVC-related many-to-many fix. As a result, upgrade tests will fail as the pre-upgrade tests are reporting errors due to metrics not being correctly important (namely due to the aforementioned many-to-many bug), but the actual upgrade and tests against that upgrade are succeeding.

### DIFF
--- a/test/e2e/metering_upgrade_test.go
+++ b/test/e2e/metering_upgrade_test.go
@@ -10,6 +10,7 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
+	"github.com/kube-reporting/metering-operator/test/deployframework"
 	"github.com/kube-reporting/metering-operator/test/reportingframework"
 	"github.com/kube-reporting/metering-operator/test/testhelpers"
 )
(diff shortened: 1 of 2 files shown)


test/e2e/metering_upgrade_test.go
```diff
@@ -18,8 +19,9 @@ const (
 	// TODO: support upgrading from a custom registry image instead
 	// of defaulting to the redhat-operators CatalogSource and waiting
 	// for GA-ed channels to be exposed.
-	upgradeFromCatalogSource          = "redhat-operators"
+	upgradeFromCatalogSource          = "custom-metering-image"
 	upgradeFromCatalogSourceNamespace = "openshift-marketplace"
+	upgradeFromImage                  = "quay.io/tflannag/metering-index:4.5"
```
timflannagan1, line 24: This is an index image containing a single bundle image where that bundle image contains a 4.5 metering CSV with an overridden metering-ansible-operator image (which contains the many-to-many 4.5 fixes).

timflannagan1:
Test failed due to e2e flake, and these changes don't touch anything that requires another spin-up of the e2e suite.

PR #1371, opened 2020-08-11
olm_deploy,test: Add tenancy to registry-related resources.

timflannagan1:
We're currently running into the problem where multiple metering e2e runs are conflicting with the registry-related resources we're spinning up for developers. In the case of the service object that gets created, we're only selecting on a single Pod label, and in the case where there's multiple registry deployments that all use the same label selector, we risk running e2e tests against a different version of metering than what we expected. This updates the static label selectors we use for the registry deployment and service, adding a `tenancy` label selector that we will populate at runtime after decoding the YAML manifest into a Go object.

### DIFF
--- a/olm_deploy/manifests/deployment.yaml
+++ b/olm_deploy/manifests/deployment.yaml
@@ -4,15 +4,18 @@ metadata:
   name: metering-ansible-operator-registry
   labels:
     registry.operator.metering-ansible: "true"
+    tenant: ""
 spec:
   replicas: 1
   selector:
     matchLabels:
       registry.operator.metering-ansible: "true"
+      tenant: ""
   template:
     metadata:
       labels:
         registry.operator.metering-ansible: "true"
+        tenant: ""
       name: metering-ansible-operator-registry
     spec:
       initContainers:
--- a/test/deployframework/helpers.go
+++ b/test/deployframework/helpers.go
@@ -615,6 +615,7 @@ func CreateRegistryDeployment(
 	deployment.Namespace = namespace
 	if deployment.Labels != nil {
 		deployment.Labels["name"] = namespacePrefix + "-" + testNamespaceLabel
+		deployment.Labels["tenant"] = namespacePrefix
 	}
 
 	// Update the invalid manifest registry image with the @registryImage provided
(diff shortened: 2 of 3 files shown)


test/deployframework/helpers.go
```diff
@@ -642,6 +643,11 @@ func CreateRegistryDeployment(
 		logger.Infof("Overriding the default metering-reporting-operator image with %s", reportingOperatorImage)
 		deployment.Spec.Template.Spec.InitContainers[0].Env[1].Value = reportingOperatorImage
 	}
+	// Populate the tenant label throughout the deployment manifest
+	// that we're currently build up. By default, that label's value
+	// is set to "", so we're injecting the namespacePrefix before creation.
+	deployment.Spec.Selector.MatchLabels["tenant"] = namespacePrefix
+	deployment.Spec.Template.ObjectMeta.Labels["tenant"] = namespacePrefix
```
timflannagan1, line 650: This is probably a good indication that we don't really need to be using a deployment to roll out a single registry Pod as we will never need to scale up additional replicas, but I think that's out of scope out of this PR anyways.

bentito:
Does this have a companion PR that actually checks the tenancy label, or how's that work?

timflannagan1:
@bentito I'm not sure I follow what you're asking here. This PR just introduces some labels that fix issues when two devs are running e2e on a shared cluster.

bentito:
Where does the new label get used? Like is that a special label that means something to the release/CI infrastructure?

timflannagan1:
> Where does the new label get used? Like is that a special label that means something to the release/CI infrastructure?

We use these resources to deploy the most up-to-date version of metering in CI (e.g. override any of the default metering-related images in the CSV with the ones that CI built, or any images passed by a dev to the e2e suite will get overridden also). In order to do this, we use some of the operator-registry binaries to package this version of the operator in a sqlite database and expose that database using a gRPC interface, all within this deployment object. We then use this service to route traffic to any Pods that match the `registry.operator.metering-ansible: "true"` label, which in the case where there are multiple devs running e2e locally, will potentially route traffic to a Pod that contains a different version of metering you intended. We can fix that by adding an additional label that helps distinguish between two versions of metering that got deployed locally.

olm_deploy/manifests/deployment.yaml
```diff
@@ -4,15 +4,18 @@ metadata:
   name: metering-ansible-operator-registry
   labels:
     registry.operator.metering-ansible: "true"
+    tenant: ""
```
timflannagan1, line 7: This could probably be changed to match the above label and have a `registry.*` prefix too.

PR #1374, opened 2020-08-14
hack: Grab the HiveTable custom resources in a testing namespace.

timflannagan1:


timflannagan1:
Those upgrade tests failures appear to be potentially problematic. Going to retest again as this PR doesn't affect from functionality but we should continue monitoring that test.
/retest

PR #1375, opened 2020-08-17
Dockerfile: Remove the inotify-tools yum package.

timflannagan1:
This package was used to feed Ansible logs into the ansible sidecar container that was removed earlier this release cycle as the ansible-operator removed it from their base image (as inotify-tools isn't in UBI8).


timflannagan1:
Relates to [openshift/release #10852](https://github.com/openshift/release/pull/10852) as we migrate towards consolidating the CI and ART Dockerfiles.

PR #1376, opened 2020-08-18
Dockerfile: Remove the exec-entrypoint ansible-operator subcommand.

timflannagan1:
It looks like the `exec-entrypoint` sub-command was removed from the ansible-operator during the 0.19.x release but a poor 4.6 rebase for ocp-operator-sdk resulted in that field still being used. Here is the updated version after that's been corrected downstream: https://github.com/openshift/ocp-release-operator-sdk/blob/master/release/ansible/bin/entrypoint#L4


timflannagan1:
/hold

Looks like we may need to wait for #1375 to merge according to the last image build failure.

PR #1382, opened 2020-08-24
hack,Makefile: Explicitly run any of the test/deployframework tests.

### LINKED ISSUE #1379: Framework tests not enough args in call
```
# github.com/kube-reporting/metering-operator/test/deployframework [github.com/kube-reporting/metering-operator/test/deployframework.test]
test/deployframework/framework_test.go:22:34: not enough arguments in call to df.NewDeployerConfig
	have (string, string, string, string, string, "github.com/kube-reporting/metering-operator/pkg/apis/metering/v1".MeteringConfigSpec)
	want (string, string, string, string, string, string, string, string, "github.com/kube-reporting/metering-operator/pkg/apis/metering/v1".MeteringConfigSpec)
```

Looks like we are missing information for `catalogSourceName, catalogSourceNamespace, subscriptionChannel`

affected line (framework_test.go):
```go
cfg, err := df.NewDeployerConfig(testNamespace, testMeteringOpRepo, testMeteringOpTag, testReportingOpRepo, testReportingOpTag, spec)
```

constructor (framework.go):
```go
// NewDeployerConfig handles the process of validating inputs before returning
// an initialized Deploy.Config object, or an error if there is any.
func (df *DeployFramework) NewDeployerConfig(
	namespace,
	meteringOperatorImageRepo,
	meteringOperatorImageTag,
	reportingOperatorImageRepo,
	reportingOperatorImageTag,
	catalogSourceName,
	catalogSourceNamespace,
	subscriptionChannel string,
	spec metering.MeteringConfigSpec,
) (*deploy.Config, error)
```

timflannagan1:
This test was being ignored as hack/e2e.sh was only running any top-level tests that were in the test/e2e directory.

There are probably cleaner ways to do this but they involve re-organizing the reportingframework/deployframework to be in the test/e2e path, or moving the second `go test ...` call to its own make target and letting the `make e2e` target call that make target so we ensure that we're running both of those tests in the openshift CI environment.

Closes #1379

### DIFF
--- a/hack/e2e.sh
+++ b/hack/e2e.sh
@@ -109,6 +109,12 @@ go test \
     ${EXTRA_TEST_FLAGS} \
     2>&1 | tee "$TEST_LOG_FILE_PATH" ; TEST_EXIT_CODE=${PIPESTATUS[0]}
 
+go test \
+    -timeout 30s \
+    -test.v \
+    "./test/deployframework" \
+    2>&1 | tee "$TEST_LOG_FILE_PATH" ; TEST_EXIT_CODE=${PIPESTATUS[0]}
+
 # if go-junit-report is installed, create a junit report also
 if command -v go-junit-report >/dev/null 2>&1; then
     go-junit-report < "$TEST_LOG_FILE_PATH" > "${TEST_JUNIT_REPORT_FILE_PATH}"
--- a/Makefile
+++ b/Makefile
@@ -126,7 +126,7 @@ e2e: $(DEPLOY_METERING_BIN_OUT)
 	hack/e2e.sh
 
 e2e-upgrade: $(DEPLOY_METERING_BIN_OUT)
-	EXTRA_TEST_FLAGS="-run TestMeteringUpgrades" ./hack/e2e.sh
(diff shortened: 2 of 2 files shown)


hack/e2e.sh
```diff
@@ -109,6 +109,12 @@ go test \
     ${EXTRA_TEST_FLAGS} \
     2>&1 | tee "$TEST_LOG_FILE_PATH" ; TEST_EXIT_CODE=${PIPESTATUS[0]}
 
+go test \
```
bentito, line 112: oh I get it, 2nd call to `go test`. But what if we made the one call with something like:
```go test -run='Test*' ./test/...```

timflannagan1: It would fail to run anything successfully as we're providing custom go test flags and there's no logic to consume those flags (besides the e2e package) which results in an error.

bentito: ✅ 
We should do the package dir reorg you mention in the init comment at some point then.

timflannagan1: Yeah completely agree. I'm not a huge fan of this approach, but it at least gets us back up and running until we have some dedicated time to re-organize these testing packages.

timflannagan1:
Closes #1379

PR #1389, opened 2020-08-31
hack,test: Remove the test-run-all-install-tests test flag in favor of t.short

timflannagan1:
This flag was added to help differentiate between the number of MeteringConfig test installations we spun up during a dev PR vs. during a nightly periodic. In the case of the former, we only want to run a subset of all available tests, and those tests were gated using the value of this field. That said, this field is redundant as we already specify a `t.short` value in the hack/e2e.sh script and we can conditionally skip tests using the value of the testing.Short() expression.


timflannagan1:
Manually adding the valid-bz label as this only affects our CI and nothing user-facing.

PR #1402, opened 2020-09-15
test/e2e: Reduce reporting-operator CPU requests and rename metering tests

timflannagan1:
With the migration to building the metering bundle and index images in CI, we needed to switch to a workflow that has smaller default AWS instance sizes than what we're use to running in CI, even when machine autoscaling has been enabled.

In order to schedule the reporting-operator Pods, we can limit the CPU requests which allows for the reporting-operator to fit inside of a recently scaled up machine.

This also renames some of the test/e2e/main_test.go test table names, which are used as a test namespace suffix as those names didn't make much sense in that context.

### DIFF
--- a/test/e2e/main_test.go
+++ b/test/e2e/main_test.go
@@ -161,7 +161,7 @@ func TestManualMeteringInstall(t *testing.T) {
 		MeteringConfigManifestFilename string
 	}{
 		{
-			Name:                      "InvalidHDFS-MissingStorageSpec",
+			Name:                      "HDFS-MissingStorageSpec",
 			MeteringOperatorImageRepo: meteringOperatorImageRepo,
 			MeteringOperatorImageTag:  meteringOperatorImageTag,
 			Skip:                      false,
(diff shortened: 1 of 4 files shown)


timflannagan1:
Ah, must-gather for the ipi-aws workflow flaked, but the actual tests passed so overriding that result.

test/e2e/main_test.go
```diff
@@ -225,7 +225,7 @@ func TestManualMeteringInstall(t *testing.T) {
 			MeteringConfigManifestFilename: "node-selector-prometheus-importer-disabled.yaml",
 		},
 		{
-			Name:                      "ValidHDFS-ReportDynamicInputData",
+			Name:                      "HDFS-ReportDynamicInputData",
```
bentito, line 228: I still don't get what these top level test names are trying to convey? Like the config'ed environment all the subtests will operate in? If so there is more than just Storage. And what about Dynamic/Static what's that telling me? I'm not saying fix this for this PR, we can percolate on it for awhile.

timflannagan1: Yeah, the test names can still be improved. Conceptually, I think we've been interested in conveying what storage type we're running metering on top of and what top-level reporting tests we're going to run once metering is installed and reporting a ready status.

PR #1405, opened 2020-09-22
Bug 1881650: Ensure the volumeName label is not empty in the kube_persistentvolumeclaim_info metric

timflannagan1:
Fix the `persistentvolumeclaim-request-bytes` ReportDataSource promQL query to ensure that the `volumeName` label is not empty. Otherwise, we risk a many-to-many error when joining the two metric series.


timflannagan1:
This is a manual cherry-pick of #1269. For context, we were seeing a decent amount of many-to-many errors in the reporting-operator container logs during the 4.5/4.6 e2e tests and fixed this in master a couple of months ago. Since then, we haven't seen the error pop up again in CI.

markmc:
> This is a manual cherry-pick of #1269. For context, we were seeing a decent amount of many-to-many errors in the reporting-operator container logs during the 4.5/4.6 e2e tests and fixed this in master a couple of months ago. Since then, we haven't seen the error pop up again in CI.

(patch manager) thanks for the context!

timflannagan1:
Hmm, it looks like the ocp-ansible-operator base image may have changed since the last week which is causing the images (and e2e) failures. I'll try and run that down later today.

PR #1413, opened 2020-10-21
Hive conf changes

bentito:
As a test to fix Bug 1798741 we are trying switch Hadoop to the latest on the 3.3 branch and thus we also need a more recent Hive. With the latest 3.3.1 branch Hadoop, we pick up the bug fix for this issue: https://issues.apache.org/jira/browse/HADOOP-13230 which we hope solves the bucket cleanup issue we see in BZ 1798741.

This PR changes `hadoop-config/core-site.xml` and Hive startup flags for log specification for these newer versions of Hadoop and Hive.


timflannagan1:
@bentito Is this still needed? Can we close this out?

bentito:
@timflannagan1 not needed, closing.

PR #1417, opened 2020-10-22
fix order of volumes in yaml 

EmilyM1:
Noticed our reporting operator deployment yaml was inconsistent with k8 best practices. They specify the volumemount preceding the name key.
https://kubernetes.io/docs/concepts/storage/volumes/
Went through and fixed our ordering in this file.
There are two I did not fix, and this is because Tim is trying to merge a PR on the CA certs, after that PR merges I will fix the ordering on the lines for those two certs.

### DIFF
--- a/charts/openshift-metering/templates/reporting-operator/reporting-operator-deployment.yaml
+++ b/charts/openshift-metering/templates/reporting-operator/reporting-operator-deployment.yaml
@@ -354,7 +354,7 @@ spec:
 {{- end }}
         volumeMounts:
         - mountPath: /var/run/configmaps/service-ca-bundle
-          name: reporting-operator-service-ca-bundleF
+          name: reporting-operator-service-ca-bundle
           readOnly: true
 {{- if $operatorValues.spec.config.hive.tls.enabled }}
         - mountPath: /var/run/secrets/hive-tls
--- a/charts/openshift-metering/templates/reporting-operator/reporting-operator-deployment.yaml
+++ b/charts/openshift-metering/templates/reporting-operator/reporting-operator-deployment.yaml
@@ -353,32 +353,32 @@ spec:
 {{ toYaml $operatorValues.spec.livenessProbe | indent 10 }}
 {{- end }}
         volumeMounts:
-        - name: reporting-operator-service-ca-bundle
-          mountPath: /var/run/configmaps/service-ca-bundle
+        - mountPath: /var/run/configmaps/service-ca-bundle
+          name: reporting-operator-service-ca-bundleF
           readOnly: true
 {{- if $operatorValues.spec.config.hive.tls.enabled }}
-        - name: hive-tls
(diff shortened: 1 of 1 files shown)


charts/openshift-metering/templates/reporting-operator/reporting-operator-deployment.yaml
```diff
@@ -353,32 +353,32 @@ spec:
 {{ toYaml $operatorValues.spec.livenessProbe | indent 10 }}
 {{- end }}
         volumeMounts:
-        - name: reporting-operator-service-ca-bundle
-          mountPath: /var/run/configmaps/service-ca-bundle
+        - mountPath: /var/run/configmaps/service-ca-bundle
+          name: reporting-operator-service-ca-bundleF
```
timflannagan1, line 357:
```suggestion
          name: reporting-operator-service-ca-bundle
```

charts/openshift-metering/templates/reporting-operator/reporting-operator-deployment.yaml
```diff
@@ -353,32 +353,32 @@ spec:
 {{ toYaml $operatorValues.spec.livenessProbe | indent 10 }}
 {{- end }}
         volumeMounts:
-        - name: reporting-operator-service-ca-bundle
-          mountPath: /var/run/configmaps/service-ca-bundle
+        - mountPath: /var/run/configmaps/service-ca-bundle
+          name: reporting-operator-service-ca-bundle
```
timflannagan1, line 357: @EmilyM1 Going through the [k8s volume documentation](https://kubernetes.io/docs/concepts/storage/volumes/) I don't see anything explicit in that documentation that would suggest this ordering of volume mounts is considered a best practice, but it instead appears to just be a k8s documentation standard.

EmilyM1: Yeah, the consistency certainly won't hurt anything though. I vote we merge it. It's structured better with order.

timflannagan1: This wouldn't be making it more consistent though as the [majority of charts in the repository](https://github.com/kube-reporting/metering-operator/search?q=mountPath) specify `name` as the first property in any given mount path array index.

EmilyM1: Clean up has to start somewhere. I volunteer to fix the rest up. Let's merge this thing, the ordering is bothering.

timflannagan1: Sure, and that's fine with me, but if we're making the effort to clean up these YAML files, let's do it all in a single PR rather than splitting up related changes into multiple PRs.

timflannagan:
(Clearing out any non-bug PRs)

PR #1421, opened 2020-10-27
hack,test: Gather the OLM registry resources during e2e runs

timflannagan1:
Adds a hack script that's responsible for gathering the CatalogSource and the corresponding PackageManifest custom resources that were used to deploy a custom version of Metering.

Updates the test helpers, adding a function that calls that bash script is responsible for setting the correct environment variables based on the user-provided e2e flags. The main_test.go e2e driver will run this gather registry artifacts test helper function when the e2e suite has been instantiated without the `--run-dev-setup` flag.

### DIFF
--- a/hack/gather-registry-artifacts.sh
+++ b/hack/gather-registry-artifacts.sh
@@ -1,0 +1,13 @@
+#! /bin/bash
+
+set -eou pipefail
+set -x
+
+CATALOG_SOURCE_NAMESPACE=${CATALOG_SOURCE_NAMESPACE:=openshift-marketplace}
+CATALOG_SOURCE_NAME=${CATALOG_SOURCE_NAME:=redhat-operators}
+
+OUTPUT_DIRECTORY=${OUTPUT_DIRECTORY:=$(mktemp -ud)/resources}
+mkdir -p ${OUTPUT_DIRECTORY}
+
+oc -n ${CATALOG_SOURCE_NAMESPACE} get catalogsources ${CATALOG_SOURCE_NAME} -o yaml > ${OUTPUT_DIRECTORY}/catalogsource.yaml
+oc -n ${CATALOG_SOURCE_NAMESPACE} get packagemanifests -l "catalog-namespace=${CATALOG_SOURCE_NAMESPACE},catalog=${CATALOG_SOURCE_NAME}" -o yaml > ${OUTPUT_DIRECTORY}/packagemanifests.yaml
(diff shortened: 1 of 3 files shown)


test/testhelpers/helpers.go
```diff
@@ -289,3 +290,28 @@ func ExecPodCommandWithOptions(config *rest.Config, client kubernetes.Interface,
 
 	return stdoutBuf, stderrBuf, nil
 }
+
+func GatherRegistryResources(outputPath, scriptPath, name, namespace string) error {
+	path, err := exec.LookPath(scriptPath)
```
timflannagan1, line 295: Just for reviewers. this `exec.LookPath` is a pretty nice utility function that I didn't realize existed before re-reading that exec package documentation. It essentially acts as a `os.Stat` operation to ensure the path to the binary or script exists, which is nice. We may want to update the rest of our instances of this package to also use this function.

timflannagan1: Also, the link to that function documentation: https://golang.org/pkg/os/exec/#LookPath.

hack/gather-registry-artifacts.sh
```diff
@@ -0,0 +1,13 @@
+#! /bin/bash
+
+set -eou pipefail
+set -x
+
+CATALOG_SOURCE_NAMESPACE=${CATALOG_SOURCE_NAMESPACE:=openshift-marketplace}
+CATALOG_SOURCE_NAME=${CATALOG_SOURCE_NAME:=redhat-operators}
+
+OUTPUT_DIRECTORY=${OUTPUT_DIRECTORY:=$(mktemp -d)/resources}
```
bentito, line 9:
```suggestion
OUTPUT_DIRECTORY=${OUTPUT_DIRECTORY:=$(mktemp -ud)/resources}
```
`-u` is dry run, avoids creating a directory twice

test/testhelpers/helpers.go
bentito: 👍

test/e2e/main_test.go
```diff
@@ -124,7 +133,26 @@ func testMainWrapper(m *testing.M) int {
 		df.Logger.Fatalf("Failed to create the CatalogSource custom resource using the %s index image: %v", indexImage, err)
 	}
 	if !df.RunDevSetup {
-		defer df.DeleteRegistryResources(registryProvisioned, catalogSourceName, catalogSourceNamespace)
+		var (
+			errArr []string
```
bentito, line 137:
```suggestion
			errors []string
```

bentito: unless that's a keyword, idk reads easier to me. I can look and see the type so having in the name doesn't help me much.

PR #1424, opened 2020-10-30
Migrate to using helm 3.x for templating resources

timflannagan1:
## Overview

Updates the metering-operator repository to use helm 3.x for it's templating needs. We're currently using an out-of-date helm version (e.g. 2.8.x) and we're running into situations where updating dependencies is a pain due to limitations in the `glide` package manager. Ideally, we would move off of helm entirely and use something like Ansible templates or jsonnet, but this is a stop-gap in the meantime.

### Updating the metering-ansible-operator Ansible role

Update the `deploy_resources.yaml` task file, which is responsible for templating helm charts into YAML manifests so we can pass that to the `k8s` Ansible module.

This migration from helm 2.x to helm 3.x is fairly trivial as we only utilized helm for it's templating and nothing deployment (i.e. tiller) facing.

Note on the changed flags:
- The global `--name` flag was removed.
- The `--execute` (`-x`) was renamed to `--show-only` (`-s`) and the functionality remains the same.

### Updating the YAML manifest bash scripts

Update the `hack/generate-metering-manifest.sh` bash script and always specify a `values.yaml` file location.

Before, we supported multi-value file situations, but that makes handling potentially empty templated files more difficult in the child `hack/create-metering-manifests` script.

Updates the `hack/create-metering-manifests.sh` bash script for general changes needed to get helm 3.x working with the existing YAML manifests.

- Deflated the top-level parameter handled to just exit early if there are fewer parameters than we expect from the call site.
- Changes how we handle potentially empty templated files. Now, with the `--show-only` flag, any templated files that are empty results in an error. Before, the expected behavior was that we would still attempt to template, but if stdout was empty, don't redirect that output to a file.

A quick note on the testing that was done locally:
- Spun up metering using a custom-built helm image that contains the helm release-3.0 branch.
- Edited the local MeteringConfig custom resource, replacing the storage configuration, and changed some of the reporting-operator custom parameters.

### DIFF
--- a/manifests/deploy/openshift/olm/bundle/4.7/meteringoperator.v4.7.0.clusterserviceversion.yaml
+++ b/manifests/deploy/openshift/olm/bundle/4.7/meteringoperator.v4.7.0.clusterserviceversion.yaml
@@ -207,7 +207,7 @@ metadata:
     categories: OpenShift Optional, Monitoring
     certified: "false"
     containerImage: quay.io/openshift/origin-metering-ansible-operator:4.7
-    createdAt: 2019-01-01T11:59:59Z
+    createdAt: "2019-01-01T11:59:59Z"
     description: Chargeback and reporting tool to provide accountability for how resources
       are used across a cluster
     operatorframework.io/cluster-monitoring: "true"
@@ -215,7 +215,6 @@ metadata:
     operators.openshift.io/capability: '["fips", "cluster-proxy"]'
     repository: https://github.com/kube-reporting/metering-operator
     support: Red Hat, Inc.
-
 spec:
   displayName: Metering
   description: |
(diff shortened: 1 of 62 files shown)


timflannagan1:
Need to update the helm fork to bump that repository to a more recent release-3.x branch.
/hold

manifests/deploy/openshift/olm/bundle/4.7/meteringoperator.v4.7.0.clusterserviceversion.yaml
```diff
@@ -207,15 +207,14 @@ metadata:
     categories: OpenShift Optional, Monitoring
     certified: "false"
     containerImage: quay.io/openshift/origin-metering-ansible-operator:4.7
-    createdAt: 2019-01-01T11:59:59Z
+    createdAt: "2019-01-01T11:59:59Z"
```
bentito, line 210: I recognize this change as something a newer Helm would allow (force?). Is the removal of trailing new-lines also something newer Helm forced?

timflannagan1: I think that was just the expected behavior of `toYAML` in the helm 2.x world: https://github.com/helm/helm/issues/3470.

That said, not entirely sure if there's any impact to double-quoting the `createdAt` timestamp. I haven't seen anything pop up in the e2e logs that would suggest that OLM cannot consume that field in that format.

PR #1425, opened 2020-10-30
hack,charts: Remove the saas-metering YAML manifests

timflannagan1:
Remove the unused telemeter YAML manifests from the hack scripts responsible for generating the overall list of metering manifests.


EmilyM1 (approved):
/lgtm
/approve

timflannagan1:
Oh right, looks like the helm3 PR already went through. Going to need to force this through in the meantime.

PR #1427, opened 2020-11-02
test/e2e: Add a post-install test that ensures the Metering leader election ConfigMap is present

timflannagan1:
Add a new post-install test that ensures that a ConfigMap is present that is used to determine the Metering Operator leader election.

Update the test/e2e/main_test.go e2e entry-point function, adding this post-install function to the list of tests that will run against the static data input Metering installation.


timflannagan1:
Need to wait for #1424 to merge as the verify check will continue to fail without it.

/hold

PR #1428, opened 2020-11-03
Migrate to using the v1.x ansible-operator

timflannagan1:
### Overview of Changes

Update the metering-ansible-operator to use the ansible-operator v1.x version. There are a few breaking changes between the 0.x and 1.x version of the operator-sdk, but we utilize very few of those CLI utilities which lower the number of changes necessary to migrate to this new major version.

#### Updates to the Dockerfile(s)

With the v1 binary, the ansible-operator now expects a `run` command:

```bash
bash-4.4$ /usr/local/bin/ansible-operator help
Usage:
  ansible-operator [command]

Available Commands:
  help        Help about any command
  run         Run the operator
  version     Prints the version of operator-sdk

Flags:
  -h, --help   help for ansible-operator

Use "ansible-operator [command] --help" for more information about a command.
```

#### Updates to the metering-ansible-operator Ansible Role

The watches.yaml file, which contains the API definition for resources that the Metering Operator is responsible for watching and ensure that the `meta` Ansible variable maps to `ansible_operator_meta`.

Before the `meta` variable was exposed through the ansible-operator and you would be able to pull information about the `MeteringConfig` custom resource that it's watching via `{{ meta.namespace }}` or `{{ meta.name }}`. Now, with the ansible-operator v1 version, that variable has been removed in favor of `ansible_operator_meta`. Explicitly mapping that variable to the `meta` Ansible variable ensures the same behavior we expect in previous ansible-operator versions.

#### Updates to the YAML manifests

Update the metering-ansible-operator Deployment and CSV helm charts to work with the ansible-operator v1 version.

A couple of things to note:
- The new default metrics port is port 8080 instead of port 8383 like in previous ansible-operator releases. Explicitly setting the `--metrics-addr=:8383` flag for the ansible-operator binary ensures the expected behavior.
- The `OPERATOR_NAME` environment variable has been deprecated which was used for leader election purposes. Now, we need to explicitly set the `--leader-election-id` flag to the `metering-operator` string.
- The `POD_NAME` environment variable is no longer needed. Before, that variable was exposed through the k8s downward API and used to create a leader election lease ConfigMap where the `$POD_NAME` was listed as the owner reference. In the ansible-operator v1 architecture, the controller runtime leader election implementation is used.
- The `--zap-level` flag has been renamed to `--zap-log-level`.

### DIFF
--- a/Dockerfile.metering-ansible-operator
+++ b/Dockerfile.metering-ansible-operator
@@ -3,7 +3,7 @@ FROM quay.io/openshift/origin-metering-helm:latest as helm
 # final image needs kubectl, so we copy `oc` from cli image to use as kubectl.
 FROM quay.io/openshift/origin-cli:latest as cli
 # the base image is the ansible-operator's origin images
-FROM quay.io/operator-framework/ansible-operator:v0.19.4
+FROM quay.io/operator-framework/ansible-operator:v1.0.1
 
 USER root
 RUN set -x; INSTALL_PKGS="curl bash ca-certificates less which openssl" \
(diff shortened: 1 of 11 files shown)


timflannagan1:
It looks like 1.x hasn't been ported downstream yet so I imagine the images check will fail.

/hold

PR #1430, opened 2020-11-03
Add support for mirroring changes to the Metering bundle manifests

timflannagan1:
Changes:
- Update the build_root container image that is used to run checks in CI. The `opm` binary is used to keep the OLM manifests stored in the package manifest format in line with the new bundle packaging format.
- Update the Makefile to expose the current release of Metering. This so far is only used in one place, but introducing this variable can be propagated to the other various hack scripts used throughout metering.
- Update the hack/generate-metering-manifests.sh bash script to mirror any changes made to the `manifests/deploy/openshift/olm/bundle/<version>` to the `olm_deploy/bundle` artifacts. The former is used downstream by ART to build the operator and the latter is used by CI to generate a bundle (and an index image containing that bundle image) for the operator that gets deployed at runtime.

Manifests stored in the olm_deploy/bundle directory were regenerated by running the following command:

```bash
$ make metering-manifests
```


timflannagan1:
Ah right, I forgot making changes to the build_root Dockerfile needs to be merged in master before you can get those images working. I'll move that change outside of this PR and force-merge that.

PR #1435, opened 2020-11-05
Remove the olm_deploy directory and migrate the Metering bundle to the bundle directory

timflannagan1:
More housekeeping/general cleanup based on top of #1434.

The olm_deploy directory previously-stored a mixture of resources that were needed to stand up the most up-to-date Metering installation using OLM in CI, but with the introduction of the Openshift CI being able to create and expose bundle and index images for us, we have no need for workarounds that were previously needed.

As a result, the following changes have been made:
- The cluster and machine autoscaling YAML manifests were moved to test/e2e/manifests.
- The default directory that contains the Metering bundle will now be the `bundle` directory in the root repository.
- The `olm_deploy` directory has been removed.


timflannagan1:
Need to wait for 1434 to be merged first. 

/hold

PR #1439, opened 2020-11-13
charts,manifests: Fix the invalid metering image tag in the upstream CSV manifest

timflannagan1:
There's no quay.io/coreos/metering-ansible-operator:4.7 image tag, but there is a release-4.7 one that got overridden by the PR that bumped all of the image tags to 4.7.


timflannagan1:
Going to push this through as there's nothing in this PR that would cause e2e failures.

PR #1440, opened 2020-11-16
test/e2e: Avoid using the fatal logrus level in the main_test.go

timflannagan1:
The logrus.Fatal logging level logs an entry at the fatal level and then calls os.Exit(1) after logging. This is problematic in the case of main_test.go where we expect that a custom defer function will be called if something problematic happens after creating the registry resources. If we instead log that error to the error log level and then return 1, we can continue getting the same behavior, but the defer function will still be ran.

### DIFF
--- a/test/e2e/main_test.go
+++ b/test/e2e/main_test.go
@@ -130,7 +130,8 @@ func testMainWrapper(m *testing.M) int {
 	var registryProvisioned bool
 	catalogSourceName, catalogSourceNamespace, err = df.CreateCatalogSourceFromIndex(indexImage)
 	if err != nil {
-		df.Logger.Fatalf("Failed to create the CatalogSource custom resource using the %s index image: %v", indexImage, err)
+		df.Logger.Errorf("Failed to create the CatalogSource custom resource using the %s index image: %v", indexImage, err)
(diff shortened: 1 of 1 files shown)


test/e2e/main_test.go
```diff
@@ -130,7 +130,8 @@ func testMainWrapper(m *testing.M) int {
 	var registryProvisioned bool
 	catalogSourceName, catalogSourceNamespace, err = df.CreateCatalogSourceFromIndex(indexImage)
 	if err != nil {
-		df.Logger.Fatalf("Failed to create the CatalogSource custom resource using the %s index image: %v", indexImage, err)
+		df.Logger.Errorf("Failed to create the CatalogSource custom resource using the %s index image: %v", indexImage, err)
+		return 1
```
timflannagan1, line 134: We could add a helper function that logs a message to the error level and returns 1 every time we call it, but that might be getting too abstract for now.

Also, `logrus.Panicf` could also be utilized as panicking still allows [defer functions to be run](https://blog.golang.org/defer-panic-and-recover) but the stack trace is a bit difficult to parse in the e2e_test.log logs.

PR #1441, opened 2020-11-16
Avoid using docker.io images in the build_root image

timflannagan1:
We're starting to get rate limited when using the openshift/origin-golang-1.15 image in the build_root test container image.

We can instead use the registry.svc.ci.openshift.io registry as the tests will already have access to that registry in the CI environment.


timflannagan1:
Force-merging as the changes made to this Dockerfile won't propagate until the next set of CI tests.

PR #1447, opened 2020-11-20
Dockerfile,Makefile: Add a make target that verifies the metering bundle contents are valid

timflannagan1:
Update the Dockerfile.src (build root Dockerfile for CI) and add the `operator-sdk binary` to the container image.

Update the Makefile and expose a `verify-bundle` target, which is responsible for running the `operator-sdk bundle validate ./bundle` command to validate the Metering bundle metadata/manifest contents.

Note: this validation of manifests for OLM consumption was previously covered by the `operator-courier` utility, but that utility doesn't support validating apiextensions.k8s.io/v1 CRD types so that check was removed.


timflannagan1:
This needs to be force-merged as any changes made to the Dockerfile.src won't propagate until it's been merged to the master branch.

Issue #1444, opened 2020-11-18
Fails on FIPS systems - bundled crypto module missing fix from Ansible 2.9


jeffcpullen:
The operator fails to deploy on FIPS enabled systems with the following error

```
Traceback (most recent call last):
  File "/tmp/.ansible-/tmp/ansible-tmp-1605639519.1729925-24168-234385709602156/AnsiballZ_openssl_privatekey.py", line 102, in <module>
    _ansiballz_main()
  File "/tmp/.ansible-/tmp/ansible-tmp-1605639519.1729925-24168-234385709602156/AnsiballZ_openssl_privatekey.py", line 94, in _ansiballz_main
    invoke_module(zipped_mod, temp_path, ANSIBALLZ_PARAMS)
  File "/tmp/.ansible-/tmp/ansible-tmp-1605639519.1729925-24168-234385709602156/AnsiballZ_openssl_privatekey.py", line 40, in invoke_module
    runpy.run_module(mod_name='ansible.modules.crypto.openssl_privatekey', init_globals=None, run_name='__main__', alter_sys=True)
  File "/usr/lib64/python3.6/runpy.py", line 205, in run_module
    return _run_module_code(code, init_globals, run_name, mod_spec)
  File "/usr/lib64/python3.6/runpy.py", line 96, in _run_module_code
    mod_name, mod_spec, pkg_name, script_name)
  File "/usr/lib64/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/tmp/ansible_openssl_privatekey_payload_rg2zb9st/ansible_openssl_privatekey_payload.zip/ansible/modules/crypto/openssl_privatekey.py", line 692, in <module>
  File "/tmp/ansible_openssl_privatekey_payload_rg2zb9st/ansible_openssl_privatekey_payload.zip/ansible/modules/crypto/openssl_privatekey.py", line 676, in main
  File "/tmp/ansible_openssl_privatekey_payload_rg2zb9st/ansible_openssl_privatekey_payload.zip/ansible/modules/crypto/openssl_privatekey.py", line 303, in generate
  File "/tmp/ansible_openssl_privatekey_payload_rg2zb9st/ansible_openssl_privatekey_payload.zip/ansible/modules/crypto/openssl_privatekey.py", line 545, in _get_fingerprint
[...]

PR #1450, opened 2020-11-20
Bug 1900125: Stop vendoring the module_utils/crypto.py Ansible library

timflannagan1:
Remove the module_utils (and as a result, the` module_utils.crypto.py` library) directory in the meteringconfig Ansible Role.

Previously, this was necessary in order to have Metering function on FIPS-enabled clusters as the crypto.py module was being referenced in the openssl-* Ansible modules, which we namely use for generating certificates for the components in the stack. This library listed the md5 algorithm as FIPS-compliant, despite md5 not being FIPS-compliant in many FIPS standards, which resulted in our Ansible role failing when attempting to call any of the openssl-* modules.

Now, this behavior has been patched upstream in Ansible (in versions greater than 2.9.6) so vendoring this library is no longer needed.


timflannagan1:
I still need to test this on a FIPS-enabled cluster to verify whether this fixes the bug that's attached to this PR, in addition to whether that Ansible versioning pinning makes sense in the Dockerfile.metering-ansible-operator.

/hold

timflannagan1:
Fired up a 4.6 FIPS-enabled cluster and verified these changes worked with a custom metering-ansible-operator image.

/hold cancel

timflannagan1:
Fixes #1444

EmilyM1 (approved):
/lgtm
/approve

PR #1452, opened 2020-11-27
pkg/operator: Minor tech debt improvements to the Reporting Operator packages

timflannagan:
List of Changes
- Update `pkg/operator/storagelocations.go`, deflating the updating logic that is stored in the op.handleStorageLocation method, into its own method.This way, we can handle the validation and update logic for any particular StorageLocation the reporting-operator is currently processing into a dedicated method, and reduce the responsibilities of the `op.handleStorageLocation` driver method to making any of the necessary API calls to the Kubernetes client.
- Update any of the comments for functions/methods that appear to have been refactored in the past, but the comment wasn't also updated.
- Update the ordering of parameters for any function/method that passes a `context.Context` to be the first parameter. This makes the Reporting Operator codebase a bit more consistent and aligns with the recommendation from Go that `context.Context` parameters should be first.
- Deflate (refactor) any unnecessary branching logic. Most of the if/else if/else branching structure can be deflated to just an if conditional and that improves readability quite a bit.

### DIFF
--- a/pkg/operator/promsum.go
+++ b/pkg/operator/promsum.go
@@ -237,7 +237,7 @@ func (op *defaultReportingOperator) importPrometheusForTimeRange(ctx context.Con
 				promConn = op.promConn
 			}
 
-			importResults, err := prestostore.ImportFromTimeRange(dataSourceLogger, op.clock, promConn, op.prometheusMetricsRepo, metricsCollectors, ctx, start, end, importCfg)
+			importResults, err := prestostore.ImportFromTimeRange(ctx, dataSourceLogger, op.clock, promConn, op.prometheusMetricsRepo, metricsCollectors, start, end, importCfg)
 			if err != nil {
 				return fmt.Errorf("error importing Prometheus data for ReportDataSource %s: %v", reportDataSource.Name, err)
 			}
@@ -274,7 +274,11 @@ func (op *defaultReportingOperator) getQueryIntervalForReportDataSource(reportDa
 	return queryInterval
 }
 
-func (op *defaultReportingOperator) newPromImporterCfg(reportDataSource *metering.ReportDataSource, query string, prestoTable *metering.PrestoTable) (prestostore.Config, error) {
+func (op *defaultReportingOperator) newPromImporterCfg(
+	reportDataSource *metering.ReportDataSource,
+	query string,
+	prestoTable *metering.PrestoTable,
+) (prestostore.Config, error) {
 	chunkSize := op.cfg.PrometheusQueryConfig.ChunkSize.Duration
 	stepSize := op.cfg.PrometheusQueryConfig.StepSize.Duration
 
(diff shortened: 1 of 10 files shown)


pkg/operator/storagelocations.go
```diff
@@ -70,6 +70,53 @@ func (op *defaultReportingOperator) syncStorageLocation(logger log.FieldLogger,
...
+		if storageLocation.Status.Hive.DatabaseName != storageLocation.Spec.Hive.DatabaseName {
+			storageLocation.Status.Hive.DatabaseName = storageLocation.Spec.Hive.DatabaseName
+			return true, nil
+		}
+		if storageLocation.Status.Hive.Location != storageLocation.Spec.Hive.Location {
+			storageLocation.Status.Hive.Location = storageLocation.Spec.Hive.Location
+			return true, nil
+		}
+	}
+	if storageLocation.Status.Hive.DatabaseName == "" {
```
bentito, line 101: I think this can go. It's already checked above, right?

pkg/operator/storagelocations.go
```diff
@@ -70,6 +70,53 @@ func (op *defaultReportingOperator) syncStorageLocation(logger log.FieldLogger,
...
+		return false, fmt.Errorf("only Hive storage is currently supported")
+	}
+	if storageLocation.Spec.Hive.DatabaseName == "" {
+		return false, fmt.Errorf("spec.hive.databaseName is required if spec.hive is set")
+	}
+	if !reportingutil.IsValidSQLIdentifier(storageLocation.Spec.Hive.DatabaseName) {
+		return false, fmt.Errorf("spec.hive.databaseName %s is invalid, must contain only alpha numeric values, underscores, and start with a letter or underscore", storageLocation.Spec.Hive.DatabaseName)
+	}
+
+	if storageLocation.Spec.Hive.UnmanagedDatabase {
```
bentito, line 89: You could fully unwind the conditional wrapping if you add this to quick return false checks added above.

pkg/operator/promsum.go
```diff
@@ -322,23 +326,34 @@ func (op *defaultReportingOperator) newPromImporterCfg(reportDataSource *meterin
...
-		promConn = op.promConn
 	}
 
+	metricsCollectors := op.newPromImporterMetricsCollectors(reportDataSource, prestoTable, cfg)
 	return prestostore.NewPrometheusImporter(logger, promConn, op.prometheusMetricsRepo, op.clock, cfg, metricsCollectors), nil
 }
 
-func (op *defaultReportingOperator) newPromImporterMetricsCollectors(reportDataSource *metering.ReportDataSource, prestoTable *metering.PrestoTable, cfg prestostore.Config) prestostore.ImporterMetricsCollectors {
+func (op *defaultReportingOperator) newPromImporterMetricsCollectors(
+	reportDataSource *metering.ReportDataSource,
```
bentito, line 353: 👍

pkg/operator/storagelocations.go
timflannagan: That was checking the `.Spec.Hive.DatabaseName`.

PR #1455, opened 2020-11-27
Dockerfile.src: Avoid using the latest Helm image tag

timflannagan1:
Update the Dockerfile.src (build root Dockerfile for CI tests) and use a release-versioned Helm image tag instead of the `latest` tag.

This is needed as the Helm version has recently been bumped to use the 3.x versioning, which comes with some slight breaking changes to the templating logic.

Note: this commit needs to be force-merged in order to have these changes propagate to CI tests.


timflannagan1:
We need to force merge this PR in order to get the verification check working again in the release-4.5 branch for backporting bug fixes. This is a low risk as this only affects the build root image used for CI, and nothing user-facing.

PR #1459, opened 2020-12-07
[release-4.6] Bug 1899136: Stop vendoring the module_utils/crypto.py Ansible library

timflannagan1:
Manual cherry-pick of #1450

Remove the module_utils (and as a result, the module_utils.crypto.py library) directory in the meteringconfig Ansible Role.

Previously, this was necessary in order to have Metering function on FIPS-enabled clusters as the crypto.py module was being referenced in the openssl-* Ansible modules, which we namely use for generating certificates for the components in the stack. This library listed the md5 algorithm as FIPS-compliant, despite md5 not being FIPS-compliant in many FIPS standards, which resulted in our Ansible role failing (and therefore installations and upgrades to 4.6/4.7 will fail) when attempting to call any of the openssl-* modules. 

Now, this behavior has been patched upstream in Ansible (in versions greater than 2.9.6) so vendoring this library is no longer needed.


timflannagan1:
> @timflannagan1: you cannot LGTM your own PR.

Oh right this was a manual cherrypick :facepalm:

trevor-viljoen:
Could we get some 👀 on this to get it moving? It looks like everything is good to go, except for a `cherry-pick-approved label`.

timflannagan1:
@trevor-viljoen It looks like this didn't make the pre-holiday z-stream release. We have dedicated people that act as "patch managers" that are responsible for marking the `cherry-pick-approved` label on PRs that are backported to already released versions of OCP components. Because the BZ attached to this backported PR is `low`, it likely doesn't get attention over other bugs from other components. I'd recommend bumping the priority so this gets into the next release.

timflannagan1:
@trevor-viljoen Actually, I'll bump the priority now and get in touch with the next patch manager as I've been getting pinged quite a bit about this BZ merging for a couple of weeks now.

timflannagan1:
@trevor-viljoen Just a quick follow-up - it looks like this likely isn't getting tagged until next week due to a large 4.6 backlog + limited QE capacity.

timflannagan1:
One of the e2e tests failed to be installed due to an OLM bug.

/retest

PR #1462, opened 2020-12-27
Enhance docu for Basic Auth with htpasswd.data

haf-tech:
Enhanced the docu to clarify the configuration to use Basic Auth and htpasswd, especially with the right parameter combination (createSecret and htpasswd.data)

imo, we should also update here the OpenShift docu, with such details. But I do not know where/how.
https://docs.openshift.com/container-platform/4.5/metering/configuring_metering/metering-configure-reporting-operator.html#metering-basic-authentication_metering-configure-reporting-operator


timflannagan1:
Any updates to the Openshift documentation need to live in the openshift/openshift-docs repository. I can make a note to follow-up with this again once people are back in the new year.

PR #1463, opened 2021-01-12
Bug 1838802: Fix compatibility issues when using MySQL 8.0+ as the underlying Hive metastore database

timflannagan1:
# Overview

Update the Hive metastore deployment to fix compatibility issues
when deploying metering using a non-mysql-5.7 version.

## Changes

Update the `configuring-hive-metastore.md`  documentation and 
remove the reference to the BZ mentioning metering integration 
issues with any non-mysql-5.7 versions.

Update the hive-scripts ConfigMap entrypoint.sh script, which gets
called in the Hive metastore and server StatefulSets and is responsible
for starting those containers.

In the case of derby (the default) databases, using this tool isn't
necessary and will cause issues when running it against an already
exists schema. This tool is needed in order to get metering working
again with MySQL and PostgreSQL versions.

Update the Hive metastore and server StatefulSets and expose an
additional environment variable, `$HIVE_UNDERLYING_DATABASE` that
defaults to `derby`. This environment variable will be used in the
hive-scripts ConfigMap's entrypoint.sh bash script to run schematool
against the correct `-dbType ${HIVE_UNDERLYING_DATABASE`.

Update the Ansible role's `configure_hive_metastore.yaml` task file to
override the `_meteringconfig_hive_metastore_db_url` helper variable.
That helper variable, which is set to the `hive.spec.config.db.url`
configuration and will be overriden with any value that we set in that
task file before the MeteringConfig values have been finalized.

When we're handling postgres resources, we need to remap what we pull
from the JDBC URL as `postgresql` isn't a valid `schematool -dbType`
option, but `postgres` is so use the value from the split URL array
unless we're configuring a postgres database for Hive metastore.

## Testing Notes

The mysql database instance was created using `oc new-app` 
using the following command:

```bash
oc create namespace tflannag
oc -n tflannag new-app --image-stream mysql:8.0 \
    -e MYSQL_USER=testuser=tflannag \
    -e MYSQL_PASSWORD=testpass \
    -e MYSQL_DATABASE=metastore \
    -l db=mysql
```

Once that deployment has been created, the following MeteringConfig
custom resource was used:

```yaml
kind: MeteringConfig
metadata:
  name: "operator-metering"
  annotations:
    "ansible.operator-sdk/verbosity": "1"
spec:
  unsupportedFeatures:
    enableHDFS: true

  storage:
    type: "hive"
    hive:
      type: "hdfs"
      hdfs:
        namenode: "hdfs-namenode-0.hdfs-namenode:9820"

  hive:
    spec:
      config:
        db:
          driver: com.mysql.jdbc.Driver
          username: tflannag
          password: testpass
          url: jdbc:mysql://mysql.tflannag.svc.cluster.local:3306/metastore
```

The changes in this PR were tested using a custom 
metering-ansible-operator image that contains the
requisite helm chart and Ansible fixes and deployed
manually using the deploy-metering CLI:

```bash
./bin/deploy-metering install --namespace tflannag --meteringconfig mc.yaml --repo <custom quay repo> --tag <custom quay tag>
```

### DIFF
--- a/charts/openshift-metering/templates/hive/hive-scripts-configmap.yaml
+++ b/charts/openshift-metering/templates/hive/hive-scripts-configmap.yaml
@@ -109,4 +109,16 @@ data:
     export HIVE_METASTORE_HADOOP_OPTS=" -Dhive.log.level=${HIVE_LOGLEVEL} "
     export HIVE_OPTS="${HIVE_OPTS} --hiveconf hive.root.logger=${HIVE_LOGLEVEL},console "
 
+    SCHEMATOOL_BIN="${SCHEMATOOL_BIN:=${HIVE_HOME}/bin/schematool}"
+    # In the case of non-derby underlying databases been configured
+    # for Hive metastore, attempt to determine if there's already
+    # an initialized schema running the `schematool` CLI with the
+    # `-info` option. If that command fails, then we can make the
+    # reasonable assumption that schema doesn't already exist,
+    # and we can initialize it ourselves before starting Hive.
+    echo "Deploying Hive using ${HIVE_UNDERLYING_DATABASE} as the underlying database"
+    if [[ $HIVE_UNDERLYING_DATABASE != "derby" ]]; then
+      "${SCHEMATOOL_BIN}" -dbType "${HIVE_UNDERLYING_DATABASE}" -info || "${SCHEMATOOL_BIN}" -dbType "${HIVE_UNDERLYING_DATABASE}" -initSchema
+    fi
+
     exec $@
--- a/images/metering-ansible-operator/roles/meteringconfig/tasks/configure_hive_metastore.yml
+++ b/images/metering-ansible-operator/roles/meteringconfig/tasks/configure_hive_metastore.yml
@@ -20,6 +20,28 @@
         message: "Configuring hive metastore"
         lastTransitionTime: "{{ now(utc=False).isoformat() + 'Z' }}"
 
+  - name: Determine the Hive underlying database
+    block:
+    - name: Split the hive.spec.config.db.url into an array
+      set_fact:
+        _url_array: "{{ _meteringconfig_hive_metastore_db_url.split(':') }}"
+
+    - name: Verify the JDBC URL is valid before parsing
+      assert:
+        that:
+        - '{{ _url_array[0] == "jdbc" }}'
+        - '{{ _url_array[1] in ["mysql", "postgresql", "derby"] }}'
+        msg: "Invalid JDBC URL {{ _meteringconfig_hive_metastore_db_url }} passed in the MeteringConfig.Spec"
+
+    # Note: the `schematool -dbType` option expects either `mysql`,
+    # `derby`, `oracle` and `postgres` as valid options. In the case
+    # of the JDBC URL, only `postgresql` is valid, so we need to do
+    # some re-mapping of naming if that is what has been specified
+    # in the hive.spec.config.db.url field in the MeteringConfig CR.
+    - name: Override the default hive metastore database type
+      set_fact:
+        _hive_metastore_db_type: "{{ 'postgres' if _url_array[1] == 'postgresql' else _url_array[1] }}"
+
   - name: Override the default hive.spec.metastore.storage.create option to false
     set_fact:
       _hive_metastore_create_default_storage: false
@@ -69,6 +91,7 @@
     _meteringconfig_hive_metastore_db_secretName: "{{ meteringconfig_spec_overrides | json_query('hive.spec.config.db.secretName') }}"
     _meteringconfig_hive_metastore_db_username: "{{ meteringconfig_spec_overrides | json_query('hive.spec.config.db.username') }}"
     _meteringconfig_hive_metastore_db_password: "{{ meteringconfig_spec_overrides | json_query('hive.spec.config.db.password') }}"
+    _meteringconfig_hive_metastore_db_url: "{{ meteringconfig_spec_overrides | json_query('hive.spec.config.db.url') }}"
     _meteringconfig_hive_metastore_db_configuration: "{{ meteringconfig_spec_overrides | json_query('hive.spec.config.db') }}"
   no_log: false
   rescue:
--- a/Documentation/configuring-hive-metastore.md
+++ b/Documentation/configuring-hive-metastore.md
@@ -52,8 +52,6 @@ kubectl -n $METERING_NAMESPACE create secret generic <name of the secret> --from
 
 ### Using MySQL for the Hive Metastore database
 
-**Note**: Metering cannot work with more recent versions of MySQL, which is being tracking in [BZ #1838802](https://bugzilla.redhat.com/show_bug.cgi?id=1838802). Instead, use the 5.7 version which has been tested.
-
 ```yaml
 spec:
   hive:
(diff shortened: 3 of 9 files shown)


charts/openshift-metering/templates/hive/hive-scripts-configmap.yaml
```diff
@@ -109,4 +109,32 @@ data:
...
+    faq -f xml -o json \
+        '.configuration.property[] | select (.name=="javax.jdo.option.ConnectionURL").value' \
+        "${HIVE_HOME}/conf/hive-site.xml" | \
+        awk -F [/:] '{ print $2 }'
+
+    # Check if there's existing metastore schema for that ${DATABASE_TYPE} and if no
+    # schema exists, i.e. no errors was returned when using the `-info` parameter, attempt
+    # to initialize that schema.
+    echo "Deploying Hive using ${DATABASE_TYPE} as the underlying database"
+    "${SCHEMATOOL_BIN}" -dbType ${DATABASE_TYPE} -info || "${SCHEMATOOL_BIN}" -dbType ${DATABASE_TYPE} -initSchema
```
timflannagan1, line 138: Just a quick comment on this:
- Likely want to move this logic to Ansible as that has URL parsing filters available to us.
- Likely need to expose that `$DATABASE_TYPE` as an environment variable in the StatefuleSet helm chart. Ansible could then set this after parsing the URL in a task.
- In order for this to work with mysql 8.0, we need to update the mysql java connector version in the Hive image.
- Need to use the `schematool -dbType mysql -initSchema` when working with a newer mysql jdbc driver version (one that supports both 5.6/5.7 and 8.0) as there's some compatibility errors that populate when attempting to create the initial metastore tables. I haven't tested whether this is the case for postgres (or derby) - we'll see in the next CI baring any obvious mistakes in the Go code.
- When running `schematool -dbType mysql -initSchema` and that metastore schema already exists in mysql, an error will be produced, which is why we first check the output using the `-info` parameter and if that doesn't fail, then we attempt to initialize the schema.

timflannagan1: PR for updating the mysql connector java version: https://github.com/kube-reporting/hive/pull/58

Right now, we don't need to wait for that PR to merge, but it also allows us to run tests against mysql 8.0 (and clear out three-related BZs).

charts/openshift-metering/templates/hive/hive-scripts-configmap.yaml
```diff
@@ -109,4 +109,35 @@ data:
...
+
+    # Check if there's existing metastore schema for that ${DATABASE_TYPE} and if no
+    # schema exists, i.e. no errors was returned when using the `-info` parameter, attempt
+    # to initialize that schema. Note: running schematool when using derby, which is what
+    # metering deploys by default, can be problematic during upgrades so only run it when
+    # mysql is the underlying database.
+    echo "Deploying Hive using ${DATABASE_TYPE} as the underlying database"
+    if [[ $DATABASE_TYPE == "mysql" ]]; then
+        "${SCHEMATOOL_BIN}" -dbType ${DATABASE_TYPE} -info || "${SCHEMATOOL_BIN}" -dbType ${DATABASE_TYPE} -initSchema
+    fi
```
timflannagan1, line 141: I was seeing failures in the post-upgrade CI job that deploys derby in a previously released version, or metering then upgrades to using an image that contains these hive-scripts ConfigMap fixes. We haven't seen any problems with derby in the past, and it looks like only mysql (haven't tested postgres yet) require running the `schematool` CLI to get Hive and MySQL playing nicely.

test/e2e/manifests/mysql/deployment.yaml
```diff
@@ -0,0 +1,50 @@
+apiVersion: v1
```
timflannagan1, line 1: Note: this manifest likely won't be needed once the Hive PR, which bumps the mysql connector java version, lands and we can instead use the `mysql:8.0` imagestream that gets shipped by default in OCP.

charts/openshift-metering/templates/hive/hive-scripts-configmap.yaml
```diff
@@ -109,4 +109,35 @@ data:
...
+    # to avoid determing what database needs to be passed to `schematool -dbType`.
+    #
+    # TODO(tflannag): we may want to use a more dedicated tool in order to parse a URL
+    # as there's currently no validation in place such that invalid ConnectionURL values
+    # being configured may result in unexpected values being returned. Going forward, we
+    # should likely use Ansible to parse this URL for us and setting that environment variable
+    # to avoid determining it using bash.
+    export SCHEMATOOL_BIN="${HIVE_HOME}/bin/schematool"
+    DATABASE_TYPE=$(faq -f xml -o json \
+        '.configuration.property[] | select (.name=="javax.jdo.option.ConnectionURL").value' \
```
bentito, line 129: maybe this could just be `ConnectionURL` as then the class pathing info can change without it breaking. I think this class name is pretty locked in, so unlikely to change.

bentito:
Just the comment, otherwise lgtm

images/metering-ansible-operator/roles/meteringconfig/tasks/configure_hive_metastore.yml
```diff
@@ -20,6 +20,31 @@
...
 
+  - name: Determine the Hive underlying database
+    block:
+    - name: Dump out the URL for debugging
+      debug:
+        var: _meteringconfig_hive_metastore_db_url
+
+    - name: Split the hive.spec.config.db.url into an array
+      set_fact:
+        _url_array: "{{ _meteringconfig_hive_metastore_db_url.split(':') }}"
```
timflannagan1, line 31: The [`urlsplit` Ansible filter](https://docs.ansible.com/ansible/latest/user_guide/playbooks_filters.html#splitting-urls) wasn't able to parse JDBC URLs as I originally expected. The `scheme` comes back as `jdbc` instead of something like `postgresql` and everything past the `jdbc://...` was put in a `path` variable, which doesn't move the needle any more than using bash to do this filter for us.

In the end, I settled on using the `split` method and splitting that URL string by the `:` delimiter and validating those inputs so we have a somewhat consistent representation of what a JDBC URL should look like.

charts/openshift-metering/templates/hive/hive-scripts-configmap.yaml
timflannagan1: Ended up going in a different direction by parsing the `spec.hive.spec.config.db.url` field in Ansible instead of attempting to parse an xml file using `faq` and hoping for the best, but agreed, that sounds reasonable to me.

test/e2e/manifests/meteringconfigs/mysql.yaml
```diff
@@ -42,6 +42,9 @@ spec:
 
   hive:
     spec:
+      image:
+        repository: quay.io/tflannag/hive
+        tag: db
```
timflannagan1, line 47: This image contains a slightly more up-to-date postgres-jdbc jar (42.2.3 vs. 4.2.14) and adds/removes the mysql-connector-java jar in favor of the mariadb one we started using in the 4.6 images.

timflannagan1:
Removed the `SCHEMATOOL_BIN` environment variable definition while cleaning up the hive-scripts entrypoint script. Re-added that back.

```bash
Adding /var/run/configmaps/service-ca-bundle/service-ca.crt to /etc/alternatives/jre/lib/security/cacerts
Certificate was added to keystore
Adding user hadoop with current UID 1000650000 to /etc/passwd
Deploying Hive using mysql as the underlying database
/hive-scripts/entrypoint.sh: line 102: : command not found
/hive-scripts/entrypoint.sh: line 102: : command not found
```

PR #1471, opened 2021-01-21
Bug 1841883: Fix the namespace-persistentvolumeclaim-usage ReportQuery

timflannagan1:
# Overview

Update the PVC-related ReportQuery YAML manifests to fix an issue where namespace PVC usage was incorrectly calculated and resulted in report results that were orders of magnitude larger than the actual usage.

#### Update the join method in the `persistentvolumeclaim-usage-with-phase-raw` ReportQuery

Update the charts/openshift-metering `persistentvolumeclaim-usage-with-phase-raw` ReportQuery YAML manifest definition and replace the `FULL OUTER JOIN ...` with `LEFT OUTER JOIN ...` when joining between the raw usage and the raw phase tables.

This is needed in order to avoid `NULL` columns being created due to row mismatches between those two tables.

We can see `NULL` columns being produced when interacting with that table locally:

For more information, reference the details in the commit that changes that joining technique.

#### Update the namespace-persistentvolumeclaim-usage SQL Query

Update SQL query defined in the namespace-persistentvolumeclaim-usage `ReportQuery` YAML manifest.

Previously, this `ReportQuery` would produce report results that are orders-of-magnitude larger than what's been actually used in the PVC.

By first averaging out PVC usage for an individual PVC in a namespace according to a date/timestamp range, we can avoid summing each metric snapshot of PVC usage for that same range. This change results in inaccurate readings of total PVC usage whereas the previous query resulted in MiBs of PVC usage in a namespace being reported as GiBs.

## Reviewer Notes

- This needs to be backported to at least 4.6.
- Most of the commits detail the testing YAML manifests used and more in-depth comments than the PR description.


timflannagan1:
I need to clean up that PR description before merging - I just copied and pasted the commit message results.

/hold

charts/openshift-metering/templates/openshift-reporting/report-queries/persistentvolumeclaim-usage.yaml
```diff
@@ -196,19 +196,32 @@ spec:
...
     type: ReportDataSource
     default: persistentvolumeclaim-usage-with-phase-raw
   query: |
+    WITH averaged_persistentvolumeclaim_usage_per_pvc AS (
+      SELECT
+        timestamp '{| default .Report.ReportingStart .Report.Inputs.ReportingStart| prestoTimestamp |}' AS period_start,
+        timestamp '{| default .Report.ReportingEnd .Report.Inputs.ReportingEnd | prestoTimestamp |}' AS period_end,
+        namespace,
+        persistentvolumeclaim,
+        avg(persistentvolumeclaim_usage_bytes) as avg_persistentvolumeclaim_usage_bytes
```
timflannagan1, line 212: Averaging out these usage values per PVC in a particular period is definitely a step in the right direction, especially given the reporting for this query is _way_ off, but I wonder if there's a better way to calculate usage over time using a more suitable aggregate function than `avg`.

That said, after testing this locally in various scenarios, this still holds up nicely and it's close to the actual PVC disk usage, but skews slightly behind in the case that usage is drastically increased in a short period of time ( e.g. large files are copied into the running container) but this may be the closest we can get without something more complex.

bentito: I can't think of a more appropriate aggregate function than AVG. Yeah it will skew off for large data storage increases over short time, but the alternative is probably no longer something that can be readily accomplished with SQL and would instead head off in the direction of a programmatic analysis of the disk usage. I doubt this is what the cloud providers do for billing, instead likely choosing something very close to the SQL AVG usage to come up with a number to bill based on.

PR #1475, opened 2021-01-22
test: Fix the Report expiration subtest handler functions

timflannagan1:
Update these report expiration subtest functions that are responsible
for testing the expiration functionality. Before, some of these report
objects that were created were flat out invalid, especially in the case
of the testing the rollup report scenarios.

In the case of the testing rollup reports, we need to create a
reportquery custom resource that acts as the aggregate query and sits
between the sub-report and the aggregate report. Then, we can list that
sub-report as an input to the aggregate report while still referencing a
valid reportquery custom resource.


timflannagan1:
Manually adding the valid-bug label as this only affects tests and nothing user-facing.

PR #1476, opened 2021-01-22
Bug 1926984: Avoid requeueing an expired Report

timflannagan:
Update the Report handler and avoid requeueing a Report that is
reporting an `expired` status. Before, this would result in the
reporting-operator hot looping as this Report custom resource would
continue to get queued up.

Instead, if we're encountering a Report that
has reached its expiration date, then we should avoid processing it in
the Report handler function and let the call site determine how to
process that Report further in the next cycle.

### DIFF
--- a/pkg/operator/reports.go
+++ b/pkg/operator/reports.go
@@ -414,11 +414,8 @@ func (op *defaultReportingOperator) handleExpiredReport(logger log.FieldLogger, 
 // the period has elapsed.
 func (op *defaultReportingOperator) runReport(logger log.FieldLogger, report *metering.Report) error {
 	// check if the report we're currently processing is considered
-	// "expired". If true, exit early and requeue that object so
-	// op.syncReport calls the proper handler for this resource.
+	// "expired". If true, exit early.
 	if reportExpired := isReportExpired(logger, report, time.Now()); reportExpired {
-		logger.Infof("requeueing report that has reached its expiration date during the op.runReport method")
-		op.enqueueReport(report)
 		return nil
 	}
 


timflannagan1:
Note: I still need to create a BZ for this and link that BZ in the PR title but I'm going to tackle that next week.

timflannagan:
Accidently added that lgtm label. Need to play around with this more.

/hold

timflannagan:
Need to land this before landing the InstallPlan fixes (known OLM bug) required to reduce the # of e2e flakes.

timflannagan:
```
    metering_upgrade_test.go:102: 
        	Error Trace:	assert.go:112
        	            				metering_upgrade_test.go:102
        	            				main_test.go:445
        	            	failed to wait for the reporting-operator deployment to report a successful upgrade: failed to wait for the reporting-operator Deployment to finish updating in the metering-hdfs-olm-upgrade namespace: timed out waiting for the condition
        	Test:       	TestMeteringUpgrades/HDFS-OLM-Upgrade
        	Messages:   	expected there would be no error
...
status:
  conditions:
  - lastProbeTime: null
    lastTransitionTime: "2021-02-09T21:48:06Z"
    message: '0/7 nodes are available: 3 node(s) had taint {node-role.kubernetes.io/master: }, that the pod didn''t tolerate, 4 Insufficient cpu.'
    reason: Unschedulable
    status: "False"
    type: PodScheduled
  phase: Pending
  qosClass: Burstable
```

timflannagan:
The upgrade test has passed a couple of times now but the flakes still persist. Going to try and tackle those flakes outside of this PR.

PR #1478, opened 2021-01-22
test/e2e: Deflate the MySQL e2e test

timflannagan1:
We're currently seeing metering e2e runs fail as OLM can't create the
metering installation due to conflict errors when updating resources
(typically CRDs), so this _attempts_ to try and space those go routines
out.


timflannagan1:
Manually adding the valid-bug label as this only affects tests and nothing user-facing.

PR #1480, opened 2021-02-03
[release-4.6] Dockerfile: Avoid using dockerhub images in the testing root image

timflannagan1:
Update the Golang base image we pull from to use the one from the
registry.svc.ci.openshift.org registry due to rate limiting when pulling
from dockerhub.


timflannagan1:
Need to force-merge as the changes in this PR affect the build root image that CI uses to run non-e2e-tests (e.g. unit). Any changes to the build root image are reflected in whatever branch CI is configured to use, and not the actual PR that changes it.

PR #1481, opened 2021-02-04
test: Handle failed InstallPlans during e2e installations

timflannagan:
Update the e2e package to deal with InstallPlan failures that we have no
control over. Currently, OLM will mark an InstallPlan as failed when it
encounters any errors, regardless of the context around that error.

In the context of metering, we're seeing more and more e2e flakes
centered around installations not being able to be installed due to
failed InstallPlan resources. These failed installations are typically
due to CRD conflicts being returned from the apiserver and OLM not
handling that error correctly and the result is metering can't be
installed.

We can workaround this by detecting any failed InstallPlan resources
that we know are due to conflicting resource versions and attempt to
re-install metering in the applicable test handlers.

### DIFF
--- a/test/deployframework/context.go
+++ b/test/deployframework/context.go
@@ -87,6 +87,10 @@ func (df *DeployFramework) NewDeployerCtx(
 	subscriptionChannel,
 	outputPath string,
 	extraLocalEnvVars []string,
+	deleteNamespace,
+	deleteCRD,
+	deleteCRB,
+	deletePVC bool,
 	spec metering.MeteringConfigSpec,
 ) (*DeployerCtx, error) {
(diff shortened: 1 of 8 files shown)


timflannagan:
I need to move a couple of things outside of this PR:

- Need to land the expired Report custom resources fix that has incorrect re-queue logic.
- Need to move any of the cmd,pkg/deploy changes outside of this PR.
- Stop running the NFS test every CI cycle.

timflannagan:
This appears to be working locally with all of the commits cherry-picked together.

/hold

timflannagan:
This may require backports to 4.6/4.7 as we're still seeing this pop up frequently in those release branches e2e runs.

PR #1482, opened 2021-02-05
test/e2e: Avoid using a custom 4.5 image in upgrade tests

timflannagan1:
Avoid creating a custom CatalogSource custom resource that contains a
custom 4.5 index image containing the metering many-to-many PVC fix. It
looks like this was already merged, but never updated, and we should be
upgrading from supported and released downstream versions of metering.


timflannagan1:
Welp, in hindsight this is a useless change - there's no 4.5 channel available in the redhat-operators catalog since 4.6, at least by default. While there's definitely value in avoiding the usage of custom index images that are used as an upgrade from baseline, we don't have many options _readily_ available.

/close

PR #1483, opened 2021-02-05
Avoid hardcoding deploy config deletion values

timflannagan:
Update the test e2e and deployframework packages and support explicitly
setting the delete* deploy config parameters while instantiating
deployer context instances. Previously, any new instances used the
default value and we had no flexibility with how we may set these values
after creating a new object.

There are no getter or setter methods in the deploy package that help us
set new values for that unexported field. We may want to invest in
creating a simple interface instead and use variadics, or something
along that line.


timflannagan1:
Hmm, unfortunately, it looks like there are no logs being produced by prow, so going to have to blindly retest that failed upgrade test.

/retest

PR #1485, opened 2021-02-05
test/e2e: Ensure all e2e test namespaces use correct label

timflannagan:
Update the MySQL and NFS pre-installation functions to use the correct
namespace label in the form of `name: <namespace_prefix>-metering-testin-ns`.
During the teardown function of the hack/e2e.sh bash script, which
serves as the entrypoint to the e2e suite, we search for any namespaces
that match that label, and attempt to delete them unless the e2e suite
has been run in developer mode.

Previously, the namespaces that were created as a result of these MySQL
and NFS pre-installation functions create a `name` label containing the
actual namespace, and not the prefix supplied to the e2e suite, in the
context of an individual install. This led to those namespaces not being
properly purged at the cleanup call site.

Other changes in this commit include:

- Aligning the grouping of package imports with the rest of the e2e
  package(s).
- Updating the other pre-installation functions to use that
  createTestingNamespace helper function.
- Minor cleanup of unnecessary commits, re-organizing functions, etc.

### DIFF
--- a/test/e2e/metering_manual_install_test.go
+++ b/test/e2e/metering_manual_install_test.go
@@ -10,18 +10,19 @@ import (
 	"testing"
 	"time"
 
-	"github.com/kube-reporting/metering-operator/pkg/deploy"
-	"k8s.io/client-go/kubernetes"
-
-	"github.com/kube-reporting/metering-operator/test/deployframework"
-	"github.com/kube-reporting/metering-operator/test/testhelpers"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+
 	corev1 "k8s.io/api/core/v1"
 	storagev1beta1 "k8s.io/api/storage/v1beta1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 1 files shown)


test/e2e/metering_manual_install_test.go
```diff
@@ -111,15 +109,41 @@ func testManualMeteringInstall(
...
+		return nil, err
+	}
+	if apierrors.IsNotFound(err) {
+		ns := &corev1.Namespace{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: namespace,
+				Labels: map[string]string{
+					"name": namespacePrefix + "-" + testingNamespaceLabel,
+				},
+			},
```
timflannagan1, line 130: I should've done a better job outlining the diff of this function before re-organizing to closer to the top of this package:

```diff
		ns := &corev1.Namespace{
			ObjectMeta: metav1.ObjectMeta{
				Name: namespace,
				Labels: map[string]string{
-					"name": namespace + "-" + testingNamespaceLabel,
+                                      "name": namespacePrefix + "-" + testingNamespaceLabel,
				},
			},
		}
```

Note: the `namespacePrefix` is defined in the e2e package, and this file is a part of this package so we can re-use, despite it being unexported.

PR #1487, opened 2021-02-08
*: Bump repository to use the 4.8 image tags and re-generate YAML manifests

timflannagan:
I ran the following script to update the majority of the repository to use the new images:

```bash
#! /bin/bash

set -eou pipefail
set -x

VERSION=${1:-4.7}
NEW_VERSION=${2:-4.8}

INCLUDE_DIR_LIST=(./{charts,manifests,bundle,hack,Documentation,pkg})
INCLUDE_FILES_LIST=(./{CHANGELOG.md,Makefile,Dockerfile*})

# Alright, I give up trying to get this work using `find` as the base command, as
# it looks like you need to do something more along the lines of `! -name <...> -o -name <...>`
# instead of passing it a list and letting it do the heavy lifting.
EXCLUDE_FILE_EXTENSIONS=(*logo.svg)

# Note: there's no logic in this script for updating the olm.skipRange annotation to
# use the new upgrade graph. In order to do that manually, update the helm chart values
# in ./charts/metering-ansible-operator/values.yaml and bump the lower bound, re-run
# `make metering-manifests` and verify those changes are correct.
# Also, there are some inputs to the hack/e2e.sh entrypoint script that need manual intervention,
# like the `$METERING_OLM_UPGRADE_FROM_SUBSCRIPTION_CHANNEL`, which needs to be updated for that
# lower bound as well. I'd recommend just searching for the lower bound channel in this case and
# making those changes ad-hoc.
sed -i "s/${VERSION}/${NEW_VERSION}/g" $(find "${INCLUDE_DIR_LIST[@]}" "${INCLUDE_FILES_LIST[@]}" -type f ! -name "${EXCLUDE_FILE_EXTENSIONS[@]}" -exec grep -Irwl "${VERSION}" {} \;) \
    && make metering-manifests \
    && rm -rf manifests/deploy/openshift/olm/bundle/${VERSION} \
    && rm -rf manifests/deploy/upstream/olm/bundle/${VERSION} \
    && rm -rf bundle/manifests/meteringoperator.v${VERSION}.0.clusterserviceversion.yaml
```

Note: there are a couple of files that need some manual intervention, and those files mainly reference the lower bound channel version (e.g. the 4.4 channel) in the supported upgrade graph.


timflannagan1:
Hmm, it looks like we're not getting any logs from those latest e2e and upgrade runs.

Hopefully, [#15664](https://github.com/openshift/release/pull/15664) solves that.

timflannagan:
Alright, we should be squared away now with e2e logs showing up again. It looks like that was due to a recent change in how artifacts were collected in the openshift CI and we were hardcoding that value in our configuration.

timflannagan:
Hmm, there are a couple of problems with the e2e suite (the report expiration work looks off) that needs to be addressed, but this needs to land first anyways. Going to push this through and get some more iterations while tackling those problems.

PR #1496, opened 2021-02-11
test/e2e: Reduce number of metering e2e tests per-run

timflannagan:
Update the main testing package and mark the NFS and node selector e2e test to be gated on the `testing.Short()` value we derive from the hack/e2e.sh bash script (entry point to the e2e suite).

Remove the Prometheus connector post-install test from the list of valid tests as we removed that functionality from the Operator.


timflannagan:
Hadoop image was broken which was causing e2e failures. Fixed (i.e. reverted the bump to 3.1.4) that image, so we should be back in business now.

/retest

timflannagan:
Alright CI should be substituting images correctly once again.

/test all

PR #1500, opened 2021-02-14
[release-4.6] Remove operator-courier from the root Dockerfile

timflannagan:
Remove the `operator-courier` dependency that was previously used to verify the metering manifests. During the 4.6 release cycle, we updated the metering-related CRDs to use the apiextensions.k8s.io/v1 version, which `operator-courier` is unable to process. As a result, we commented out any make targets that attempted to call this dependency but failed to remove that dependency from the root CI build Dockerfile.


timflannagan:
This needs to be force-merged (changes made to the CI build root don't propagate until they're merged) to get the release-4.6 CI green again.

PR #1507, opened 2021-02-16
[release-4.7] Handle failed InstallPlans in metering test installations

timflannagan:
Cherry-pick of all of the e2e and deploy package changes needed to fix how we handle failed InstallPlan custom resources in the various metering test installations we created.


timflannagan:
Manually adding the group-lead-approved label as this only affects testing code.

PR #1519, opened 2021-03-15
Bug 1939121: Add deprecation notice to the Metering CSV YAML manifests

timflannagan:
Add deprecation notice to the Metering CSV YAML manifests
  
- Update the name from `Metering` to `Metering (Deprecated)`
- Add deprecation notice to the long description. Link to the 4.6 OpenShift release notes.
- Add deprecation notice to the short description.

Screenshot from 2021-03-15 14-35-09

Needs to be backported to the release-4.6 branch.

PR #1520, opened 2021-03-18
[release-4.7] Bug 1940584: Add deprecation notice to the Metering CSV YAML manifests

timflannagan:
Manual cherry-pick of #1519.

- Update the name from `Metering` to `Metering (Deprecated)`
- Add deprecation notice to the long description. Link to the 4.6
  OpenShift release notes.
- Add deprecation notice to the short description.

Likely needs to be backported to the release-4.6 branch as well.

### DIFF
--- a/bundle/manifests/meteringoperator.v4.7.0.clusterserviceversion.yaml
+++ b/bundle/manifests/meteringoperator.v4.7.0.clusterserviceversion.yaml
@@ -208,16 +208,24 @@ metadata:
     certified: "false"
     containerImage: quay.io/openshift/origin-metering-ansible-operator:4.7
     createdAt: "2019-01-01T11:59:59Z"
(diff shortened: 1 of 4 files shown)


russellb:
@timflannagan see my comment on the bz.  This doesn't seem like high severity

(I am patch manager this week.)

timflannagan:
@russellb I would agree that it doesn't make much sense as high severity BZ, and TBH it was only marked high so the patch manager would evaluate it as a backport. We've run into issues where it's not obvious to people that metering has been deprecated, and this served as a barrier of entry for people that want to continue installing metering but aren't aware of the deprecation of the product.

mfojtik:
[patch-manager] :rocket: Approved for z-stream by score: 1.00

